PR #35, opened 2016-05-06
Dipole timestreams

tskisner:
This is a work in progress, do not merge this pull request yet- there needs to be more detailed unit tests.  Just opening this for feedback.

### DIFF
--- a/tests/test_ops_dipole.py
+++ b/tests/test_ops_dipole.py
@@ -1,0 +1,219 @@
+# Copyright (c) 2015 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by 
+# a BSD-style license that can be found in the LICENSE file.
+
+
+import sys
+import os
+
+if 'PYTOAST_NOMPI' in os.environ.keys():
+    from toast import fakempi as MPI
+else:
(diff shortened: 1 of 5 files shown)


keskitalo:
Calling this operator doubles the memory allocated for TOD. Often the need for the dipole is either to
1) subtract it, in which case the user should be able to provide the target cache name but have the dipole subtracted rather than added.
2) calibrate against it, in which case the dipole is (likely) only needed for one detector at a time. 
Case 1) is easy to add. Case 2) isn't critical as the pointing overwhelms the TOD memory requirement. I suppose fully supporting selective dipole generation would require making the methods available as functions with user-controlled inputs rather than operators?

tskisner:
Your point 2 is important.  I should split out the actual generation of the dipole to a free function (actually several functions).  I can put that in the tod_math.py file.  Then it could be called directly by higher-level operators doing calibration.  I will also add an option to the constructor which determines whether to add or subtract the generated timestream from the cache object.  Thanks for the feedback!

tskisner:
These commits have been squashed and rebased off master.  Unit test coverage exists, but is not yet complete.

Issue #31, opened 2016-05-02
Helper function for application of time-varying calibration


tskisner:
Add a free function to toast.tod.tod.py that takes a sparsely sampled set of gain factors, interpolates them to detector timestamps (using some interpolation order), and applies them element-wise to a data vector.

PR #37, opened 2016-05-06
Tod math

keskitalo:
Utility functions for manipulating arrays of quaternions and TOD.


keskitalo:
Addresses issue #31 .

PR #43, opened 2016-05-18
By default, do not flag the pointing matrix samples.  

tskisner:
Calling code can enable flagging of pointing matrix elements at construction time, using a bit mask applied to the common TOD flags.

### DIFF
--- a/toast/tod/pointing.py
+++ b/toast/tod/pointing.py
@@ -55,13 +55,15 @@ class OpPointingHpix(Operator):
         hwpstep: if None, then a stepped HWP is not included.  Otherwise, this
             is the step in degrees.
         hwpsteptime: The time in minutes between HWP steps.
+        common_flag_name (str): the optional name of a cache object to use for
+            the common flags.
         common_flag_mask (byte): the bitmask to use when flagging the pointing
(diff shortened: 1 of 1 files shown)


keskitalo:
Should there be a keyword to allow retrieving the common flags from the cache?

tskisner:
I guess I had never thought of that, and had just assumed that the TOD class for a particular experiment would return common flags based on housekeeping and other data.  However, I guess if the calculation of the common flags is external to the TOD class, then those flags might exist only in the cache.

If that is the typical scenario, then it seems to call into question the design of the TOD class.  What exactly should be provided only by the TOD class (and never in the cache)?  Just raw data from the data acquisition system?  For example, one could treat the pointing as some kind of "processed output" that is the result of using raw sensor data.

I will add another keyword for now, but we should seriously consider what we want the TOD class to return.  Maybe it should literally just return raw data at various sample rates and that is it.  Then we would have operators that produced pointing and other products in the cache.

tskisner:
Does that latest commit address your request for using cached flags?

keskitalo:
Yes they do! Merged.

PR #44, opened 2016-05-18
Improve fakempi to allow running toast_planck_reduce

keskitalo:
Small tweaks to better match real MPI communicator.

### DIFF
--- a/toast/fakempi.py
+++ b/toast/fakempi.py
@@ -55,6 +55,20 @@ class Comm(object):
         """
         return self._rank
 
+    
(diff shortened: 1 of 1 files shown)


tskisner:
This change is fine, but the fake mpi stuff is only intended to be used when installing toast at NERSC from the login node (where mpi is not available).  It should never really be used in production.

keskitalo:
Actually the fakempi is really useful for debugging on the login node. I
can set a break point and go inspect the contents of the arrays.

PR #46, opened 2016-05-24
Add reporting method to cache

keskitalo:


tskisner:
Is this method for just debugging?  We don't want thousands of processes making print statements...

keskitalo:
You wouldn't call it from all of the processes but it is informative to
look at the contents of the cache for even just one process.

PR #47, opened 2016-06-01
Distributed pixel FITS I/O

tskisner:
This adds read / write functions for healpix FITS format to the DistPixels class.  A few other small cleanups and a super-simple allreduce function.  We can do something fancier with that later if it becomes the bottleneck.  Test case for read/write works on 4 processes with differing submaps on each process.  I enabled stdout for the non-root processes when running the unittests, but we can disable again if it becomes too verbose.


tests/test_cov.py
```diff
@@ -191,18 +200,63 @@ def test_invnpp(self):
...
+        build_invnpp = OpInvCovariance(detweights=detweights, invnpp=invnpp, hits=hits)
+        build_invnpp.exec(self.data)
+
+        invnpp.allreduce()
+        hits.allreduce()
+
+        # invert it
+        checkdata = np.copy(invnpp.data)
+        covariance_invert(invnpp.data, 1.0e-3)
+        covariance_invert(invnpp.data, 1.0e-3)
```
keskitalo: Duplicate line?

tskisner:
no- if you view the whole file, you'll see I'm inverting it twice and comparing to the original.

tskisner:
Hello @keskitalo, if you have no further issues with this, can we merge it?

PR #53, opened 2016-07-29
Added method to multiply diagonal covariance matrices. Zero detector …

keskitalo:
…waits now trigger a shortcut in diagonal matrix accumulation.


tskisner:
This looks good Reijo.  I think in _multiply_covariance, you don't need to create a work vector, correct?

keskitalo:
Right! Just pushed a fix.

PR #55, opened 2016-08-08
Fix use of the RNG for noise generation

tskisner:
When using wrapped Random123 functions, use the MC index and stream as the key to select the generating function.  One element of the counter is used for the sample index.  Only use one returned value from the generator when creating uniform randoms, which ensures reproducibility of a given sample.  Use both returned values when generating a single normal distributed sample with Box-Mueller.  Also add much more extensive tests of the noise TOD normalization.


tskisner:
I changed the PSD integration in the unit test to match what is done in the noise TOD generation.  The results do not change.

PR #56, opened 2016-08-18
Autocov

keskitalo:
Add method to compute the power spectral density of gappy data. The cython extension is threaded to make best use of available resources. The method is called across a communicator and it will communicate the necessary overlaps and return the PSDs distributed across the communicator.


tskisner:
Hi Reijo, this looks great.  We could also make a timestream operator that computed the PSDs for all detectors and replaced the noise object for each observation with this estimate.  But that can be a future feature.

PR #65, opened 2016-10-10
Improvements to noise model and timestream simulations

tskisner:
Changes the Noise class to allow independent frequency bins for each stored PSD, extracts the simulation of one timestream into a function that can be better unit-tested and incorporated into multiple noise simulation functions, improves the frequency spacing of the AnalyticNoise class, and extends the unit tests to have a better variety of knee frequencies.

### DIFF
--- a/toast/tod/noise.py
+++ b/toast/tod/noise.py
@@ -30,10 +30,14 @@ class Noise(object):
         if detectors is not None:
             self._dets = detectors
             if psds is None:
-                raise RutimeError("you must specify a psd for each detector")
+                raise RutimeError("you must provide a dictionary of PSD arrays for all detectors")
             if freqs is None:
-                raise RutimeError("you must specify a frequency array for each detector")
+                raise RutimeError("you must provide a dictionary of frequency arrays for all detectors")
             for det in self._dets:
+                if det not in psds:
+                    raise RuntimeError("no PSD specified for detector {}".format(det))
+                if det not in freqs:
+                    raise RuntimeError("no frequency array specified for detector {}".format(det))
                 if psds[det].shape[0] != freqs[det].shape[0]:
                     raise RuntimeError("PSD length must match the number of frequencies")
                 self._freqs[det] = np.copy(freqs[det])
--- a/toast/tod/noise.py
+++ b/toast/tod/noise.py
@@ -15,32 +15,31 @@ class Noise(object):
 
     Args:
         detectors (list): names of detectors we have
-        freq (ndarray): array of frequencies for our PSDs
+        freqs (dict): dictionary of array of frequencies for our PSDs
         psds (dict): dictionary of arrays which contains the PSD values
             for each detector.
     """
 
-    def __init__(self, detectors=None, freq=np.array([0.0,], dtype=np.float64), psds=None):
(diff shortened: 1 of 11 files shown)


toast/tod/noise.py
```diff
@@ -15,32 +15,31 @@ class Noise(object):
...
-        
-        # the last frequency point should be nyquist
-        self._rate = 2.0 * self._freq[-1]
+        self._rate = {}
 
         if detectors is not None:
             self._dets = detectors
             if psds is None:
                 raise RutimeError("you must specify a psd for each detector")
+            if freqs is None:
```
tskisner: ok, will address this in a second.

toast/tod/sim_detdata.py
```diff
@@ -24,6 +24,82 @@
...
+
+    interp_freq = np.linspace(df, df*half, num=half, endpoint=True)
+    loginterp_freq = np.log10(interp_freq)
+
+    # ignore zero frequency
+
+    trimzero = False
+    if freq[0] <= 0.0:
+        trimzero = True
+
```
keskitalo: What is "freq" includes the negative frequencies? Rather than cropping just the zero frequency, we could discard all non-positive frequencies:
good = freq > 0
rawfreq = freq[good]

tskisner: True, this port of C code did not make best use of slicing.  Will change that.

toast/tod/sim_detdata.py
```diff
@@ -24,6 +24,82 @@
...
+    # we need.  Otherwise the extrapolation is not well defined.
+
+    if (rawfreq[0] > interp_freq[0]):
+        raise RuntimeError("input PSD does not go to low enough frequency to allow for interpolation")
+    if (rawfreq[-1] < interp_freq[-1]):
+        raise RuntimeError("input PSD does not go to high enough frequency to allow for interpolation")
+
+    # interpolate
+
+    interp = si.InterpolatedUnivariateSpline(lograwfreq, lograwpsd, k=1, ext=2)
```
tskisner: I guess there is no reason we would ever go to higher order?  Hopefully setting k=1 would fall through to a simpler algorithm, but I agree we should just use scipy.interpolate.interp1d for now.

tskisner:
Reijo and I had communication offline...  The high-pass filter technique above was copied from the old C++ toast source.  The original reasoning was that we have a Fourier domain kernel that has high resolution.  The fewest low-frequency values correspond to lengths of time that are longer than the simulation interval.  This means that those modes will contribute to an overall DC offset.  The original code incorrectly tried to avoid this by setting those frequencies to zero.

Instead, our plan is to leave in the all the low frequency modes in the kernel (except the zero frequency), and then subtract the DC level after simulating the timestream.

tskisner:
Another offline communication- I ran the unit tests with 500 MC realizations with and without the high-pass of the lowest few frequencies, and it seems like cutting those one or two modes does impact the distribution of TOD variances.  It also leaves the potential for cyclic correlations in the FFT.  So we have decided to maintain the current oversampling factor of two, and to maintain the highpass of the lowest couple points below the frequency corresponding to the simulation length.

PR #67, opened 2016-11-01
Fix issue with FFT

keskitalo:
Replace scipy fft calls with numpy rfft. Use np.fft.rfftfreq. Disable dysfunctional high pass filter. Use basic interpolation rather than loglog-interpolation for the PSD.


tskisner:
I believe the last commit fixes the problem, but am in progress implementing a unit test which compares the binned PSD values estimated from 100 noise realizations to the binned input PSD.

tskisner:
@keskitalo I think this is now ready, can you take a look at my changes and merge if you agree?

PR #72, opened 2016-11-23
Change the way that RNG streams are determined.

tskisner:
The underlying RNG library uses 2x64bit numbers for the key and 2x64bit numbers
for the counter.  These are now set according to:

key1 = realization * 2^32 + telescope * 2^16 + component
key2 = obsindx * 2^32 + detindx
counter1 = currently unused (0)
counter2 = sample in stream

The detector index is now returned as a dictionary from the TOD
class.  The observation index is contained in the observation
dictionary with the id key.  The component and telescope
indices are not yet used.  These changes result in different
realizations of the random timestream values compared to
previous versions of the code, but all statistical and unit
tests still pass.  For runs with multiple observations, this
now ensures reproducibility of the timestreams regardless of
which process group is assigned a given observation.

### DIFF
--- a/toast/tod/sim_detdata.py
+++ b/toast/tod/sim_detdata.py
@@ -24,7 +24,47 @@ from ..operator import Operator
 from .. import rng as rng
 
 
-def sim_noise_timestream(realization, stream, rate, samples, oversample, freq, psd):
+def sim_noise_timestream(realization, telescope, component, obsindx, detindx, rate, firstsamp, samples, oversample, freq, psd):
+    """
+    Generate a noise timestream, given a starting RNG state.
+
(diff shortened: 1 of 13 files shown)


toast/tod/sim_detdata.py
```diff
@@ -146,32 +188,31 @@ def exec(self, data):
...
         """
         comm = data.comm
-        rngobs = 0
 
         for obs in data.obs:
+            obsindx = 0
+            if 'id' in obs:
+                obsindx = obs['id']
+            else:
+                print("Warning: observation ID is not set, using zero!")
```
keskitalo: This is a good idea. It maintains backwards compatibility for the single observation data distributions we have had until now.

PR #83, opened 2017-02-14
Tweak autocovariance calculation: take the mean over unflagged sample…

keskitalo:
…s and require a reasonable number of unflagged samples.


tskisner:
Looks good to me!

PR #85, opened 2017-04-12
Change the toast data distribution. 

tskisner:
We now support simultaneous distribution in both the detector and sample direction.  Also add support for specifying hard breaks in the data distribution in both directions.  Update the TOD base class constructor (and all derived classes) to use these new parameters.

### DIFF
--- a/src/python/dist.py
+++ b/src/python/dist.py
@@ -141,7 +141,8 @@ def distribute_partition(A, k):
             low = mid + 1
     return low
 
-def distribute_discrete(sizes, groups, pow=1.0):
+
+def distribute_discrete(sizes, groups, pow=1.0, breaks=None):
     """
     Distribute indivisible blocks of items between groups.
 
@@ -156,6 +157,7 @@ def distribute_discrete(sizes, groups, pow=1.0):
         sizes (list): The sizes of the indivisible blocks.
         groups (int): The number of groups.
         pow (float): The power to use for weighting
+        breaks (list): List of hard breaks in the data distribution.
 
     Returns:
         A list of tuples.  There is one tuple per group.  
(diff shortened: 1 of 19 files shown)


src/python/dist.py
```diff
@@ -173,16 +175,32 @@ def distribute_discrete(sizes, groups, pow=1.0):
 
     off = 0
     curweight = 0.0
-    proc = 0
+
+    all_breaks = None
+    if breaks is not None:
+        # Check that the problem makes sense
+        all_breaks = np.unique(breaks)
```
keskitalo: I guess we can still call np.unique but originally it was needed because the breaks input was a list of lists.

src/python/dist.py
```diff
@@ -203,48 +221,126 @@ def distribute_uniform(totalsize, groups):
...
-            dist_samples = distribute_uniform(samples, mpicomm.size)
+
+def distribute_samples(mpicomm, detectors, samples, detranks=1, detbreaks=None, sampsizes=None, sampbreaks=None):
+    """
+    Distribute data by detector and sample.
+
+    Given a list of detectors and some number of samples, distribute
+    the data in a load balanced way.  Optionally account for constraints
+    on this distribution.  The samples may be grouped by indivisible
+    chunks, and there may be force breaks in the distribution in both
```
keskitalo: Forced breaks?

src/python/dist.py
tskisner: I think it is still useful to prune accidental duplicates in the input list.  We could construct a set from the breaks, which would also remove duplicates, but since we need a ndarray eventually anyway, I think this is fine...

src/python/dist.py
tskisner: Will fix typo and then merge

PR #90, opened 2017-05-30
Add polynomial filter operator.

keskitalo:
Add a simple polynomial filter operator.

### DIFF
--- a/src/python/tod/polyfilter.py
+++ b/src/python/tod/polyfilter.py
@@ -1,0 +1,137 @@
+# Copyright (c) 2015-2017 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by 
+# a BSD-style license that can be found in the LICENSE file.
+
+
+import numpy as np
+
+from ..op import Operator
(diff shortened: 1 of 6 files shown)


src/python/tod/polyfilter.py
```diff
@@ -0,0 +1,137 @@
...
+                    # read_flags always returns both common and detector
+                    # flags but we already cached the common flags.
+                    flag, dummy = tod.read_flags(detector=det)
+                    flag_ref = tod.cache.put(cachename, flag)
+                    del flag, dummy
+
+                # Iterate over each chunk.
+
+                chunk_first = tod_first
+                for curchunk in range(tod.local_chunks[1]):
```
tskisner: This operator uses the data distribution chunks (stationary intervals) to specify the L/R scans, and then flags the turnarounds.  We could also use the "Interval" list to specify the good data regions.  If we use the data distribution chunks to specify the subscans, then they will be very small, which will limit the size of timestream over which noise can be simulated.  If we don't want to change this, then we need some other way of specifying the noise stationary period.

PR #91, opened 2017-05-31
Performance improvements

tskisner:
This includes many small improvements to performance from a variety of sources:

- All quaternion operations and diagonal covariance functions
   now threaded.

 - Compiled quaternion operations now intrinsically support
   both single and multiple quaternions working on single or
   multiple quaternions / vectors.  This reduces the need for
   tiling in the python code.

 - Add fast inverse error function which is used when MKL is not
   available.  Gaussian random generation now uses the inverse
   error function.

 - When passing memory to quaternion functions with ctypes,
   ensure that memory is C-contiguous.

 - Fix a bug / performance issue in the DistPixels class.


tskisner:
Indeed.  Will fix in a moment.

PR #92, opened 2017-06-08
Add new C++ function for generic healpix pointing matrix calculation.…

tskisner:
…  Wrap that in C/Python.  Use this function in the OpPointingHpix class.  Other small updates to ctoast.py to enable passing NULL pointers to C functions as well as numpy arrays.

### DIFF
--- a/src/libtoast/ctoast.cpp
+++ b/src/libtoast/ctoast.cpp
@@ -642,7 +642,7 @@ void ctoast_pointing_healpix_matrix ( ctoast_healpix_pixels * hpix, int nest,
 
     toast::healpix::pixels * hp = reinterpret_cast < toast::healpix::pixels * > ( hpix );
 
-    std::string modestr = std::string(mode);
+    std::string modestr(mode);
     bool bnest = true;
     if ( nest == 0 ) {
         bnest = false;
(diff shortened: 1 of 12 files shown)


src/libtoast/ctoast.cpp
```diff
@@ -635,8 +635,24 @@ void ctoast_atm_sim_observe( ctoast_atm_sim * sim, double *t, double *az, double
...
 //--------------------------------------
 
+void ctoast_pointing_healpix_matrix ( ctoast_healpix_pixels * hpix, int nest, 
+    double eps, double cal, char const * mode, size_t n, double const * pdata,
+    double const * hwpang, uint8_t const * flags, int64_t * pixels, 
+    double * weights ) {
 
+    toast::healpix::pixels * hp = reinterpret_cast < toast::healpix::pixels * > ( hpix );
+
+    std::string modestr = std::string(mode);
```
keskitalo: why not just 
`std::string modestr(mode);
`?

tskisner: no reason except silly oversight.  Just fixed it.

src/python/tod/pointing.py
```diff
@@ -98,6 +136,11 @@ def __init__(self, pixels='pixels', weights='weights', nside=64, nest=False, mod
         # initialize the healpix pixels object
         self.hpix = hp.Pixels(self._nside)
 
+        if self._mode == "I":
+            self._nnz = 1
+        else:
+            self._nnz = 3
```
keskitalo: We'll have to revisit this if we are mapping only the differenced time streams or we have spurious components.

tskisner: Sure, this operator currently only supports 2 modes.  I can do an explicit elif here and throw an exception for unsupported modes if you like.

PR #94, opened 2017-06-16
Implement memory counter operator and use it in the ground simulation…

keskitalo:
… pipeline.


keskitalo:
I also cached the glob2loc vector in the distpixels class so it gets included in the counts.

tskisner:
Looks good to me, and it is a good idea to have the glob2loc vector in the cache- although if it gets too large that can hurt the lookup performance.  Also thanks for the white space / line length cleanup.

PR #101, opened 2017-07-12
cache should refuse to create objects under None key.

keskitalo:


tskisner:
Looks good, thanks for adding these checks.

PR #102, opened 2017-07-26
Working implementation of the mixing matrix noise.

keskitalo:
The noise stream index (earlier detector index) is stored in the Noise class instead of the TOD class. In a correlated noise simulation it would be confusing having to get the indices for every noise mode from the TOD object. This implies a required change in pipelines that have been setting the detector indices.

Also:
- Tweaking the type of raised Exceptions to better reflect the errors encountered
- Improve the docstrings in the affected files
- replace accessor functions with public members where the accessor added no value.
- number of tweaks (white space, variable naming) of the source to improve readability and conform to the style guide.
- Use a python3 feature to enforce setting named method arguments. This feature should probably be used everywhere where we require all keyword arguments to be set (instead of checking them explicitly)

### DIFF
--- a/src/python/tests/ops_simnoise.py
+++ b/src/python/tests/ops_simnoise.py
@@ -1,25 +1,17 @@
 # Copyright (c) 2015-2017 by the parties listed in the AUTHORS file.
-# All rights reserved.  Use of this source code is governed by 
+# All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
+
+import os
+
+import numpy as np
 
 from ..mpi import MPI
 from .mpi import MPITestCase
 
-import sys
-import os
-
-import numpy as np
-import numpy.testing as nt
-
-import scipy.interpolate as si
-
-from ..tod.tod import *
-from ..tod.pointing import *
-from ..tod.noise import *
-from ..tod.sim_noise import *
-from ..tod.sim_det_noise import *
-from ..tod.sim_tod import *
-
+from ..dist import Comm, Data
+from ..tod import (Noise, sim_noise_timestream, AnalyticNoise,
+                   OpSimNoise, TODHpixSpiral)
 from .. import rng as rng
 
 
@@ -33,19 +25,20 @@ class OpSimNoiseTest(MPITestCase):
 
         # Note: self.comm is set by the test infrastructure
         self.worldsize = self.comm.size
-        if (self.worldsize >= 2):
-            self.groupsize = int( self.worldsize / 2 )
+        if self.worldsize >= 2:
+            self.groupsize = self.worldsize // 2
             self.ngroup = 2
         else:
             self.groupsize = 1
             self.ngroup = 1
         self.toastcomm = Comm(world=self.comm, groupsize=self.groupsize)
-        self.data = Data(self.toastcomm)
+        self.data = Data(self.toastcomm)  # Uncorrelated simulation
+        self.data2 = Data(self.toastcomm)  # Correlated simulation
 
         self.dets = ["f1a", "f1b", "f2a", "f2b", "white", "high"]
-        self.fp = {}
-        for d in self.dets:
-            self.fp[d] = np.array([0.0, 0.0, 1.0, 0.0])
+        self.focalplane = {}
+        for det in self.dets:
+            self.focalplane[det] = np.array([0.0, 0.0, 1.0, 0.0])
 
         self.rate = 20.0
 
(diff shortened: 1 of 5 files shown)


src/python/tod/noise.py
```diff
@@ -1,102 +1,147 @@
...
+            detector must be defined in `freqs` and `psds`.
+        indices (dict):  Integer index for every PSD, useful for
+            generating indepedendent and repeateable noise realizations.
+            If absent, runnign indices will be assigned and provided.
+
+    Attributes:
+        detectors (list): List of detector names
+        keys (list): List of PSD names
+
+    Raises:
```
keskitalo: We should do a better job setting and specifying the Exception types.

src/python/tod/noise.py
```diff
@@ -1,102 +1,147 @@
...
-                if psds[det].shape[0] != freqs[det].shape[0]:
-                    raise RuntimeError("PSD length must match the number of frequencies")
-                self._freqs[det] = np.copy(freqs[det])
-                self._psds[det] = np.copy(psds[det])
-                # last frequency point should be Nyquist
-                self._rate[det] = 2.0 * self._freqs[det][-1]
-
-
-    @property
-    def detectors(self):
```
keskitalo: This is a good example of an accessor method not adding value to the code.

src/python/tod/sim_noise.py
```diff
@@ -21,29 +21,17 @@ class AnalyticNoise(Noise):
...
-            raise RuntimeError("you must specify the frequencies for high pass")
-        if fknee is None:
-            raise RuntimeError("you must specify the knee frequencies")
-        if alpha is None:
-            raise RuntimeError("you must specify the exponents")
-        if NET is None:
-            raise RuntimeError("you must specify the NET")
-
-        self._detectors = detectors
+    def __init__(self, *, detectors, rate, fmin, fknee, alpha, NET):
```
keskitalo: This is how we can remove a lot of boiler plate and still require user to set the keyword arguments. Optional keyword arguments are still supported by giving them a default value.

src/python/tod/sim_noise.py
```diff
@@ -21,29 +21,17 @@ class AnalyticNoise(Noise):
...
-        if fmin is None:
-            raise RuntimeError("you must specify the frequencies for high pass")
-        if fknee is None:
-            raise RuntimeError("you must specify the knee frequencies")
-        if alpha is None:
-            raise RuntimeError("you must specify the exponents")
-        if NET is None:
-            raise RuntimeError("you must specify the NET")
-
-        self._detectors = detectors
```
keskitalo: There was no need to set this private member in the derived class.

src/python/tod/tod.py
```diff
@@ -201,77 +209,74 @@ def grid_ranks(self):
         """
         return (self._rank_det, self._rank_samp)
 
-
     def _get(self, detector, start, n):
-        raise RuntimeError("Fell through to TOD._get base class method")
+        raise NotImplementedError("Fell through to TOD._get base class method")
```
keskitalo: This exception type is specific to virtual and yet unimplemented methods.

src/python/tests/ops_simnoise.py
```diff
@@ -1,25 +1,17 @@
...
-import numpy.testing as nt
-
-import scipy.interpolate as si
 
-from ..tod.tod import *
-from ..tod.pointing import *
-from ..tod.noise import *
-from ..tod.sim_noise import *
-from ..tod.sim_det_noise import *
-from ..tod.sim_tod import *
```
keskitalo: The wildcard import statements caused much grief with pylint. I have also grouped them according to the style guide to pass the pylint checks.

src/python/tod/noise.py
tskisner: Sure, that is a good goal :-)

src/python/tod/noise.py
tskisner: I disagree.  The list of detectors was intended to be set at construction time and never changed by the user.  Although the user can of course manually change the internal _dets variable, the use of @property is fairly standard practice:

https://stackoverflow.com/questions/14594120/python-read-only-property

src/python/tod/sim_noise.py
tskisner: I was unaware of the use of a bare asterisk to separate positional and forced keyword arguments.  Sounds fine.

src/python/tod/sim_noise.py
tskisner: Indeed, that is set by the base class constructor

src/python/tod/tod.py
tskisner: Agreed, that does seem more appropriate.

src/python/tests/ops_simnoise.py
tskisner: I basically only used wildcard imports inside the unittest source files.  Previously the unittests were external from the package, and so it seemed fine to sacrifice ultra-picky standards in the name of convenience.  However, we have now moved the unit tests into an internal sub-package.  Since this code is now part of the package itself, it does seem to make sense to enforce the normal best practices (i.e. no wildcard imports at all).

PR #107, opened 2017-08-10
cleanup `external/rules/mpi4py.sh`

ickc:
There seems to be a bug here. Out of the 4 commits here, c6dd0d4 is a bug fix and others are "cosmetic".

Before commit c6dd0d4, the `mpi.cfg` written looks something like this:

```
mpicc = @MPICC@
\
mpicxx = @MPICXX@
\
...
```

Here-doc is used to avoid this and made the syntax looks more readable.

I suspect similar things might happened elsewhere, I could take a look at other similar files if you want to.


tskisner:
Yes, I just fixed this yesterday as part of another cleanup of the build system.  The line continuation character are intentional, since this shell snippet is used both in the install script and in the creation of a docker file (where we want the command run as a single line).  I'm going to close this and instead add you as a reviewer of the pull request I'm about to open.

tskisner:
I Should also add that in general, these install scripts for external dependencies are only needed for maintaining the software stack at NERSC.  On a laptop, just apt-get or conda install mpi4py (which should pull in the correct mpich / openmpi version.

PR #108, opened 2017-08-10
Clean up build files

tskisner:
This moves the C language bindings into a separate sub directory in order to support parallel builds (i.e. "make -j xx").  Also fixes a problem where some of the gtest headers were not being placed in the distribution tarball, preventing the build from succeeding.  Also cleans up broken line endings in the mpi.cfg file generated when building the toast-deps module at NERSC.

@keskitalo , please wait for @ickc to look at this before merging.  I can't add Kolen as a reviewer since he is not on the project.


ickc:
I add one more patch in #109 (which merge to the cleanbuild branch that this pull request is from): it uses pipe whenever a tar is used. It should be functionally identical to without this patch.

In <https://github.com/hpc4cmb/toast/pull/107#issuecomment-321533764>, @tskisner said (emphasis mine),

>... The line continuation character are intentional, since this shell snippet is used both in the install script and in the creation of a docker file (**where we want the command run as a single line**)...

I'm not familiar with it enough to know if the current `external/rules/mpi4py.sh` in this pull request will still counted as a single line? I've another trick that make it absolutely a one liner while being somewhat readable:

```bash
printf $"%s\n" \
$"[toast]" \
$"mpicc = @MPICC@" \
$"mpicxx = @MPICXX@" \
$"include_dirs = @MPI_CPPFLAGS@" \
$"library_dirs = @MPI_LDFLAGS@" \
$"runtime_library_dirs = @MPI_LDFLAGS@" \
$"libraries = @MPI_LIB@" \
$"extra_compile_args = @MPI_EXTRA_COMP@" \
$"extra_link_args = @MPI_EXTRA_LINK@" \
```

`printf` is considered more portable than `echo` so it should be okay. I'm not sure about the necessity of the locale translation though. I thought as long as it is ASCII it is ok, no? And that translation is discouraged because of portability and security: [GNU gettext utilities: bash](https://www.gnu.org/software/gettext/manual/html_node/bash.html).

tskisner:
I guess I am not so concerned about the beauty of a shell snippet that is only used at NERSC to install dependencies and is not part of toast itself.  However, if @ickc really wants this printf change, I can make that change and test it next time I install the toast-deps module at NERSC (which will be very soon).  However, I don't want that to hold up this pull request, which fixes some long standing annoyances with the build system.  @keskitalo , if this looks good to you, could you merge when it is convenient?

ickc:
The printf trick is not important if it already works, I was just not sure how far we need to make it as a oneliner in the build process.

PR #109, opened 2017-08-10
external/*: when tar is used, use | instead of intermediate files

ickc:


tskisner:
Hello @ickc , untarring on the fly just saves us the temporary disk space of the tarball during the build of each specific tool.  The source directory and tarball are both wiped at the end of each install.  On the other hand this change does not hurt anything, so merging now.

Issue #110, opened 2017-08-19
MKL FFT failures


tskisner:
When building with gcc and linking to modern MKL versions, the MKL unit tests fail.  This happens for both single and batched FFTs.  This problem is not seen when using FFTW.  I will disable MKL FFTs in the mean time.

tskisner:
I thought this problem looked suspiciously familiar...  I actually mostly fixed this in an earlier commit:

https://github.com/hpc4cmb/toast/commit/30c24f91356fcae43e10bfece6f9f69c750b4ed2

But this was done directly to master accidentally as part of other changes and then reverted.  The fix / cleanup was then never migrated to a branch for proper testing and merging.  I'll cherry pick this into a branch now for testing.

tskisner:
With the latest master branch (after merging the reworking of MKL FFTs in ee43a97c9dace1a4cc1b881f78dc0fa5a9b8fae6), all C++ and python unit tests pass on a workstation using Intel compilers and MKL version 2018-beta.  On edison.nersc.gov, using gcc and MKL, the following behavior is seen:

1.  C++ "single" (one FFT) roundtrip passes.
2.  C++ "multi" (batch of FFTs) roundtrip fails.
3.  Python unit test of wrapper around C++ FFT roundtrip fails.  This uses a single FFT, but uses a persistent cache of FFT plans.
4.  Noise simulation unit test "fails", but this may or may not be related, since the threshold is kind of arbitrary.

Leaving this issue open to track why this combination (gcc + MKL on edison) fails while a pure Intel build does not.

tskisner:
Testing this on cori.nersc.gov (KNL) with Intel compilers and MKL also passes all tests.  So this seems to be an interaction between MKL and gcc-6.3 on edison.  I will test that same combination on a workstation to determine if it is NERSC-specific.

tskisner:
Ok, confirmed that I see the same failures on a linux workstation with latest MKL and gcc-6.3.  That should make it much easier to debug.

tskisner:
[...]

PR #114, opened 2017-09-15
Do not use MKL with GNU compilers

tskisner:
As described in #110 , do not use MKL except with the Intel compilers.  This is due to the fact that Python / numpy scripts frequently load MKL prior to importing toast- which itself will try to dlopen MKL.  This means that MKL used by python (with the Intel thread layer) must be compatible witth the MKL used with toast.  This implies that toast must also use the Intel threading layer, which implies compilation with the Intel compilers.


tskisner:
Since I've tested all unit tests with both the GNU and Intel toolchains, I'm going to self-merge this so that we can continue with updating toast-deps installations at NERSC.

PR #117, opened 2017-10-05
Add some example slurm scripts for NERSC

tskisner:
Also improve MPI lock class and get the documentation into a buildable state again.


tskisner:
This pull request just adds new scripts and fixes things like documentation.  Self-merging.

PR #118, opened 2017-10-11
Install healpy and pyephem from conda-forge

zonca:


tskisner:
conda forge is great!  However, their policy is to never use MKL and only standard blas/lapack.  Also, they have a policy that conda-forge packages should only use other packages from conda-forge.  So once you pull in packages from that channel, I believe your entire conda environment is switched to using conda-forge packages.  In other words, any numpy / scipy from Intel or Continuum channels that use MKL will be replaced by numpy / scipy from conda-forge.  Are we ok with that?

tskisner:
One benefit to requiring conda-forge packages is that we could also then create a feedstock for toast and enable installation through conda-forge.  Lack of Intel / MKL is annoying, but that topic is already being discussed:  https://github.com/conda-forge/staged-recipes/issues/3022

zonca:
on my machine when I installed `healpy` from conda-forge it didn't modify `numpy`, but it would be better to test this on other machines

tskisner:
I'll test this on an Intel python install I have.

PR #120, opened 2017-10-11
Bugfix in MPI import

zonca:
MPI is defined in the `mpi.py` module so relative import should be from `.mpi` instead of `.`


tskisner:
Fixed directly by 003973440f9ad5cd06e30996cb736050db962ddd

Issue #116, opened 2017-10-02
Preliminary test of integration of PySM


zonca:
I believe it would be useful to start wrapping our mind around integrating `PySM` (https://github.com/bthorne93/PySM_public) into `TOAST`.
We don't need to implement it yet, but it would be interesting to understand:

* if this integration would pose any constraint to `TOAST` or to `PySM`
* how `PySM` can leverage some `TOAST` features

In particular `PySM` is now a serial code but the core developers would like to make it parallel, first multi-threaded and then (possibly) distributed, can we do in a way that makes integration with `TOAST` easier?

@tskisner do you have already anything in mind about this? (we can discuss at the next call if it is quicker)

zonca:
I'll start writing a code based on the `OpSim` class see:

Scanning from a map.  Top-level operator:
https://github.com/hpc4cmb/toast/blob/b985b3dde91df0fdcfdc0aa24e9cf184bbb03e97/src/python/tod/sim_det_map.py#L116
This calls a function in the compiled library to do the scanning:
https://github.com/hpc4cmb/toast/blob/b985b3dde91df0fdcfdc0aa24e9cf184bbb03e97/src/python/ctoast.py.in#L836
and then:
https://github.com/hpc4cmb/toast/blob/b985b3dde91df0fdcfdc0aa24e9cf184bbb03e97/src/libtoast/tod/toast/sim_map.hpp#L15

@keskitalo you proposed to use "global" pixels (map spread on processes with no overlap) instead of "local" pixels, are global indices already available in the code?

@tskisner what is the more suitable test case I can use for testing this?

zonca:
@tskisner I checked out `tiny_satellite_shell.sh`, it uses `toast_satellite_sim.py`, I think a simpler script that uses directly the TOAST API would be probably easier for testing the PySM operator, would you have something like that?

tskisner:
The global pixel indices are what is actually stored in the cache.  These are then converted into locally hit submap and pixel within the submap.  Each process stores locally the submaps which are hit by its local data:  https://github.com/hpc4cmb/toast/blob/19f6315365e97fc1472c3f80154df27db70f73d7/src/python/tod/sim_det_map.py#L174

tskisner:
As for testing, I could create a stub of the operator and a unit test that creates fake pointing (spiral around healpix ring-ordered pixels.  I agree that the example satellite simulation pipeline is overkill for an isolated test- it is designed to show a full pipeline of operations.

[...]

PR #122, opened 2017-10-11
Add a stub of the new PySM operator and the beginning of a unit test.

tskisner:
This works towards addressing #116.  It is just the shell of an operator and unit test.  Much more would need to be added to both.  After installation, recall that you can run just this unit test with:

    $> python3 -c 'import toast; toast.test("ops_sim_pysm")'

### DIFF
--- a/src/python/tests/Makefile.am
+++ b/src/python/tests/Makefile.am
@@ -35,6 +35,7 @@ ops_dipole.py \
 ops_madam.py \
 ops_pmat.py \
 ops_simnoise.py \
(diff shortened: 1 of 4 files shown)


zonca:
thanks @tskisner this was very useful to implement #137 and #145, I think we can close this PR but it would still be useful to keep the branch alive (even if your commit is embedded in my pull requests).

zonca:
@tskisner we can close this

zonca:
@keskitalo @tskisner we can close this and possibly leave the branch alive

PR #123, opened 2017-10-16
Warnings fixes

jrmadsen:
Three primary updates:
1) Added warnings to platform scripts for GCC compilers
2) Greatly simplified getting the number of threads and the thread number in toast_qarray.cpp and toast_cov.cpp
3) Fixed many warnings involving sign comparisons, unused variables, etc. There are still many, many warnings when compiled with Elemental enabled.

... I may have accidentally auto-formatted some of the function parameters in 2 files (toast_cov.cpp and toast_qarray.cpp) while playing with the settings in Eclipse... I cleaned them up the best I could but many of these changes were a result of lines that vastly exceeded 80-character widths. I didn't realize it until I made this request. If there is a significant problem with these changes, I will do my best to restore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,10 @@ __pycache__/
 
 # Build directories when using ./cmake/platforms/...
 /build-*
+
+# Files from Eclipse IDE
+/.cproject
+/.settings
+/.project
+
+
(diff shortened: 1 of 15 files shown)


tskisner:
We have multiple branches with changes to these source files, and you ripped out the load balancing code I added (I was aware of the segfault).  I also have a separate branch that made changes to qarray, changing threading to omp simd, etc (I thought that was on my task list...).  Before merging anything from the recent hackathon we should have a local discussion- probably tomorrow.  The dungeon branch contains multiple orthogonal development efforts that we need to split into different merges.

PR #131, opened 2017-11-03
Simd array Implementation

jrmadsen:
It passes the compiled "toast_test". I haven't run the Python unit tests because macOS is making my life difficult... to the point that I clearly have to install a Linux in VirtualBox to keep from having to push so that I can pull on a Linux system.


tskisner:
Thanks for isolating this aligned memory alloc into a well-contained pull request.  I just pushed a tiny change to remove a generated script that got accidentally.  I assume that the vtune example changes were unrelated, but intentional.  Will merge after unit tests finish.

PR #137, opened 2017-11-07
[WIP] PySM operator

zonca:


src/python/tests/ops_sim_pysm.py
```diff
@@ -0,0 +1,127 @@
...
+        del self.data
+
+
+    def test_pysm(self):
+        start = MPI.Wtime()
+
+        npix = 12 * self.nside * self.nside
+
+        npix_local = int(np.ceil(npix / float(self.toastcomm.comm_world.size)))
+        local_pixels = np.arange(
```
zonca: I think the operator should work either with unique pixels in RING ordering, so that we can then call `libsharp` for smoothing. But it should be more general and allow any distribution. @tskisner how should we handle this? for now I just spread the pixels around.

src/python/tod/sim_det_map.py
```diff
@@ -193,3 +193,77 @@ def exec(self, data):
...
+            'use_smoothing': False,
+            'pixel_indices': self._local_pixels
+        }
+
+        local_map = {}
+        for ch_name, bandpass in self.bandpasses.items():
+            pysm_instrument_config["channels"] = [bandpass]
+            instrument = pysm.Instrument(pysm_instrument_config)
+            local_map[ch_name], _ = instrument.observe(self.sky, write_outputs=False)
+
```
zonca: @tskisner I don't know how to store the output. do you have an example of saving a map in the toast cache?

zonca: this map will then be loaded by the smoothing operator

zonca:
moved to #152

PR #144, opened 2017-11-16
Timing implementation

jrmadsen:
- Created C++ "auto_timer" and Python "auto_timer"
  - Start timer upon construction
  - Stop timer upon deconstructor
  - Python layer uses various utility functions to get unique tags
- All timers started in Python have prefix "[pyc]"
- All timers started in C++ have prefix "[cxx]"
- C++ timers are placed in the ctoast.cpp file
- All pipelines have been updated to use auto_timers


src/libtoast/util/toast_timing_manager.cpp
```diff
@@ -62,14 +63,41 @@ toast::util::timer& toast::util::timing_manager::timer(const string_t& key)
...
     timer_pair_t _pair(key, m_timer_map[key]);
     m_timer_list.push_back(_pair);
 
     return m_timer_map[key];
 }
+
 //============================================================================//
 
 
+
```
keskitalo: There is a ton of whitespace at the end of this file.

src/python/timing.py
```diff
@@ -3,27 +3,74 @@
...
+
+from os.path import dirname
+from os.path import basename
+from os.path import join
+
+#------------------------------------------------------------------------------#
+
+def FUNC(back = 1):
+    """Function that emulates __FUNCTION__ macro"""
+    ret = ("%s" % (sys._getframe(back).f_code.co_name))
```
keskitalo: Old school string formatting?

jrmadsen:
Ah yes. We can delete that whitespace at the end of the file. 

I already have a threshold set (see sex/libtoast/until/toast_base_timer.cpp. I think the function is called "output"). Basically the default precision of the timer is 3 decimal places (which in its current units, this is milliseconds) so anything less than 1/pow(10, precision) is omitted from the result.

The main timer will always appear first and it will set the width of the "time" string so even in larger runs it will be easy to gauge the magnitude of the time values (and thereby dismiss the irrelevant times) because all of the decimal places will always be exactly lined up vertically. Doing anything beyond that might be difficult given various run times and might introduce a potential for lost data. 

Feel free to remove any of the old timing. After basically spending an hour+ semi-mindlessly copy-and-pasting the "auto_timer" everywhere in Python, I couldn't mentally handle that task until later. 

Sent from my iPhone

--
Jonathan R. Madsen

jrmadsen:
This is used by the python auto_timer to get the file that the timer was created in. It's one of many I implemented just so I didn't have to customize each insertion of an auto-timer. 

Sent from my iPhone

--
Jonathan R. Madsen

tskisner:
The curly brace relocating in toast.cpp is just noise, but will be reverted anyway during the big format cleanup.  Please change the python string formatting to use format() instead of the percent sign, and then feel free to merge.

PR #145, opened 2017-11-21
[WIP] Smooth operator

zonca:
@tskisner @jdborrill first pass at the libsharp smoothing operator.
Once we have the distributed map object we can use that instead of the manual distribution that I am doing now.

Todo

* [ ] make travis run tests
* [ ] support multiple detectors
* [ ] use input distributed map object
* [ ] save output properly

### DIFF
--- a/src/python/tests/ops_smooth.py
+++ b/src/python/tests/ops_smooth.py
@@ -8,12 +8,85 @@ from .mpi import MPITestCase
 import sys
 import os
 
+import numpy as np
+import healpy as hp
+import libsharp
+
 from ..tod.tod import *
 from ..tod.pointing import *
 from ..tod.sim_tod import *
 from ..tod.sim_det_map import *
 from ..map.pixels import *
 
+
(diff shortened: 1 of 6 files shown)


src/python/tests/ops_smooth.py
```diff
@@ -0,0 +1,148 @@
...
+    to be optimized with cython or numba
+    """
+    local_pix = np.empty(local_npix, dtype=np.int64)
+    i = 0
+    for start, num in zip(startpix, ringpix):
+        local_pix[i:i+num] = np.arange(start, start+num)
+        i += num
+    return local_pix
+
+def distribute_rings(nside, rank, n_mpi_processes):
```
zonca: @tskisner I think this function could be useful for the implementation of the ring-distributed map

zonca:
for some reason Travis doesn't run on this branch, @tskisner can you please check if there is anything wrong in the Travis-CI configuration?

zonca:
to be added to #152

PR #146, opened 2017-11-23
Add units to input parameters in focalplane generation

zonca:
I guessed a couple from the default, please crosscheck before merging


keskitalo:
Merging this PR before adding randomized detector parameters.

Issue #149, opened 2017-11-30
Distributed Ring Map Data


tskisner:
We need a new class that stores distributed pixel values that are compatible with harmonic transforms.  This is one of the pieces needed by #148.  Each process will have a disjoint set of isolatitude rings.  The distribution of pixels will be compatible with libsharp.

The new class should be similar to the DistPixels class:
https://github.com/hpc4cmb/toast/blob/master/src/python/map/pixels.py#L117

and should take in the constructor the communicator, healpix nside, and number of nonzeros per pixel.  Does libsharp store (I,Q,U) values contiguously per pixel?  The constructor should use the same technique as this existing code:
https://github.com/hpc4cmb/toast/pull/145/files#diff-b09bed3487e621995f1f7d685f905064R34

The data member should be a toast.Cache() object:
https://github.com/hpc4cmb/toast/blob/master/src/python/map/pixels.py#L164

and should store the data values in one buffer and the pixel indices in another:
https://github.com/hpc4cmb/toast/blob/master/src/python/map/pixels.py#L182

The class should have some properties to access the communicator, nnz, etc.  It should also store the libsharp grid object.

zonca:
implemented in #152, this issue can be closed

PR #152, opened 2017-12-04
[WIP] DistRings, PySM and Smooth

zonca:
Work in progress, do not merge.

Implemented `DistRings` operator as described in #149 

No need to review the code yet, just need to ask a question, see following inline question.


src/python/map/rings.py
```diff
@@ -0,0 +1,160 @@
...
+        # returns start index of the ring and number of pixels
+        startpix, ringpix, _, _, _ = hp.ringinfo(
+                self._nside, self._local_ring_indices.astype(np.int64))
+
+        local_npix = self._libsharp_grid.local_size()
+        self._local_pixels = self._cache.create(
+                "local_pixels", shape=(local_npix,), type=np.int64)
+        expand_pix(startpix, ringpix, local_npix, self._local_pixels)
+
+        self.data = self._cache.create(
```
zonca: @tskisner quick question: how should I store maps for each channel? should I get `num_channels` as input parameter to `DistRings` and then create a 3D `data` buffer `(npix, nnz, num_channels)`?

pipelines/toast_satellite_sim.py
```diff
@@ -356,6 +420,60 @@ def main():
...
+            smooth = tm.LibSharpSmooth(comm.comm_world, signal_map="sky_"+ch_tag,
+                                       lmax=lmax, grid=dist_rings.libsharp_grid,
+                                       fwhm_deg=fwhm_deg, beam=None)
+            smooth.exec(local_maps)
+
+            n_components = 3
+
+            full_map_rank0[ch_tag] = assemble_map_on_rank0(comm, local_maps[ch_tag], n_components, npix)
+
+        #full_map_rank0 dict contains on rank 0 the smoothed PySM maps
```
zonca: @tskisner I have started working on satellite sim, I am assembling the map on rank 0. My idea was to use then `scan_signal` like in the ground sim

src/python/map/pixels.py
```diff
@@ -472,6 +472,76 @@ def read_healpix_fits(self, path, comm_bytes=None):
         return
 
 
+    def broadcast_healpix_map(self, fdata, comm_bytes=None):
```
zonca: @tskisner then inside `scan_signal` use this copy of `read_healpix_fits` that instead of reading from disk broadcasts the map I assembled on rank 0. Do you think this is fine?

zonca:
@tskisner maybe simpler instead to do what we were talking before and just copy the map to all processes and then scan it locally, where can I find the pixel indices from the pointing?

pipelines/toast_fake_focalplane.py
```diff
@@ -79,6 +85,9 @@
     dets[d]["fmin"] = args.psd_fmin
     dets[d]["alpha"] = args.psd_alpha
     dets[d]["NET"] = args.psd_NET
+    dets[d]["fwhm_deg"] = fwhm  # in degrees
+    dets[d]["bandcenter_ghz"] = args.bandcenter_ghz
+    dets[d]["bandwidth_ghz"] = args.bandwidth_ghz
```
zonca: here I set `fwhm` and band parameters into the focalplane

pipelines/toast_ground_sim.py
```diff
@@ -1310,6 +1339,94 @@ def main():
...
+            bandcenter, bandwidth, fwhm_deg[det] = extract_detector_parameters(det, schedules)
+            bandpasses[det] = \
+                (np.linspace(bandcenter-bandwidth/2,
+                             bandcenter+bandwidth/2,
+                             N_POINTS_BANDPASS),
+                 np.ones(N_POINTS_BANDPASS))
+        pysm_sky = tm.PySMSky(local_pixels=dist_rings.local_pixels,
+                              nside=args.nside,
+                              pysm_sky_config=pysm_sky_config,
+                              bandpasses=bandpasses)
```
zonca: pysm already runs only on local detectors, so no need to use rank communicator

pipelines/toast_ground_sim.py
```diff
@@ -1310,6 +1339,94 @@ def main():
...
+            #full_map_rank0 dict contains on rank 0 the smoothed PySM maps
+
+            if comm.comm_world.rank == 0:
+                print(full_map_rank0[:10])
+
+            distmap.broadcast_healpix_map(full_map_rank0)
+            scansim = tt.OpSimScan(distmap=distmap, out='signal_'+det)
+            scansim.exec(data)
+            if comm.comm_world.rank == 0:
+                print("Time domain signal", data["signal_fake_0A"][:10])
```
zonca: @tskisner @keskitalo  at this point I have `signal_detname` inside `data`, in the orginal script there was only one single signal, do you have any suggestion on how to handle this? I don't understand what `sigcopy` is doing after this point.

pipelines/toast_ground_sim.py
```diff
@@ -1310,6 +1339,94 @@ def main():
...
+        if comm.comm_world.rank == 0:
+            print('Collecting, Broadcasting map', flush=args.flush)
+        start = MPI.Wtime()
+
+        full_map_rank0 = {}
+        for det, fwhm in fwhm_deg.items():
+            # LibSharp also supports transforming multiple channels together each with own beam
+            smooth = tm.LibSharpSmooth(comm.comm_world, signal_map="sky_"+det,
+                                       lmax=lmax, grid=dist_rings.libsharp_grid,
+                                       fwhm_deg=fwhm, beam=None)
```
zonca: I think here we could use the rank communicator, if I understand how to use it.

pipelines/toast_ground_sim.py
```diff
@@ -16,16 +16,39 @@
...
+    for obs in data.obs:
+        tod = obs['tod']
+        local_dets.update(tod.local_dets)
+    return local_dets
+
+def assemble_map_on_rank0(comm, local_map, n_components, npix):
+    full_maps_rank0 = np.zeros((n_components, npix), dtype=np.float64) if comm.comm_world.rank == 0 else None
+    comm.comm_world.Reduce(local_map, full_maps_rank0, root=0, op=MPI.SUM)
+    return full_maps_rank0
+
```
tskisner: These last 2 functions belong inside the operator which generates timestreams, such as this:  https://github.com/hpc4cmb/toast/pull/122/files#diff-22d562860acea7384e79d8faca453dc5R197
The extract_local_dets() would be called at the very beginning of the exec() method, before the main loop over detectors.  This timestream operator does not exist in this pull request, but the skeleton was in that other example.

pipelines/toast_ground_sim.py
tskisner: This code block is basically the beginning of the exec() method of the "OpSimPySM" operator (or whatever you want to call it) which should live in the toast.tod package.

pipelines/toast_ground_sim.py
```diff
@@ -1310,6 +1339,94 @@ def main():
...
+                (np.linspace(bandcenter-bandwidth/2,
+                             bandcenter+bandwidth/2,
+                             N_POINTS_BANDPASS),
+                 np.ones(N_POINTS_BANDPASS))
+        pysm_sky = tm.PySMSky(local_pixels=dist_rings.local_pixels,
+                              nside=args.nside,
+                              pysm_sky_config=pysm_sky_config,
+                              bandpasses=bandpasses)
+        local_maps = dict()  # FIXME use Cache instead
+        pysm_sky.exec(local_maps, out="sky")
```
tskisner: This is not a toast operator, so is not writing timestreams to the toast.tod.cache() for every observation.  So what is the "out" parameter used for?

pipelines/toast_ground_sim.py
tskisner: Ah, now I see.  You are using the OpSimScan operator to do the actual timestream generation.  Originally (before the decision to reduce the data to a single process), I had thought that the sampling would be part of the "OpSimPySM" operator.  But since you are using the OpSimScan operator to do the sampling, then all the pysm work becomes just map-domain classes and functions.  In the medium term, it might still be nice to put all this into its own operator class (tiny amount of copied code from OpSimScan), just so that all the "setup" is in one place and we can then call it more easily from other pipelines.

src/python/map/pysm.py
```diff
@@ -0,0 +1,60 @@
...
+            'add_noise': False,
+            'output_units': 'uK_RJ',
+            'use_smoothing': False,
+            'pixel_indices': self._local_pixels
+        }
+
+        for ch_name, bandpass in self.bandpasses.items():
+            pysm_instrument_config["channels"] = [bandpass]
+            instrument = pysm.Instrument(pysm_instrument_config)
+            local_map[out + "_" + ch_name], _ = \
```
tskisner: Ah, now I see why you are using the "out" constructor key word, since you are caching the maps for every detector.  If all this was wrapped inside a toast.Operator, would you instead be able to generate the map for one detector, smooth it, sample from it for all observations, and then throw it away to save space?

pipelines/toast_ground_sim.py
zonca: sure, I'll wrap it in an operator when it is almost completed, it is easier to develop directly in the script

src/python/tod/sim_det_map.py
```diff
@@ -193,3 +195,124 @@ def exec(self, data):
...
+                             N_POINTS_BANDPASS),
+                 np.ones(N_POINTS_BANDPASS))
+
+        lmax = 512
+
+        if self.comm.comm_world.rank == 0:
+            print('Collecting, Broadcasting map', flush=True)
+        start = MPI.Wtime()
+        local_maps = dict()  # FIXME use Cache instead
+        for det in local_dets:
```
zonca: @tskisner now I'm generating one detector at a time with PySM

pipelines/toast_ground_sim.py
```diff
@@ -1310,9 +1315,21 @@ def main():
 
     localpix, localsm, subnpix = get_submaps(args, comm, data)
 
-    # Scan input map
+    if args.input_pysm_model:
 
-    signalname = scan_signal(args, comm, data, counter, localsm, subnpix)
+        op_sim_pysm = tt.OpSimPySM(comm=comm,
```
zonca: @tskisner everything moved inside the operator

pipelines/toast_ground_sim.py
```diff
@@ -1310,9 +1315,21 @@ def main():
...
+                                   out='signal',
+                                   pysm_model=args.input_pysm_model,
+                                   focalplanes=[s[2] for s in schedules],
+                                   nside=args.nside,
+                                   subnpix=subnpix, localsm=localsm)
+
+        op_sim_pysm.exec(data)
```
zonca: @tskisner @keskitalo before `data` had just one object `signal`, now instead I have one `signal_detname` per detector. How should modify the copy operations `setup_sigcopy` and `copy_signal`?

keskitalo: Those functions have changed quite a bit in my working copy of the ground simulation script. We no longer create persistent CacheCopy operators. The pysm operator just needs to cache the signal under the same name as the old ScanSignal operator did.

zonca: thanks, the problem is that the old `scan_signal` had only 1 signal, now I have 1 signal per channel. Now I am saving it under `signal_fake_0A`, `signal_fake_0B` and so on. Is it ok?
Also should I rebase on your new version? is it in master?

keskitalo: That is the correct naming convention for cache objects. I'm still testing the changes to the pipeline script in my branch. I'll let you know once it is posted.

The layout of the pipeline has changed enough that you may find it easier to insert the operator call directly into the new pipeline rather than rebasing.

zonca:
now implementing this directly into the new ground script, see #164

PR #154, opened 2017-12-06
Fix unit test segfault

tskisner:
This renames the gtest unit tests so that we can select only the ones in the toast package.  It also disables tidas import until upstream problems are fixed.


keskitalo:
Looks good! I'm merging this to master.

PR #163, opened 2017-12-18
Split mapmaking

keskitalo:
Split mapmaking in the ground simulation is ready to go. Also, added a separate example.


zonca:
better if we first merge #164 into this branch before merging into master, it should be ready in a couple of hours

keskitalo:
Merging after receiving the go-ahead from @tskisner.

PR #171, opened 2017-12-20
Randomize focalplane

keskitalo:
Also, add checks for missing libsharp.


keskitalo:
Self-merging this small PR.

PR #172, opened 2017-12-20
Add timers to ground simulation pipeline, weather module and the PySM…

keskitalo:
… objects


keskitalo:
Merging after revision.

PR #175, opened 2017-12-20
Remove Cereal submodule

tskisner:
Rather than have users do submodule commands, just include the latest cereal headers directly in the source tree.


tskisner:
Approved offline by @jrmadsen

PR #179, opened 2017-12-21
Signal capture handling with timing_manager::report() callback

jrmadsen:
- Implemented signal capture for SIGINT, SIGABRT, SIGSEGV, etc.
- Floating point exceptions are not covered in this commit
- When catching a signal, timing_manager reports
- timing_manager will also report to timing_report_err_<ERRCODE>.out
- Verified that at least SIGINT will get captured when running via Python, suggesting that SIGABRT, etc. in Python will also work.
- CMake changes improving the CTest/CDash suite
- Minor updates to examples


jrmadsen:
Yes, the caveat is that some of the signals, although attempted to be caught, will not be caught because of how the OS treats them. For example, SIGKILL doesn't seem to be ever caught and the man pages state explicitly that SIGSTOP cannot be caught.

Other than that, as long as the Python script calls the toast::init() function, we should, it seems, theoretically be able to catch the signals whether they occur C++, Python, and Fortran. Some issues may arise with the Python due to Python raising signals in the Python virtual machine, which the latest docs stated is why the Python signals library wouldn't catch signals from C/C++ extensions. Fortran code raising signals don't seem likely to give us problems. These are OS-defined signals that the process is raising and since the Fortran code is in the same process and both the C++ and Fortran code are compiled into binaries, I would think it wouldn't make a difference whether this signal handling was activated in either C or Fortran. 

Jonathan 
 
--
Jonathan R. Madsen

PR #180, opened 2017-12-22
switch PySM to K_CMB

zonca:


jdborrill:
Presumably this just changes the output units, not the band integration
units? Seems too simple to be true. Are the RJ/CMB conversion factors
stored/calculated inside the code, and do you know which convention they
use?

J

zonca:
bands are always integrated in K_RJ

PR #184, opened 2018-01-03
Fix small bugs

keskitalo:
- Fix small issue in the PySM operator that lead to Cache warnings. The member objects must be destroyed in the right order to avoid clearing a cache with outside references.
- Fix issue in non-pole ground scheduling


keskitalo:
Merging these fixes before docker build.

PR #190, opened 2018-01-08
Fix error in atmosphere boundary calculation and add better error han…

keskitalo:
…dling.

Also, add threading back to atm::observe with exception handling suited for OpenMP.


keskitalo:
Self-merging after clearing the mpi_shmem edit with @tskisner. All other changes affect the atmosphere simulation.

PR #191, opened 2018-01-10
BurstBuffer/FsBuffer implementation

jrmadsen:
- New FsBuffer (file system buffer) class in cache.py to handle burst-buffe/file-system caching of cache objects
- Disable with Cache.use_fsbuffer = False (python)
- Disable with TOAST_NO_FSBUFFER > 0 (env variable)
- Set fscache temporary directory with FsCache.temporary_directory (python)
- Set fscache temporary directory with TOAST_TMPDIR (env variable)
- Added BurstBuffer support to SLURM examples
- Added TOAST_TMPDIR to all SLURM and shell examples


src/python/cache.py
```diff
@@ -308,9 +452,35 @@ def reference(self, name):
         ref = self.exists(name, return_ref=True)
         if ref is None:
             raise RuntimeError("Data buffer (nor alias) {} does not exist".format(name))
+        if Cache.use_fsbuffer and free_alloc and name in self._refs.keys():
+            self.destroy(name, remove_fsbuffer=False)
         return ref
 
 
+    def clone(self, name):
```
keskitalo: I don't think this is what we had in mind during the discussion. We wanted a `copy` method that would return a plain Numpy ndarray. Operators using the ndarrays could then use the acquired copies as they like and have them garbage collected once they are done.

keskitalo (commented):
This is a great start but it is missing the `swap` and `copy` methods we agreed upon. Unswapping would then happen implicitly when the client code asks for a reference to a swapped object. Asking for a copy, instead, would not wipe the swapped object but rather the swapped object would be loaded and copied without anything left in the cache memory.

jrmadsen (commented):
`FsBuffer` provides the capabilities to read/write to the file-system/burst-buffer. Any of the additional edits involving the logic in `Cache` or method names in `FsBuffer` can be replaced/modified to y'all's preference.

I think the `swap` and `copy` methods exist, just under different names. I avoided using the word _swap_ as a function because that function name is possibly confused with function switching the assignment of variables (more easily expressed in C):

```c
void swap(int* a, int* b)
{
    int* tmp = b;
    b = a;
    a = tmp;
}

struct simple
{
    void swap(int* rhs) { swap(m_data, rhs); }
private:
    int* m_data;
}
```

A `swap` method in the above interpretation doesn't seem to make much sense though when dealing with file-systems. Instead, one would just have FsBuffer objects for `a` and `b` objects.  

Thus, instead of `swap` and `copy` as the names, I used `unload` and `load` in `FsBuffer`. Loading a numpy array from the file-system is inherently an operation that returns a copy of the data since this is read-only memory. And I called writing a copy of the numpy array in RAM to physical swap "unloading". Maybe `store` would be a better function name?

Cache is then responsible for handling freeing the RAM. For this, I provided `Cache.free(...)` as an alternative to `Cache.clear(...)`. Where the clear function will remove all references in RAM and FsBuffer references, the free function will remove all the references in RAM but keep the FsBuffer references so that when a reference is requested later, it can load the reference from the FsBuffer. The `Cache.clone(...)` method allows one to obtain a copy of the data in the way that you requested: if there is a FsBuffer reference, if will load the copy and not insert a reference into the Cache object. I think I should switch the logic though, so that it checks for the RAM copy first and then use the FsBuffer as the fallback instead of vice versa.

The logic that I provided (that you should replace if you don't like it) for handling whether to implicitly or explicitly remove the Cache reference is handled through the `remove_fsbuffer=...` arguments. By default, nothing is implicitly removed from RAM. The current default arguments to previously existing methods do not implicitly free any memory, as requested in our discussion.

PR #192, opened 2018-01-16
Faster initialization

keskitalo:
- Speed up the ground TOD initialization by explicitly measuring the time-dependent coordinate transformation quaternions between horizontal and equatorial frames.
- Fix error in interpreting the azimuth angle. It is measured in the opposite direction than usual longitude.


tskisner:
I manually restarted the travis tests to check before merging.

keskitalo:
How does this PR look like, @tskisner?

tskisner:
Sorry for the delay, yes merge away.

PR #193, opened 2018-01-17
Clean up top-level imports

tskisner:
@keskitalo, this implements the changes to the importing of the tests that we discussed last week.  Obviously we won't know if this helps the high-concurrency problem until the next large run, but it is a good cleanup regardless.  Also trailing whitespace removal.

### DIFF
--- a/src/python/__init__.py
+++ b/src/python/__init__.py
@@ -11,21 +11,8 @@ designed to allow the processing of data from telescopes that acquire
 data as timestreams (rather than images).
 """
 
-import sys
-
-# ensure mpi4py hasn't been imported yet
-if 'mpi4py' in sys.modules:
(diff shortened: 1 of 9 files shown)


src/python/__init__.py
```diff
@@ -11,21 +11,8 @@
...
-# import MPI if not done already
-# (should be) backwards compatible with old recommendation of:
-#   from toast.mpi import MPI
-# as the first line
-if not 'toast.mpi' in sys.modules:
-    from . import mpi
-
-from .tests import test
+# Import MPI as early as possible, to prevent timeouts from the host system.
+from . import mpi
```
keskitalo: We no longer need to check if mpi4py was imported before toast?

keskitalo:
Merging after review.

PR #194, opened 2018-01-17
Fix slewing of the precession axis in satellite example

### LINKED ISSUE #159: Satellite precession axis in example script not slewing correctly
In the satellite example script, we have switched to using independent observations, but forgot to apply the global sample offset when slewing the precession axis in the anti-sun direction.

tskisner:
@zonca , I believe this will fix https://github.com/zonca/pico-simulations/issues/2.  At least I now see the proper hit map when running a boresight sim for several different numbers of observations.  Closes #159

### DIFF
--- a/pipelines/toast_satellite_sim.py
+++ b/pipelines/toast_satellite_sim.py
@@ -172,7 +172,7 @@ def main():
     stop = MPI.Wtime()
     elapsed = stop - start
     if comm.comm_world.rank == 0:
-        print("Create focalplane:  {:.2f} seconds".format(stop-start), 
+        print("Create focalplane:  {:.2f} seconds".format(stop-start),
             flush=True)
     start = stop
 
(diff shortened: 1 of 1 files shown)


zonca:
thanks @tskisner I'll run the pico sim again and report back here

zonca:
my sims are failing with a runtime error just after `  0 : reduce_pixels completed in   1.339 s`, however the example tests work, so I am debugging what has changed in the slurm scripts

    *** Error in `python3': double free or corruption (fasttop): 0x00002aaaf00450e0 ***

zonca:
it was an issue with groupsize, I am doing just 1 pixel and I needed to force it to 1 otherwise it would give a warning about groupsize to high and crash.

Now the hitmap is as expected:
https://github.com/zonca/pico-simulations/issues/2#issuecomment-358660600

keskitalo:
Merging after successful test.

PR #198, opened 2018-01-19
Toast planck edits

keskitalo:
Toast Planck requires the ability to pass keyword arguments through the `tod` class. This pull request adds the missing `**kwargs` arguments to the new `local` methods. It also encodes default cache names for the usual TOD vectors into the `tod` object.


tskisner:
Looks fine, merge when ready.

PR #203, opened 2018-01-26
Support for PySM and libsharp in the satellite simulation pipeline

zonca:


pipelines/toast_satellite_sim.py
```diff
@@ -553,6 +663,9 @@ def main():
             nse = tt.OpSimNoise(out="noise", realization=mc)
             nse.exec(data)
 
+            # add sky signal
+            add_sky_signal(args, comm, data, totalname="noise", signalname=signalname)
```
zonca: do you prefer I rename the key to "signal" now?

zonca:
@tskisner @keskitalo this is ready for a first review.

Should I also implement the capability of rescanning an existing map?

pipelines/toast_satellite_sim.py
tskisner: Sure, that would make sense.

tskisner:
I've not tested the modifications, but they look fine to me.

pipelines/toast_satellite_sim.py
zonca: ok, I will rename it to `tot_signal` because `signal` is already the sky signal

zonca:
ok @tskisner, I renamed the object in the cache and tested on `tiny_satellite_shell.sh`

Issue #200, opened 2018-01-23
Circular dependencies between toast.map and toast.tod


keskitalo:
We have a circular dependency:

> $ python3 -c 'import toast.map'
> OpenMP number of threads: 4
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/Users/reijo/Software/lib/python3.6/site-packages/toast/map/__init__.py", line 8, in <module>
>     from .madam import OpMadam
>   File "/Users/reijo/Software/lib/python3.6/site-packages/toast/map/madam.py", line 18, in <module>
>     from ..tod import TOD
>   File "/Users/reijo/Software/lib/python3.6/site-packages/toast/tod/__init__.py", line 23, in <module>
>     from .sim_det_map import OpSimGradient, OpSimScan, OpSimPySM
>   File "/Users/reijo/Software/lib/python3.6/site-packages/toast/tod/sim_det_map.py", line 13, in <module>
>     from ..map import DistRings, PySMSky, LibSharpSmooth, DistPixels
> ImportError: cannot import name 'DistRings'

We'll need to create a new package for the modules that require both `map` and `tod`.

PR #204, opened 2018-01-29
create new package todmap for pysm operator

zonca:
@keskitalo is this ok to address #200?

do you have a better suggestion for the name of the package?
I used `scan` because the operator scans a map domain object to a TOD.


zonca:
I have some trouble installing the new package. @tskisner can you please point me (I might have asked it already) to where do I tell the build system to install other files?

Is there a spot in the docs I can add this so I can stop asking it?

keskitalo:
@zonca this works for me. Although the name could be something more generic so we can add other operators that depend on `map` and `tod` later on. @tskisner what do you think about `todmap`?

tskisner:
@keskitalo , todmap is a fine suggestion.  Then we can move the other stuff in sim_det_map.py into that new subpackage.  @zonca, for the build system, the new directory needs to be added to src/python/CMakeLists.txt and also src/python/Makefile.am (in subdirs at the top).  Then the new directory needs a Makefile.am (copy the one from the tod subdirectory).  Add that new directory to the top-level configure.ac.

I'm happy to make the necessary build system changes if you push your branch to hpc4cmb/toast.  Then we can open a new pull request from their against master.  The build system will be made simpler soon.

zonca:
ok, renamed to `todmap`,
@tskisner I tried to add to the build system but didn't work, it gives "no target install" error.
Every time I submit a PR, I tick the "allow edit from maintainers" box so you have push access to my branch.

zonca:
actually maybe I fixed it, I had forgot the changes in `configure.ac`, retesting now

zonca:
ok, I think I fixed it, once Travis passes, this is ready to merge

keskitalo:
@zonca the tests seem to fail because they cannot import `OpSimPySM`. Perhaps they are still looking for it in the old location?

It is easy to run the tests on your own system with `'python3 -c "import toast.tests; toast.tests.run()"'`. This way you don't have to wait for Travis.

zonca:
thanks @keskitalo fixed now

keskitalo:
Looks great. I'll merge.

PR #209, opened 2018-02-02
Support for dipole simulation in satellite pipeline

zonca:
It also includes a bugfix related to pysm, see 03cfcad

@tskisner, @keskitalo I thought it was less error-prone to always run `get_submaps` instead of checking all the conditions where it should run, see e65919e, is this ok?

See example run with solar dipole only:

soldiptest

### DIFF
--- a/pipelines/toast_satellite_sim.py
+++ b/pipelines/toast_satellite_sim.py
@@ -212,6 +212,19 @@ def main():
                         help='Apply beam convolution to input map with gaussian '
                         'beam parameters defined in focalplane')
 
+    parser.add_argument('--input_dipole', required=False,
+                        help='Simulate dipole, possible values are '
+                        'total, orbital, solar')
+    parser.add_argument('--input_dipole_solar_speed_kms', required=False,
+                        help='Solar system speed [Km/s]', type=float,
+                        default=369.0)
+    parser.add_argument('--input_dipole_solar_gal_lat_deg', required=False,
+                        help='Solar system speed galactic latitude [degrees]',
+                        type=float, default=48.26)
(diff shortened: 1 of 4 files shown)


zonca:
@keskitalo if I run solar-only I get a dipole of about +-3.2mK, orbital-only +-0.26mK, if I set total I get +-4.6mK, I expected some effect from relativistic addition, but isn't this too much?

keskitalo:
@zonca that definitely sounds like too much. The dipole facility in TOAST has not been used much because toast-planck has its own. Would you be able to take a look? At the very least, we should raise an issue.

zonca:
ok, thanks @keskitalo , I'll open a separate issue and investigate more.

zonca:
ok, added a fix for #210 and an additional unit test

zonca:
rebased on  `master`

pipelines/toast_satellite_sim.py
```diff
@@ -211,6 +206,19 @@ def main():
                         help='Apply beam convolution to input map with gaussian '
                         'beam parameters defined in focalplane')
 
+    parser.add_argument('--input_dipole', required=False,
+                        help='Simulate dipole, possible values are '
+                        'total, orbital, solar')
+    parser.add_argument('--input_dipole_solar_speed_kms', required=False,
+                        help='Solar system speed [Km/s]', type=float,
```
keskitalo: That is a typo. Not `[Km/s]` but `[km/s]`

tskisner:
Yes, thanks for the work fixing this!

zonca:
is this good to merge?

PR #212, opened 2018-02-05
Add qarray::to_position method

keskitalo:
`to_position` is a reduced version of `to_angles`. It skips the position angle calculation.

### DIFF
--- a/src/libtoast/math/toast_qarray.cpp
+++ b/src/libtoast/math/toast_qarray.cpp
@@ -1,6 +1,6 @@
 /*
-Copyright (c) 2015-2017 by the parties listed in the AUTHORS file.
-All rights reserved.  Use of this source code is governed by 
+Copyright (c) 2015-2018 by the parties listed in the AUTHORS file.
+All rights reserved.  Use of this source code is governed by
 a BSD-style license that can be found in the LICENSE file.
 */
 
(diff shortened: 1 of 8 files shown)


src/libtoast/math/toast_qarray.cpp
```diff
@@ -794,3 +794,47 @@ void toast::qarray::to_angles ( size_t n, double const * quat, double * theta,
...
+            norm = 1.0 / ::sqrt ( norm );
+            qtemp[0] = quat[qf] * norm;
+            qtemp[1] = quat[qf+1] * norm;
+            qtemp[2] = quat[qf+2] * norm;
+            qtemp[3] = quat[qf+3] * norm;
+
+            toast::qarray::rotate ( 1, qtemp, 1, zaxis, dir );
+
+            theta[i] = toast::PI_2 - ::asin ( dir[2] );
+            phi[i] = ::atan2 ( dir[1], dir[0] );
```
zonca: what about doing `atan2( -dir[1], -dir[0] ) + pi` to avoid branching? https://stackoverflow.com/questions/1311049/how-to-map-atan2-to-degrees-0-360

keskitalo: @zonca if we are optimizing this method, I'd take a hard look at the need to normalize the supplied quaternions. For now, it is enough to repeat what is done in to_angles.

PR #213, opened 2018-02-06
Replace timing with TiMemory

jrmadsen:
- It says 171 files were changed but many of these were the deletion of cereal subpackage
- In general, this is all the PR is: 
    - sed -i '/TOAST_AUTO_TIMER/d' ...
    - sed -i 's/import timing/import timemory/g' ...
    - sed -i 's/timing.auto_timer/timemory.auto_timer/g'
    - sed -i 's/timing.timing_manager/timemory.timing_manager/g'


keskitalo:
I cannot test this on my laptop today, because it hangs trying to fetch the example data. Having access to NERSC should not be a prerequisite for compiling TOAST. I'll open an issue.

tskisner:
It's great to see timemory exist as a reusable package.  I branched from this branch in order to make some small interface tweaks- using the provided decorators and also providing a thin wrapper around timemory so that we can gracefully keep running if timemory is not found.  I also wrapped the decorators so that they work both on class instance methods as well as functions (using inspect).  I did not bother handling class or static methods since we don't use those anywhere currently.  For example:

https://github.com/tskisner/toast/blob/timemory-opt/src/python/healpix.py#L41

and

https://github.com/tskisner/toast/blob/timemory-opt/src/python/healpix.py#L97

This is still a work in progress:

https://github.com/tskisner/toast/blob/timemory-opt/src/python/timing.py

In particular, I was trying to wrap the argparsing and realized that this is a larger issue of controlling toast internal package variables.  So I'm working now on moving that bit into a separate file.  I'll open a pull request soon against this branch for further review from everyone.

In addition to the tedious switching to decorators, I also removed them in places where they did not make sense (trivial functions used mainly in auxiliary scripts) and added them in a couple places that were missing. 

This post is just to communicate things in progress.

jrmadsen:
I see that “inspect" is part of the Python standard library so I would encourage that this extension be put into the base TiMemory code instead of being custom to TOAST. Feel free to fork and create a PR.

A fallback creates more problems that it is worth, in my opinion. It would add a whole new set of tests to the CI because at some point, a feature of TiMemory will be used that isn’t in the fallback and it will break TOAST. It is a pip-installable package.

I can say, unequivocally, that any system that can compile TOAST can pip-install TiMemory. I have got to great pain to ensure this is the case. My CI is 36 variants of Python, Linux, Windows, GCC, Clang, and MPICH/OpenMPI and CDash for the Intel compilers is in the works.

If a user does not want the timing, they can simply disable it. I would strongly encourage y’all to reconsider implementing this fallback. 

And as a final aside, a fallback would significantly reduce the expose of my package and create a situation where TOAST users are not even aware of it’s availability. This is something I care about. It is not like it hasn’t already be noticed: when I first added this to TOAST, shortly thereafter we had an email from Yuji Chinone who stated he thought it would be useful. It is not like it is something that only TOAST is interested in.

Jonathan

—

Jonathan R. Madsen
Lawrence Berkeley National Laboratory
email: jrmadsen@lbl.gov
work: +1 (510) 486-7658

jrmadsen:
As an example on why the fallback creates more continuous integration testing situations, there is already an issue in the fallback with the word "elapsed" misspelled as "elpased". Clearly a simple and trivial typo but it demonstrates why new tests would need to be added.

https://github.com/tskisner/toast/blob/56dbabc7b5b973abd22d5bb69732cf21b8fe405e/src/python/timing.py#L111

tskisner:
As the commits to that branch state it is a work in progress [WIP].  No pull request is open on that branch yet.  

Currently our hard dependencies are very minimal (C++11, lapack, mpi4py, fftw/mkl, python3, healpy).  All other packages we use are optional dependencies which simply disable certain features if they are not available.  Certainly timemory will be installed in our standard stack at NERSC, and on larger collaboration workstations / clusters.  It will be in our dependency install scripts that people use, and it will be in the documentation as one of the things people should install to gain access to all the features in toast.  Currently this project has had no formal policy on how to decide whether something should be a hard dependency, except that "less is better".  I'll raise that issue with @keskitalo and @jdborrill .

tskisner:
Just opened #217 for reference.  That PR will likely never get merged directly, but I wanted to illustrate the design ideas / requirements that I had in mind.  Even if we make timemory a hard requirement, the 3 items listed there are still the desired default behavior we are aiming for in toast.

zonca:
this PR has changes all over the package, I think we should give high priority in reviewing and merging (or rejecting). @keskitalo @tskisner please let me know if you want me to do some testing of this branch by running the examples.

tskisner:
Hello @zonca, I had discussion with @keskitalo about this a couple weeks ago.  I am going to merge a subset of that dead PR into here first and then yes, we will need a big campaign of review and testing.  I am partially offline today, but will work on it this evening.  I agree that we should close out this work.

jrmadsen:
The entire purpose of this PR was to separate the transition into two stages for the sole purpose of avoiding this current situation because it does change so much. This PR was supposed to be simple and not require extensive testing because all this PR was simply removing the source code from TOAST and replacing timing.XYZ with timemory.XYZ. The second PR was intended to actually address changing the code to context managers and decorators and require more extensive testing. There was no supposed to be PRs against this PR. That was supposed to happen in the next PR.

tskisner:
The challenge with this PR as it stands is that it introduces a hard requirement on the external timemory package.  @keskitalo and I agreed that this is not desired.  Also, the "end goal" behavior is:

1. By default, no timing is done and nothing is written to stdout or extra files.

2. With a simple option (e.g. --toast-timing-enable), user pipelines can enable internal timing with reasonable defaults.

3. The user can supply other options to customize the output or type of information that is collected / reported.

This desired behavior is something we punted on when this code was first added, because we were in crisis mode before the large run in December.  Now it is time to address that.

So at the very least this PR needs a try / except block in a central place, testing the import of timemory and providing simple stubs in place of the functions.  Which is what I was doing here:

https://github.com/tskisner/toast/blob/timemory-opt/src/python/timing.py

I agree that formally one could put stubs in for the normal timing functions and then switch to decorators later.  However that basically means going back through all the source twice.   Also, the toast package needs a centralized place for "global package options" (including timing).  That's why I was moving towards a central location for this:

https://github.com/tskisner/toast/blob/timemory-opt/src/python/control.py

It sounds like further discussion is needed, so I'll get in touch with @keskitalo and @jdborrill .

jrmadsen:
I do not have any input or argument with how the timing is handled. That is y'all's prerogative. 

In most software developments, the master branch is more-or-less a beta/release-candidate of the upcoming release and it is recommended that users checkout the tagged/release versions as there are where issues like whether or not TiMemory is a hard-requirements have been properly resolved. If TOAST is deviating from this norm, then fair enough, but if not, this should be merged and the hard-requirement issued can be addressed further down the road before the next release

tskisner:
We take a more "continuous integration" approach where master should always be deployable.  I'm going to investigate the minimal set of changes needed to feel comfortable merging this branch without diverging from our longer-term goals.

PR #217, opened 2018-02-08
Global package configuration and timemory possibilities

tskisner:
This branch should not be merged without discussion.  This branch does several things:

- Adds a package-wide function for configuring internal parameters through
      the parsing of command line arguments to user pipelines.

- Makes use of timemory decorators.

- Allows the code to work if timemory is not available, by alternatively
      providing stubs which do nothing.

- Converts the example templates to use the timing-related internal
      command line options.

- Cleans up some whitespace issues.

There are other possible ways of handling the functionality in timing.py and
control.py.  The required features are:

1.  By default, no timing is done and nothing is written to stdout or 
        extra files.

2.  With a simple option (e.g. --toast-timing-enable), user pipelines can
        enable internal timing with reasonable defaults.

3.  The user can supply other options to customize the output or type of
        information that is collected / reported.

If we choose to make timemory a required dependency, then that would simplify
timing.py or eliminate the need for it altogether.  However, the centralized
control of timemory options in order to implement the 3 points above is still
needed (and is useful for other package-wide configuration in the future).

Although I did attempt to make the example pipelines work with the new command line arguments, I stopped that work until we can have further discussion on the structure of stuff in timing.py.  For example, if timemory is a hard dependency, then the only critical thing in timing.py is the decorator wrappers to infer the function type.  That could be merged into upstream timemory and eliminate the need for it here.

There were also a couple of typos fixed here that I will pull out into a separate pull request.


tskisner:
Just to further clarify- Ignore all the changes in this PR for now, *except* the addition of this one file:

https://github.com/tskisner/toast/blob/timemory-opt/src/python/control.py

This is a central place for control of package-wide parameters through the use extra command line arguments.  This is a generalization of what Jonathan was doing before (we can use it for other options that are not timing related).  If timemory is a hard dependency (i.e. TOAST will not run without it), then we just need to remove the "if use_timemory".  The critical feature I am aiming for is this centralized place for control of package options, and the behavior in the 3 points above.

All the rest of the changes in this branch is tedious fluff trying to get the unit tests and examples working.  But obviously that is silly to fix until we converge on the what the target interface is.

PR #218, opened 2018-02-09
Update the fake focalplane generation script

tskisner:
Update the fake focalplane generation script to match the new internal API.  Also reduce the size of the "tiny" ground simple example so that it will fit into RAM on reasonable laptop (use 7 pixels instead of 61)


tskisner:
@keskitalo , any comments before I merge?

keskitalo:
Merge away.

Issue #97, opened 2017-07-11
Map domain distribution is tedious


tskisner:
The use of the DistPixels class and externally calculating the locally hit pixels is "clunky".  We should provide a more intuitive way of just getting the already-distributed hits and noise covariance without lots of boilerplate.

tskisner:
To expand on this.  We should have a class that separately represents the global --> local pixel mapping, but does not store any distributed data array.  Then we pass an instance of that class into the distributed pixel constructor.

tskisner:
This is implemented in the toast3 branch by the `PixelDistribution` and `PixelData` classes.  Operators which generate pointing can optionally create a `PixelDistribution` instance which stored in the data.  All other `PixelData` instances use this.

PR #219, opened 2018-02-09
Move custom utils from pipelines into the library

zonca:
I am preparing my work on configuration files.

I think it would be easier if first I simplify a bit the pipeline scripts.
So as suggested some time ago by @tskisner I would like to move some of the custom utility functions into the TOAST library.

I just started from one and I'd like some feedback before I go ahead. Is this ok? Is `toast.Data` the right place for functionality related to looping through observations?

### DIFF
--- a/pipelines/toast_ground_sim.py
+++ b/pipelines/toast_ground_sim.py
@@ -751,28 +751,6 @@ def get_submaps(args, comm, data):
     return localpix, localsm, subnpix
 
 
-def add_sky_signal(args, comm, data, totalname_freq, signalname):
-    """ Add previously simulated sky signal to the atmospheric noise.
-
-    """
-    if signalname is not None:
-        autotimer = timing.auto_timer()
-        for obs in data.obs:
-            tod = obs['tod']
-            for det in tod.local_dets:
-                cachename_in = '{}_{}'.format(signalname, det)
-                cachename_out = '{}_{}'.format(totalname_freq, det)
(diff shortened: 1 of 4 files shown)


tskisner:
I think this is the right direction @zonca, and I see you are also working towards addressing #97, which is great!  Let's see what feedback @keskitalo has.

zonca:
I have a suggestion:
what about I create a new thin object that holds no data, but it is just a reference to all of the TOD objects then we implement `__add__` and `__mult__` and so on:

```
class TODReference:
    def __init__(self, data, key):
        self.data = data
        self.key = key
    def __add__(self, other):
        for ob in self.data.obs:
            # loop through dets
                 #sum tod of self.key and other.key
        
toast.Data["signal"] returns a TODReference with the "signal" key.

# So I can do:

toast.Data["total"] = toast.Data["signal"] + toast.Data["dipole"]

```

zonca:
what about my suggestion of implementing `TODReference`? Otherwise let me know asap and I'll just copy over @keskitalo's code and tweak it a bit.

tskisner:
I have no preference here.  Whatever you both think would work best.

zonca:
@keskitalo what about my suggestion of implementing `TODReference`? Otherwise let me know asap and I'll just copy over your code and tweak it a bit.

keskitalo:
I like it. A nice way of implementing TOD arithmetic. We'll want scalar multiplication as well.

PR #220, opened 2018-02-12
PySM precomputed cmb

zonca:
PySM supports reading a precomputed CMB map using the `c2` CMB model.
However, you cannot specify which map to read, it always reads `lensed_cmb.fits` from the templates folder.
Until now we have been installing PySM from my branch so I just overwrote that file.
However PySM has merged all my changes so I'd like to stop using my branch.

Therefore I have implemented support in TOAST for accepting a input path for the precomputed CMB and create a PySM model with that.


tskisner:
Looks good to me...

zonca:
ok, we can merge it now, then I'll do another pull request to install a tagged release of PySM once they make one https://github.com/bthorne93/PySM_public/issues/19

Issue #223, opened 2018-03-04
Timestamp repeats across observation


zonca:
I have hit a small issue on timestamps simulating PICO timelines for calibration work with @ziotom78 , it looks like the last sample of one observation and the next sample of the next observation have the same timestamp.

Last 2 samples of the first observation are:

    86399.99206349,  86400. 

The first 2 of the next:

     86400.        ,  86400.00793651

We have an interval of 0 between observations, is this the expected behavior?

tskisner:
No, that is not the expected behavior.  I starting place to look is in the helper function used to compute those time ranges:

https://github.com/hpc4cmb/toast/blob/master/pipelines/toast_satellite_sim.py#L327
https://github.com/hpc4cmb/toast/blob/master/src/python/tod/sim_interval.py#L15

This takes the requested time ranges and then creates actual time spans that are a whole number of samples.

zonca:
ok, I reproduced it in the simplest possible case: https://gist.github.com/zonca/6ace24831ad6fd0fb742cb426880d85c

Next I'll work on fixing it.

PR #226, opened 2018-03-06
[WIP] fix wrong num samples in interval

zonca:
@tskisner I think I found the issue causing #223, this fix is not complete, but wanted to ask for feedback.

Num of samples per interval was computed as:
    
```
duration = 3600
rate = 1
int(0.5 + duration * rate) + 1
```

But this gave 3601 samples in 1 hour at 1 Hz, so I think I should remove the `+1`.

However I am concerned about the following part of the code:

```
# Compute the actual time span for this number of samples                                                 
tottime = (totsamples - 1) * invrate
```

I would say that `tottime` is just `totsamples * invrate`, are you making any different assumption here?

### DIFF
--- a/src/python/tod/sim_interval.py
+++ b/src/python/tod/sim_interval.py
@@ -40,8 +40,8 @@ def regular_intervals(n, start, first, rate, duration, gap):
 
     # Compute the whole number of samples that fit within the
     # requested time span (rounded to nearest sample).
-    totsamples = int(0.5 + (duration + gap) * rate) + 1
-    dursamples = int(0.5 + duration * rate) + 1
+    totsamples = int(0.5 + (duration + gap) * rate)
+    dursamples = int(0.5 + duration * rate)
     gapsamples = totsamples - dursamples
 
     # Compute the actual time span for this number of samples
--- a/src/python/tod/sim_interval.py
+++ b/src/python/tod/sim_interval.py
@@ -42,19 +42,16 @@ def regular_intervals(n, start, first, rate, duration, gap):
     # requested time span (rounded to nearest sample).
     totsamples = int(0.5 + (duration + gap) * rate)
     dursamples = int(0.5 + duration * rate)
-    gapsamples = totsamples - dursamples
-
-    # Compute the actual time span for this number of samples
-    tottime = (totsamples - 1) * invrate
-    durtime = (dursamples - 1) * invrate
-    gaptime = tottime - durtime
 
     intervals = []
 
     for i in range(n):
         ifirst = first + i * totsamples
         ilast = ifirst + dursamples - 1
-        istart = start + i * tottime
+        # The time span between interval starts (the first sample of one interval
+        # to the first sample of the next) includes the one extra sample time.
+        istart = start + i * (totsamples * invrate)
+        # The stop time is the timestamp of the last valid sample (thus the -1).
         istop = istart + (dursamples - 1) * invrate
         intervals.append(Interval(start=istart, stop=istop, first=ifirst, last=ilast))
 


tskisner:
Here is an example:  assume the sample rate is 1Hz.  Then invrate is 1 second.  Now assume you have 5 samples.  That should give me:
```
first = 0
last = 4
start = 0.0
stop = 4.0
```
So there are (nsamples - 1) time spans between the first and last samples.  This function takes any time span (```duration```) as input, which is not necessarily evenly divisible by the sample delta time.  To continue this example, let's say we call the function with:
```
regular_intervals(n=5, start=0.0, first=0, rate=1.0, duration=4.5, gap=0.0)
```
In this case, I would expect the code to find the maximum number of samples that fit within ```duration+gap``` and then use that as the number of whole samples in each interval.  Looking at the code, I think there are 2 mistakes.  In computing the samples, I rounded instead of taking a floor.  And then later, when computing the timespan between interval starts, I did not include the extra additional sample time.  How about this change:
```
# Compute the whole number of samples that fit within the
# requested time span.
totsamples = int((duration + gap) * rate)
dursamples = int(duration * rate)

intervals = []

for i in range(n):
    ifirst = first + i * totsamples
    ilast = ifirst + dursamples - 1
    # The time span between interval starts (the first sample of one interval
    # to the first sample of the next) includes the one extra sample time.
    istart = start + i * (totsamples * invrate)
    # The stop time is the timestamp of the last valid sample (thus the -1).
    istop = istart + ((dursamples - 1) * invrate)
    intervals.append(Interval(start=istart, stop=istop, first=ifirst, last=ilast))
```
Caveat- I did not actually test the above code!  Just seeing what you think about that change.

zonca:
I don't know, I need to do testing, I'll get back to you with some results

zonca:
@tskisner first a more general question, why do we need `ifirst` and `ilast`? aren't `istart` and `istop` enough? I find it confusing.

I think it works as expected for divisible duration:

```
regular_intervals(n=2, start=0, first=0, rate=1, duration=4, gap=5)
[<Interval 0.0 - 3.0 (0 - 3)>, <Interval 9.0 - 12.0 (9 - 12)>]
```

but for non-divisible duration:

```
regular_intervals(n=2, start=0, first=0, rate=1, duration=4.5, gap=5)
[<Interval 0.0 - 4.0 (0 - 4)>, <Interval 10.0 - 14.0 (10 - 14)>]
```

is this ok? or should we get:

    [<Interval 0.0 - 4.0 (0 - 4)>, <Interval 9.5 - 13.5 (10 - 14)>]

for non-divisible gap:

```
regular_intervals(n=2, start=0, first=0, rate=1, duration=5, gap=4.5)
[<Interval 0.0 - 4.0 (0 - 4)>, <Interval 10.0 - 14.0 (10 - 14)>]
```

same here, should we get this one instead, or the result above is fine?

    [<Interval 0.0 - 4.0 (0 - 4)>, <Interval 9.5 - 13.5 (10 - 14)>]

tskisner:
Hi @zonca, the orignal reason for having both the sample ranges and the timestamps in the Interval class was the following:

1.  Usually when considering "events" in the timestream (cooler cycling, turnarounds, etc) we have timestamps from some housekeeping source.  We have a time range that we want to describe by an "Interval" object.  The start and stop times do not correspond to detector samples.

2.  Doing a lookup from time stamps to detector samples might be expensive (say if we have to read timestamps from disk.  If we are going to be using an Interval object multiple times, it is nice to do the time --> sample calculation once and store the sample indices corresponding to the time range.

For this particular case, we are simulating intervals at a fixed sample rate using generated times, so it would be easy to recompute the sample indices.  However, I don't think that is always the case for real data.

Back to your test:  I extracted the code above into a standalone test case [test_intervals.zip](https://github.com/hpc4cmb/toast/files/1832215/test_intervals.zip).  I think we are just missing a "one" in the sample calculation.  Running it with various values of duration and gap:
```
$ python3 test_intervals.py 5.0 0.0
first = 0, last = 5, start = 0.0, stop = 5.0
first = 6, last = 11, start = 6.0, stop = 11.0

$ python3 test_intervals.py 4.5 0.0
first = 0, last = 4, start = 0.0, stop = 4.0
first = 5, last = 9, start = 5.0, stop = 9.0

$ python3 test_intervals.py 4.5 5.0
first = 0, last = 4, start = 0.0, stop = 4.0
first = 10, last = 14, start = 10.0, stop = 14.0

$ python3 test_intervals.py 5.0 4.5
first = 0, last = 5, start = 0.0, stop = 5.0
first = 10, last = 15, start = 10.0, stop = 15.0
```
So those numbers seem ok, right?

zonca:
pasted your code on gist: https://gist.github.com/zonca/e1fb7a5917651c91518c3d0ec1dfb890

I think this is wrong:

```
$ python3 test_intervals.py 5.0 0.0
first = 0, last = 5, start = 0.0, stop = 5.0
first = 6, last = 11, start = 6.0, stop = 11.0
```

if we start at 0, 1 Hz, duration is 5 seconds, we have 5 samples, 0,1,2,3,4  sample at 5s is the first of the second interval.

so:

```
$ python3 test_intervals.py 5.0 0.0
first = 0, last = 4, start = 0.0, stop = 4.0
first = 5, last = 10, start = 5.0, stop = 10.0
```

Let's first focus on understanding the behavior we want. I will write the code later.

do you agree with me? If so I'll consider the other cases next.

zonca:
replaced by #232

Issue #225, opened 2018-03-05
Toast examples using shifter


zonca:
I built a docker image with toast (Edison-gcc), it was super easy with @tskisner 's scripts (only issue is Ubuntu 17.04 is not maintained anymore, so switched to 17.10).
It would be nice to add a couple example slurm scripts using shifter in the `examples/` folder I can use to run on NERSC.

tskisner:
If you have done a "shifterimg pull" to get the image to cori or edison (the machines have different registries), then you just have to add the correct "image" slurm option at the top of the slurm script and then do "srun shifter ...."

However, using shifter is a super-advanced topic (at least with a custom image), so I'm not sure how we want to handle that.  I agree we should document it somewhere, but probably not critical for it to be in the main examples.  We do not even have standard shifter images available since they are a moving target.

zonca:
ok, thanks @tskisner 

@keskitalo can you please just dump an example slurm file from the megarun here for my usage? then we can close this issue

keskitalo:
Sure. Here you go:

#!/bin/bash -l
#SBATCH --reservation=mp107debug
#SBATCH --image=docker:tskisner/toast:2.2.5
#SBATCH --nodes=365
#SBATCH --account=mp107
#SBATCH --partition=regular
#SBATCH --constraint=knl
#SBATCH --core-spec=4
#SBATCH --time=08:00:00
#SBATCH --licenses=SCRATCH
#SBATCH --job-name=atmosphere_030_year1

ulimit -c unlimited

export OMP_NUM_THREADS=4
export OMP_PROC_BIND=true
export OMP_PLACES=threads

export PYTHONSTARTUP=""

date

srun -n 5840 -c 16 --cpu_bind=cores \
    shifter ./toast_ground_sim.py @general.par \
    --fp focalplanes/focalplane_030_19.pkl \
    --freq 30 \
    --schedule schedules/atacama_schedule_year1.txt,schedules/pole_schedule_1year_scan_per_day.txt \
    --groupsize 16 \
    --out "out/atmosphere_030_year1_hwprpm_0" \
    --weather weather/weather_Atacama.fits,weather/weather_South_Pole.fits \
    --hwprpm 0 \
    --scanrate 5 \
    --toast-output-dir "out/atmosphere_030_year1_hwprpm_0" \
    --toast-timing-fname "timing_report_main" \
    --skip_noise \
    --skip_hits \
    --atm_gangsize 4 \
[...]

PR #227, opened 2018-03-16
how to run shifter with the toast docker image

zonca:
addresses #225


zonca:
this is just some docs, @tskisner do you think it is useful to merge?

PR #230, opened 2018-03-29
Tidas updates and MPI finalization

tskisner:
Hello @keskitalo , this is mostly updates to the tidas.py interface and the associated unit tests.  However I also made small changes to ensure that MPI_Finalize is called at the end of all pipelines.  This is needed because we are initializing MPI ourselves, so mpi4py will not call finalize for us when python exits.

### DIFF
--- a/pipelines/toast_cov_invert.py
+++ b/pipelines/toast_cov_invert.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
 
 # Copyright (c) 2015-2017 by the parties listed in the AUTHORS file.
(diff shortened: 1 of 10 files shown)


tskisner:
Although the finalization here removes an MPI warning under OpenMPI on an ubuntu system, using this actually triggers a different warning on Cray systems:  "Attempting to use an MPI routine after finalizing MPICH".  So do not merge until this is sorted out.

tskisner:
@keskitalo said offline this was ok to merge.

PR #232, opened 2018-04-11
Fix simulation of regular intervals

### LINKED ISSUE #223: Timestamp repeats across observation

tskisner:
We always round down to the whole number of samples that fit within the requested time span.  Add more documentation about the intention of this function and how number of samples is calculated.

Based on some small tests, I believe this function behaves as I expect (and I have tried to clarify that expected behavior).

Essentially, the requested time spans for the "good" data and the total of the "good" + "gap" are treated as a goal, and then the whole number of samples that fits within that goal is used.  So if you ask for a total time (good + gap) of 6 hours, then each good+gap number of samples will be slightly less that this, so the "6 hours" will occur between the last sample of the previous interval and first sample of the next.

This was the intended behavior of this function- to make it easy to simulate some repeating intervals given timespans which were not a whole number of samples.  In addition, this should fix the repeating timestamp problem, which was a bug.  Closes #223

### DIFF
--- a/src/python/tests/intervals.py
+++ b/src/python/tests/intervals.py
@@ -38,7 +38,7 @@ class IntervalTest(MPITestCase):
         intrvls = regular_intervals(self.nint, self.start, self.first,
                                     self.rate, self.duration, self.gap)
 
-        goodsamp = self.nint * ( int(0.5 + self.duration * self.rate) + 1 )
+        goodsamp = self.nint * ( int(self.duration * self.rate) + 1 )
 
         check = 0
 
--- a/src/python/tod/sim_interval.py
+++ b/src/python/tod/sim_interval.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2015-2017 by the parties listed in the AUTHORS file.
-# All rights reserved.  Use of this source code is governed by 
+# All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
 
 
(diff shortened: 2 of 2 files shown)


zonca:
here is some testing:

```
In [3]: regular_intervals(n=2, start=0, first=0, rate=1, duration=4, gap=5)
Out[3]: [<Interval 0.0 - 3.0 (0 - 3)>, <Interval 9.0 - 12.0 (9 - 12)>]

In [4]: regular_intervals(n=2, start=0, first=0, rate=1, duration=4.5, gap=5)
Out[4]: [<Interval 0.0 - 4.0 (0 - 4)>, <Interval 10.0 - 14.0 (10 - 14)>]

In [5]: regular_intervals(n=2, start=0, first=0, rate=1, duration=5, gap=4.5)
Out[5]: [<Interval 0.0 - 4.0 (0 - 4)>, <Interval 10.0 - 14.0 (10 - 14)>]

In [6]: regular_intervals(n=2, start=0, first=0, rate=1, duration=3600, gap=0)
Out[6]: [<Interval 0.0 - 3599.0 (0 - 3599)>, <Interval 3600.0 - 7199.0 (3600 - 7199)>]
```

I think it is fine

PR #233, opened 2018-04-11
Multiple cleanups to satellite boresight pointing

### LINKED ISSUE #224: Pointing jump across observation
In PICO simulations for calibration work with @ziotom78 I have hit a issue in pointing.

As we already discussed I have a version of `toast_satellite_sim.py` that dumps timelines to FITS (https://github.com/zonca/toast/pull/1/files).

I noticed that across observations there is a significant jump in pointing, see the following plot of 200 minutes of THETA across the observation boundary.

theta

How do you suggest to debug this?

tskisner:
This work expands the unit tests for satellite boresight simulations.  It fixes a bug in the slewing of the precession axis.  It also adds a constructor option to the TODSatellite class to specify the starting global offset of the observation.  This offset is then passed to the calculation functions for the precession axis and the boresight quaternions.  This ensures that the phase of the boresight pointing is correct, even if there are gaps between observations.  Closes #224.

### DIFF
--- a/pipelines/toast_satellite_sim.py
+++ b/pipelines/toast_satellite_sim.py
@@ -360,6 +360,7 @@ def main():
             comm.comm_group,
             detquats,
             obsrange[ob].samples,
+            firstsamp=obsrange[ob].first,
             firsttime=obsrange[ob].start,
             rate=args.samplerate,
             spinperiod=args.spinperiod,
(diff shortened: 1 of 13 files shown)


zonca:
ok, the pointing jumps are gone, see another printout of the quaternions:

https://gist.github.com/zonca/43e808aa82c7e4a83300ca597aa2e189

thanks!

Issue #199, opened 2018-01-19
Expected memory usage in TOAST satellite sim


zonca:
In the PICO noise simulations, I am running with just 1 pixel and 1 MPI process per node and I can fit 3 days worth of data per node. I am using a configuration very similar to the tiny satellite example for everything else.
PICO is 1 KHz, so I expect 1 day of 1 pixel to occupy `1000*3600*24*8*2/1e9 = 1.4 GB`

so 3 days would be `4.2 GB`

I see madam allocating `6.1 GB` for TOD + `14 GB` pointing.

TiMemory reports `34 GB` current and `53 GB` peak.

I am just checking if you think all these numbers look reasonable, so I can use the metric of fitting between `4 GB` and `5 GB` of data per node to guess the initial size of jobs.

zonca:
@tskisner @keskitalo does this sound reasonable?

keskitalo:
@zonca while Madam reports TOD and pointing memory being allocated, the Toast OpMadam operator is set up so that the toast buffers containing the TOD, pixel numbers and pointing weights are deallocated when the data are moved into the Madam buffers (and then generated again when Madam finishes.

Your calculation above seems to indicate only two doubles per detector sample. Instead, there should be
* 1 double for signal
* 1 long integer for pixel number
* 3 doubles for I, Q and U pointing weights
* 1 char for detector flags
* 1 char for common flags
* **potentially** 4 double for detector quaternions
* **potentially** 4 doubles for the bore sight quaternions
* 1 double for time stamp

Since you are running a single detector simulation, the bore sight pointing and the timestamps can be a significant part of your memory overhead. You should check (possibly calling `cache.report()`) that the quaternions are purged from the cache once the pixel numbers and pointing weights have been generated.

zonca:
ok, thanks, I'll experiment more with the next PICO runs to make sure I understand how RAM is being used.

zonca:
I am running a full focal plane sim for PICO band 9 with 684 detectors.
For my first test I am simulating 1 observation of 6 hours on 138 nodes, just noise. It fails with memory error.

That is 5 detectors each node, so about 126*3600*6*8*5/1e9 = 0.1 GB of signal TOD. Even with all the overhead I don't see how this cannot fit.

@tskisner @keskitalo what do you recommend I do?

zonca:
See log in ` /global/cscratch1/sd/zonca/pico/full_focal_plane_noise/log`

keskitalo:
[...]

PR #236, opened 2018-05-01
Use buffered operations in all places where detector pointing is used.

tskisner:
This makes several changes:

- If pointing quaternions are not to be cached, then read the pointing directly, rather than caching it and then deleting it.

- Perform all operations with the pointing in buffered chunks, with a buffer size that will eventually be set in the control.py file to be added in a separate PR.

- Do not cache detector quaternions in the satellite example pipeline.

See comments in #199 with results on a low-sample rate sim.  The impact will be even larger with higher sample rates and fewer detectors per process (since every process has a copy of the boresight).

### DIFF
--- a/pipelines/toast_satellite_sim.py
+++ b/pipelines/toast_satellite_sim.py
@@ -8,6 +8,7 @@ from toast.mpi import MPI, finalize
 
 import os
 import sys
+import time
 
 import re
 import argparse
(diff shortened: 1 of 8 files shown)


tskisner:
The buffer size is a million samples, so there is no performance hit- the cost of the work in each iteration is much more than the cost of doing 100 loops (for example).

Ted

tskisner:
I forgot to fix up the unit tests, will do that.

zonca:
I merged #237, we can rebase this on `master` to check if unit tests pass

tskisner:
@zonca, sounds good, I'll do that now.

PR #237, opened 2018-05-01
Fix coordinates satellitesim pipeline

zonca:
@ziotom78 and I noticed that the satellite pipeline was in Ecliptic coordinates, while PySM and Dipole were in Galactic coordinates.
These 2 commits fix the issue, we have tested them in PICO simulations, see https://github.com/zonca/pico-simulations/issues/11

### DIFF
--- a/pipelines/toast_satellite_sim.py
+++ b/pipelines/toast_satellite_sim.py
@@ -461,7 +461,7 @@ def main():
                 keep_quats=True,
                 keep_vel=False,
                 subtract=False,
-                coord="G",
+                coord="E",
                 freq=0,  # we could use frequency for quadrupole correction
                 flag_mask=255, common_flag_mask=255)
         op_sim_dipole.exec(data)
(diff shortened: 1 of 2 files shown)


zonca:
I just realized I also need to update all the expected results of the integration tests

zonca:
ok, I updated the expected map for satellite sim on https://github.com/hpc4cmb/toast-test-data/commit/6cc97c6d67eb3fc636a32d8d196644f83eb38dab and restarted travis

zonca:
@keskitalo I think the PySM coordinates issue also affects the ground simulations, unless you assume already that PySM is galactic

tskisner:
@zonca, this looks good for the satellite sims.  In principle we should add an option to all the TOD classes to specify what coordinate system to return the boresight and detector pointing in.  In fact, the TODGround class has such an option.  That is missing in TODSatellite.

For the ground sim pipeline I believe that:

- If ```--coord G``` is passed, then the TODGround quaternions will be in galactic, and so the pointing matrix will be in galactic, the dipole will be in galactic, and PySM is of course in galactic.

- If ```--coord C``` (or E) is passed, the PySM will still be in galactic and all others would be in the other system.

This is a small change to the ground example pipeline to pass the coordinate system into this modified PySM operator.

We should probably add a similar ```--coord``` option to the satellite script.

tskisner:
Hello @zonca, do think we should add coordinate transforms to TODSatellite and the satellite script command line options as part of this PR, or a separate one?  If we do it as a separate one, let's merge this and #236 before doing that.

zonca:
better do separate, I'll merge this so we can rebase other PR's and have units tests passing again

PR #238, opened 2018-05-03
Add coordinate system options to TODSatellite class and example pipeline

tskisner:
Make things consistent between the satellite and ground simulation example pipelines.


tskisner:
Thanks @zonca!  Is this ok to merge now that tests pass?

PR #240, opened 2018-05-08
Add support for computing cross spectra

keskitalo:


tskisner:
Looking through the changes, the interface and algorithmic techniques look good.  I have not actually run any tests, but assume you have :-)  Note that this uses numpy.fft rather than our internal wrapper around MKL/FFTW.  If it becomes an issue we could always test which one is faster for typical FFT sizes.  However, I am guessing that computing the covariance is more expensive than the FFT.  The unit test failures are unrelated, so feel free to merge if you like and we can fix those separately.

zonca:
please do not merge Pull Requests with failing tests, if you think the failure is due to the integration tests setup, please assign me to review the problem.

zonca:
the satellite integration test is passing, but the ground not.

@keskitalo it looks like your changes modified the output of the ground tests. Is this expected?
if so can you update the examples?
I added some docs on how to do it here: https://github.com/hpc4cmb/toast/pull/245

PR #241, opened 2018-05-14
4pi beam convolution support (conviqt) for satellite pipeline

zonca:
Not ready for review yet, just added to more easily ask for feedback

### DIFF
--- a/pipelines/toast_satellite_sim.py
+++ b/pipelines/toast_satellite_sim.py
@@ -28,6 +28,20 @@ import toast.timing as timing
 
 from toast.vis import set_backend
 
+def extract_local_dets(data):
+    """Extracts the local detectors from the TOD objects
+
+    Some detectors could only appear in some observations, so we need
+    to loop through all observations and accumulate all detectors in
+    a set
+    """
+    autotimer = timing.auto_timer()
+    local_dets = set()
+    for obs in data.obs:
+        tod = obs['tod']
+        local_dets.update(tod.local_dets)
+    return local_dets
+
 def add_sky_signal(args, comm, data, totalname, signalname):
     """ Add signalname to totalname in the obs tod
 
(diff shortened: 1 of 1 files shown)


zonca:
it currently runs to completion on a simple test case, still figuring out if the output is reasonable

zonca:
testing in https://github.com/zonca/pico-simulations/issues/17

pipelines/toast_satellite_sim.py
```diff
@@ -243,6 +296,32 @@ def main():
                         help='Solar system speed galactic longitude[degrees]',
                         type=float, default=263.99)
 
+    # Conviqt beam convolution
+
+    parser.add_argument( "--run_conviqt", required=False, default=False,
+        action="store_true", help="Convolve map with full beam alms" )
+    parser.add_argument( "--conviqt_beamfile_alm", required=False, default="beam_DETECTOR_alm.fits",
```
zonca: should we instead have the path to the beam `alms` inside the focalplane pkl?

pipelines/toast_satellite_sim.py
```diff
@@ -243,6 +296,32 @@ def main():
...
+        help="Full beam alms filename, they should be defined in the same reference frame"
+        "both for A and B detectors, i.e. no rotation, not even for the polarization angle is"
+        "performed by TOAST."
```
zonca: @keskitalo about the reference frame, I would start now supporting just 1. which is currently used by Karl for PICO that defines all the beams in the same reference frame, see https://github.com/zonca/pico-simulations/issues/17

pipelines/toast_satellite_sim.py
```diff
@@ -124,6 +138,45 @@ def simulate_sky_signal(args, comm, data, mem_counter, focalplanes, subnpix, loc
...
+    local_dets = extract_local_dets(data)
+
+    for pattern in args.conviqt_beamfile_alm.split(','):
+        detectordata = []
+        for det in local_dets:
+            skyfile = args.conviqt_input_alm.replace('DETECTOR', det)
+            beamfile = pattern.replace('DETECTOR', det)
+            # polarization leakage
+            epsilon = 0.
+            psipol = 0. if det.endswith("A") else np.pi/2
```
zonca: @keskitalo assuming `psipol` 90 deg is just temporary. should we add this to the focalplane pkl?

zonca: or computing it from the detector quaternion?

keskitalo:
This PR is now superseded by #304

Issue #242, opened 2018-05-17
OpSimPySM does not convert coordinates properly for Q/U components


ziotom78:
As it is written now, `OpSimPySM.exec` converts PySM maps from Galactic to XXX coordinates using a rotator and a call to `healpy.get_interp_val` (line [193](https://github.com/hpc4cmb/toast/blob/master/src/python/todmap/pysm_operator.py#L195)). However, it does not recalculate the value of Q and U in the new coordinate system, using a call to `healpy.rotator.Rotator.angle_ref`.

I am not fully sure of how this should be implemented. I believe that the calculation should proceed along the following lines: if `q_map` and `u_map` are the Q and U maps, then
```python
# This is a *direct* rotator, from Galactic to whatever
rot = healpy.rotator.Rotator(coord=["G", self.coord])

# Add the angle calculated by angle_ref to the argument of the L parameter (complex number)
L_map = (q_map + u_map * 1j) * numpy.exp(1j * rot.angle_ref(theta_gal, phi_gal))

# Overwrite the Q and U maps with the correct values
q_map = numpy.real(L_map)
u_map = numpy.imag(L_map
```
where `L_map` is a map of the complex-valued L parameter reported in the [Wikipedia page about Stokes parameters](https://en.wikipedia.org/wiki/Stokes_parameters#Properties).

This should work for some definition of `theta_gal` and `phi_gal`. I'm not sure of this however, as I've just discovered the existence of `healpy.angle_ref`. Thoughts?

zonca:
thanks @ziotom78, do you have a test case at hand that we can use to check if the implementation is correct?
Otherwise I'll build one from one of the `healpy` test cases.

zonca:
also we can check the IDL implementation in HEALPix https://github.com/healpy/healpixmirror/blob/master/src/idl/misc/rotate_coord.pro

ziotom78:
Sorry, I have no test cases at hand. I realized this last Friday while discussing with a student of mine at the blackboard.

zonca:
thanks, it was missing a factor of 2:

    L_map = (q_map + u_map * 1j) * numpy.exp(1j * 2 * rot.angle_ref(theta_gal, phi_gal))

I ran IDL and compared this to `rotate_coord.pro`, it gets the same results within 1e-4.

I wrote the implementation for `healpy` in https://github.com/healpy/healpy/pull/447

zonca:
implementing a fix in #246

[...]

PR #246, opened 2018-06-12
Fix PySM operator polarization change of coordinate system

zonca:
Addresses #242

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,7 @@ __pycache__/
 /examples/*.out
 /examples/*.jpg
 /examples/*.png
+/examples/ref*
 
 # Build directories when using ./cmake/platforms/...
 /build-*
@@ -99,3 +100,10 @@ __pycache__/
 
 # Darwin Finder files
 *.DS_Store
+
+# CMake
+CMakeFiles
+CMakeCache.txt
+
+# Notebooks
+*.ipynb
--- a/examples/check_maps.py
+++ b/examples/check_maps.py
@@ -13,12 +13,17 @@ filenames = glob.glob(ref + "/**/*.fits", recursive=True)
 assert len(filenames) > 0, "No FITS files in the reference folder"
 for filename in filenames:
 
-    print("Compare", filename)
 
-    m = hp.read_map(filename.replace(ref, folder), verbose=False)
-    m_ref = hp.read_map(filename, verbose=False)
-    m_diff_std = (m - m_ref).std()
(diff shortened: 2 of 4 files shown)


zonca:
* I modified the `check_maps.py` script to check also polarization
* I updated the expected outputs for satellite
* `satellite` test is passing on my machine
* I probably need to update also the ground results, but better first wait for the issue caused by #240 to be sorted out.

zonca:
restarted build to check if it picks up `healpy` 1.12

zonca:
this is complete and ready for review
I also updated the pipelines tests, they also compare polarization.
This branch requires healpy 1.12.3 for the polarization rotation to work correctly.

ziotom78:
Andrea, I tested a bit `rotate_alm` and find that there are cases where it would be more useful than `Rotator.rotate_map`, as it properly preserves CMB TT/EE/BB spectra. What do you think about something like this:

```python
if self.comm.rank == 0 and self.coord != "G":
    # PySM is always in Galactic, make rotation to Ecliptic or Equatorial
    angles = hp.rotator.coordsys2euler_zyz(coord = ["G", self.coord])
    alms = hp.map2alm(full_map_rank0)
    hp.rotate_alm(alms, *angles)
    full_map_rank0 = hp.alm2map(alms, nside)
```

In this way one would preserve the shape of C_l spectra even in other coordinate systems. (Compare this with the [E/B mode mixing](https://github.com/healpy/healpy/issues/460) caused by `healpy.Rotator.rotate_map`.)

tskisner:
Did we ever resolve the conflict that @keskitalo saw with healpy 1.12 and the stand-alone libsharp plus python bindings?  I believe his use case was that he already had libsharp installed and then installed healpy 1.12 to the same prefix, leading it to try and link to the existing libsharp.  The libsharp bundled with healpy is newer / different than the stand-alone version.

tskisner:
If our PySM code requires both healpy-1.12 and libsharp, and it is impossible to install those simultaneously, then...  Isn't that a problem?  Or are you saying that installing first healpy then libsharp into a new installation prefix will work?  It seems like a mess, but I agree that it is something that needs to be sorted out upstream of toast.

It does not have to hold up this PR- I'll separately go try to install latest healpy / libsharp in our NERSC dependency stack.

zonca:
@tskisner this is ready to merge

tskisner:
Looks fine to me, merge away.

PR #248, opened 2018-06-12
modify sed -i to support OSX

### LINKED ISSUE #247: examples/run_tiny_tests.sh is not compatible with OSX sed command
There are slight differences between GNU `sed` and a BSD version of the same tool that ships with OSX. Presently `examples/run_tiny_tests.sh` requires a GNU version of the tool to be available.

zonca:
should fix #247 
I cannot test on OSX


keskitalo:
Works. I'm merging this.

Issue #211, opened 2018-02-04
Unit test failures for >1 process


tskisner:
Running the unit tests on a workstation with >1 process fails during the shmem tests.  This should be fixed.

tskisner:
Digging deeper on this, our MPI unit test running is not ideal in several ways.  The desired behavior is:

- All processes run each test concurrently (with a barrier in between)

- Only the root process should print informational unit test things (the things that are normally printed by TextTestRunner).

- Processes can write independently to stdout / stderr inside a test (useful for debugging).

- By default, the output should NOT be captured by the test runner.

tskisner:
Closed by #251

PR #249, opened 2018-06-16
Work on data I/O interfaces

tskisner:
This work is focused on data reading and writing in toast.  It includes the following:

 - Adding of SPT3G to the external dependency scripts.  Also
   some cleanup to those scripts and support for libbz2 in
   boost.

 - Rewrite of the TIDAS data classes to support exporting
   to multiple named groups within the same observations.
   Also more careful deletion of TIDAS objects, to ensure
   that the underlying C++ destructors are called.

 - Add new SPT3G classes for data I/O.  Since the storage
   format (frames) does not match the internal toast storage
   layout, data is loaded into memory at construction,
   modified in the cache, and then only written on export.

 - Modifications to the example simulation pipelines.  Also
   begin work on an example pipeline to load data from disk
   and make a map.

 - Still working on:  performance of spt3g write (switching
   to mpi4py Gatherv), selective load of subsets of
   observations (useful for interactive work), and the data
   load-and-map example.


tskisner:
FYI @keskitalo , I will be making some further small changes today, but they will not affect the overall structure of these changes (just performance improvements).  Feel free to review, but don't merge yet.

src/python/tod/interval.py
```diff
@@ -171,3 +171,45 @@ def exec(self, data):
...
         return
+
+
+def intervals_to_chunklist(intervals, nsamp, startsamp=0):
+    """Create a list of contiguous sample chunks from intervals.
+
+    Given a list of (possibly discontinuous) intervals, construct a
+    list of contiguous chunks of samples.  The chunks are defined between
+    the starting points of each interval.  An additional chunk at the
+    beginning and end are
```
tskisner: I'll fix this docstring in a future commit soon.

chinoney:
I'm not 100% sure I should comment it here. But anyway I have one quick and naive question on I/O of 3g file. I guess it depends on a readout technology, but at least the current dfmux system outputs 2 timestreams: I and Q. Does the new module/functionality in toast deal with both or only I? Ideally we read both and apply some operations to maximize the responsivity, then inject it into a map-maker. Or uMUX just outputs I component?

tskisner:
Individual experiments will definitely need customized TOD classes for their specific needs.  Those would be implemented (for example) in the toast-so package or the equivalent for Simons Array.  And we can’t implement those until there is a formal specification of what the data frames / schema is going to look like.  The work in this PR is a step towards that.  In particular trying to figure out how to handle the orthogonal data distribution models of toast (individual detector timestream based) and spt3g (frame based).

giuspugl:
what's the status of this PR? I've recently pulled from toast and run the test suite and two tests failed. namely test_export and test_io in toast.tests.tidas.TidasTest.

tskisner:
I have paused work on this PR for a moment and am focusing on #211 first.  This is because in the process of debugging the SPT3G unit tests and code, I was hit by #211, which prevents testing the unit tests on multiple processes.  I am systematically cleaning up all unit tests and ensuring that they work on (at least) 4 processes.

tskisner:
Ok @keskitalo this is ready for review.  Note that travis tests will not work until I've restored the tarballs of dependencies at NERSC (which I will do after the power failure is over).  I ran the full unit tests on 4 processes with 4 threads each on a cori haswell node with no errors.

Obviously the big changes are in the tod/tidas*.py and tod/spt3*.py files, along with associated tests.

Also small changes in MPI comm splitting on some systems in cases where groupsize == 1.  Other changes to external software build scripts.  Some small fixes to unrelated things discovered along the way.

pipelines/toast_ground_sim.py
```diff
@@ -1242,22 +1245,55 @@ def output_tidas(args, comm, data, totalname):
     if args.tidas is None:
         return
     autotimer = timing.auto_timer()
-    from toast.tod.tidas import OpTidasExport
+    from toast.tod.tidas import OpTidasExport, TODTidas
```
keskitalo: Should we put this import to the top of the file and inside a try-catch clause?

pipelines/toast_ground_sim.py
```diff
@@ -1510,8 +1546,9 @@ def main():
 
             if (mc == firstmc) and (ifreq == 0):
                 # For the first realization and frequency, optionally
-                # export the timestream data to a TIDAS volume.
+                # export the timestream data.
                 output_tidas(args, comm, data, totalname)
```
keskitalo: These functions will probably go into the pipeline toolbox once we get to the refactor.

pipelines/toast_ground_sim.py
tskisner: I don't think we should import spt3g or tidas unless we need to.  However, thanks for pointing out that this code does not make use of the tidas_available and spt3g_available variables from toast.tod.  Those should be checked (i.e. we already have a try/except check in place- we should just use the result).  I'll make that change shortly.

pipelines/toast_ground_sim.py
tskisner: Yes, the pipelines have huge amount of duplicated code that will all be ripped out and consolidated into the pipeline sub-package.  In particular the "toast_map.py" pipeline which just loads data and makes a map is not well organized.  Rather than copy and paste yet more things from the other pipelines into that one, I punted on making it better until we have done that reorganization.

tskisner:
I just restored the travis dependency tarballs.  After pushing the small changes above I believe the tests will run again.

keskitalo:
That is a deviation from PEP-8. Is the import so expensive that it is
necessary to keep it conditional?

tskisner:
Conditional imports seem to be a much debated area of PEP-8 :-)  Forcing all imports at the top essentially forbids conditional imports.  Apparently even the python standard library uses conditional imports:

https://mail.python.org/pipermail/python-list/2009-August/546986.html

Another good point- if an imported package requires environment variables to be set at import time, then you might want to import it after you have set those variables:

https://github.com/PyCQA/pycodestyle/issues/264#issuecomment-73584055

And of course if 2 packages have conflicting / duplicate symbols, you might be choosing to import one or the other based on some criteria.

I guess I also find it silly to import those I/O packages if we are not using them.  I do believe that all imports which are going to be used should be at the top in accordance with PEP-8.  In this case, I'm fine moving the imports to the top and using them if the package is available.  However, there may be times where violating PEP-8 makes good sense.

src/python/tests/spt3g.py
```diff
@@ -0,0 +1,551 @@
+# Copyright (c) 2015-2017 by the parties listed in the AUTHORS file.
```
keskitalo: Should be 2018-> ?

tskisner: Ok, I should probably recursively regex-check all the files.  Maybe pre-emptively update them to 2019 :-)

tskisner:
Travis tests passing with updated / restored dependency tarballs, however I notice that libmadam is not found inside the travis environment (so those tests were skipped).  I'll investigate that separately.  Merging this now.

PR #250, opened 2018-06-17
Support for applying a set of gains to timelines

zonca:
I am using this for @ziotom78 's pico calibration work.

A gain file is a FITS file with 1 extension per detector with columns TIME and GAIN, is this ok?


tskisner:
Sorry for the delay @zonca, the operator here seems very generic, so that is good.  And since we are modifying the data, it makes sense to use local_signal(), which caches the data on read.

This FITS format is fine for small tests (and it is only used in the pipeline, so doesn't matter as much).  However, I think we should be careful when we have thousands of detectors (and hence thousands of HDUs).  Although there is no longer a formal limit to the number of HDUs (see https://listmgr.nrao.edu/pipermail/fitsbits/2003-May/001303.html), I don't think this kind of use is normal for FITS...

Again, I am fine with using this FITS table format for now.  If we want to use FITS for this in the future with thousands of detectors, and if the table scheme breaks down, then we could do something like:

HDU 0 : empty primary image
HDU 1 : table with list of detector names
HDU 2 : table with common timestamps (are these the same for all detectors?)
HDU 3 : 2D image of floats with one row per detector and one column per value.

zonca:
thanks @tskisner for your feedback.
I'd rather define 1 format now that is also ready for larger focalplanes.

I thought that having 1 common timestamp for all detectors was a bit too strict, but it is probably reasonable, even if calibration is somehow generated with different timestamps per detector they could be re-sampled/interpolated to a common set of timestamps.

@tskisner in the format you suggest in HDU 3 the FITS standard allows to read in just the gains for a single detector at a time, right? instead this would not be possible if we store a column for each detector.

jdborrill:
Given the multiplexing, at least large subsets of detectors will have
common timestamps; perhaps that should be a requirement for detectors to be
in the same file.

Julian

tskisner:
Correct that for FITS image access, each row is contiguous in memory / disk, so we can easily just read a slice of the image for one detector.  Of course astropy fits by default just loads the whole thing (unless memmap is enabled).

As @jdborrill suggests, we could have a different FITS file for each group of detectors that is sampled synchronously.

I am open to other suggestions for how to store the detector names and timestramps, but I think that the calibration data should be packed into a 2D image if possible.  This is similar to how one would store the data in an HDF5 file:  a 2D dataset of values and separate datasets for the timestamps and detector names (or maybe have the names as attributes).  Anyway, FITS is fine for now.

zonca:
@tskisner I am back working on this, I will rebase later, I need some help with the unit test

src/python/tests/tod.py
```diff
@@ -133,3 +135,46 @@ def test_local_signal(self):
...
+
+        self.detnames = ['bore']
+        self.dets = {
+            'CH-1' : np.array([0.0, 0.0, 1.0, 0.0]),
+            'CH-2' : np.array([0.0, 1.0, 0.0, 0.0])
+            }
+
+        # generate input data of ones for each of the 2 channels and 10 timestamps
+        # from 1 to 10, no need for pointing
+
```
zonca: @tskisner I need to generate fake data of just ones for the 2 channels, what is the easiest way to do that? I don't need pointing information.

zonca: thanks, I postponed creating `tod/applygain.py` because I don't remember how you add a new file to the build system. can you please remind me?

zonca: also, with `TODCache` I get this error:

```
Traceback (most recent call last):
  File "/global/common/software/pico/zonca/toast_prefix_cori/lib/python3.6/site-packages/toast/tod/tod.py", line 1037, in __del__
    self.cache.clear()
AttributeError: 'TODCache' object has no attribute 'cache'
```

zonca:
rebased on master

src/python/tests/tod.py
tskisner: Just add the file to the Makefile.am in that directory (see where the other files are listed).  Then re-run autogen.sh, configure, etc.  I have not seen that error in the destructor before...  Having said that, I see that both the TOD base class and the derived TODCache class are trying to clear the cache in their destructors.  It should not harm anything to do that, but we only really need that in the base class destructor.

zonca: thanks, another thing, what is the key in the TODCache created by default by write? I tried with "signal" but didn't work, then I tried to track it down into the code but couldn't find it.

tskisner:
Hello @zonca (replying here rather than the collapsed sub-thread from before rebase).  The TODCache class reads / writes data to cache fields / prefixes defined in the constructor:

https://github.com/hpc4cmb/toast/blob/master/src/python/tod/tod.py#L1027

zonca:
ok, in the last commit I moved to a new file, the build system picks it up and installs it:

`/global/project/projectdirs/planck/software/zonca/software/toast/src/python/tests/ops_applygain.py`

but the test runner cannot find it:

```
python -c 'import toast.tests; toast.tests.run("ops_applygain")'
OpenMP number of threads: 272
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/global/common/software/pico/zonca/toast_prefix_cori/lib/python3.6/site-packages/toast/tests/runner.py", line 117, in test
    suite.addTest( loader.loadTestsFromModule(sys.modules[modname]) )
KeyError: 'toast.tests.ops_applygain'
```

tskisner:
Did you add the test/ops_applygain.py file to the tests/Makefile.am list?  Did you import it at the top of runner.py and then load the suite further down runner.py?  We are not doing automatic discovery at this point, since we want to optionally just run one set of tests.

zonca:
thanks, I created a README with this info in  the tests folder for later reference.

This PR is ready for a final review and merge

tskisner:
Thanks @zonca, this all looks good.  Merge whenever you like

PR #251, opened 2018-07-09
Large refactor of unit tests to work correctly with multiple processes

tskisner:
This work refactors all unit tests with the following changes:

 - Googletest runner only prints test results on root process

 - Googletest has an MPI barrier in between tests and aborts
   if one process fails.

 - Python MPI test result and runner classes reimplemented to
   print overall status from all processes and ensure that a
   failure on one process results in a failure for that test.

 - Add test helper module for common operations like creating
   a toast communicator, distributed data objects, and
   focalplanes.

 - Refactor every single test to use these helper functions.

 - Fix or improve every unit test so that it works correctly
   with multiple toast groups, each with multiple processes.

Additionally, I have modified the travis file to attempt to run on 4 MPI processes.  This may not work, so expect several more pushes to sort that out.

As part of this work, I also found and fixed some small issues throughout the code, which were caught by the improved unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: python
 
 matrix:
   include:
-    # GCC 4.8
+    # GCC 4.8 - this is the default version for many distros
     - os: linux
       dist: trusty
       sudo: false
@@ -28,32 +28,6 @@ matrix:
             - pkg-config
       env:
         - MATRIX_EVAL="export CC=$(which gcc) && export CXX=$(which g++) && export FC=$(which gfortran)"
-    # GCC 4.9
-    - os: linux
-      dist: trusty
-      sudo: false
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-            - build-essential
-            - gcc-4.9
-            - g++-4.9
-            - gfortran-4.9
-            - autoconf
-            - automake
-            - m4
-            - libtool
-            - libhdf5-dev
-            - libmpich-dev
-            - mpich
-            - fftw3-dev
-            - cfitsio-dev
-            - libatlas-base-dev
-            - pkg-config
-      env:
-        - MATRIX_EVAL="export CC=$(which gcc-4.9) && export CXX=$(which g++-4.9) && export FC=$(which gfortran-4.9)"
     # GCC 5
     - os: linux
       dist: trusty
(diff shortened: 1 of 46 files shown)


tskisner:
Another note- I have disabled tidas unit tests in this refactor, because those tests have been separately refactored in #249.  I'll go and update them in that branch after this PR is merged.

tskisner:
I continue to see deadlocks on one unit test in travis that do not show up when running locally.  I'll attempt to duplicate this error locally using a travis container (https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally)

tskisner:
I've now manually run the commands in travis.yml inside the same docker container used by travis-ci for that failing test.  The tests pass on my local machine.

However, when going through this exercise, I see that our travis config is installing dependencies in a different way than we normally do.  It also does not install the dependencies for atmosphere simulation so all those unit tests are skipped.  I'm going to work on improving our travis setup locally and push updates to that once it is working.

tskisner:
Ok, I'm taking a new approach.  We compile many dependencies, which exceed both the run time and log length limits of travis.  Instead, I'm writing a standalone script that installs all the dependencies, echos periodically to keep travis happy, and on error or at the end simply tails the last few hundred lines to the log.  I'll put this with our other static install scripts and travis can wget it and run it.

After this works, we should also enable caching of this directory of built dependencies so that we don't have to take so long to run our tests.

tskisner:
This has now been rebased to squash all those debugging commits.  Just to review these most recent changes:

 - We use the travis virtual env rather than installing miniconda.  We then pip install as many things as possible (healpy, pysm, astropy, etc)
    
 - For compiled dependencies, we download and extract a tarball of pre-built software for each entry of our build matrix.
    
 - We run on only 2 processes (2 gangs of one process each) since travis does not seem to support oversubscription.
    
 - The actual test commands are wrapped in a script which echos things to prevent travis from timing out.
    
Additionally, there is now a README and script for updating the dependency tarball whenever versions of dependencies change.  This dependency tarball includes MPICH, so that we are sure to build it with the compiler in the build matrix.

@zonca and @keskitalo , I know this is a large PR to review.  It would be great if you could look in particular at:

1.  The updated travis file:
https://github.com/tskisner/toast/blob/issue_211/.travis.yml

2.  The new README for updating travis dependencies:
https://github.com/tskisner/toast/blob/issue_211/travis/README_travis

3.  The helper functions that tests can use:
https://github.com/tskisner/toast/blob/issue_211/src/python/tests/_helpers.py

4.  And any particular unit tests that you like and how I have modified them.  Mostly this was just to change the setup function to use the new helpers.  For example:
https://github.com/tskisner/toast/blob/issue_211/src/python/tests/ops_groundfilter.py

keskitalo:
This looks great!

PR #272, opened 2019-04-03
So sim prep

keskitalo:
Number of small changes to accommodate upcoming SO simulations.

### DIFF
--- a/pipelines/toast_ground_schedule.py
+++ b/pipelines/toast_ground_schedule.py
@@ -50,9 +50,18 @@ class Patch(object):
     current_el_min = 0
     current_el_max = 0
 
-    def __init__(self, name, weight, corners,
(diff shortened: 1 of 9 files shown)


tskisner:
These all look good.  I'll manually sort out the changes to the test runner file when merging into the 2.3 branch.  Looks like tod.py was formatted with black- which is fine.  Perhaps I should run that on all files in master and refactor_2.3 before trying to manually compare changes to see how they have diverged.  Are you using the default 88 character lines?  Just for my info so that I can use the same setting and add that to the standalone formatting script in the 2.3 branch.

Anyway, fine to merge and I'll be careful to not lose anything when merging things into my branch.

keskitalo:
The formatting is with default line length for black.

PR #278, opened 2019-07-10
Scan-synchronous signal simulation operator

keskitalo:


src/toast/tod/sss.py
```diff
@@ -0,0 +1,242 @@
...
+        self._power = power
+        self._report_timing = report_timing
+        return
+
+    @function_timer
+    def exec(self, data):
+        """Generate timestreams.
+
+        This iterates over all observations and detectors and generates
+        the atmosphere timestreams.
```
tskisner: Update docstring (atmosphere --> scan synchronous signal)

src/toast/tod/sss.py
```diff
@@ -0,0 +1,242 @@
...
+        counter1 = 0
+        counter2 = 0
+        return key1, key2, counter1, counter2
+
+    def _simulate_sss(self, key1, key2, counter1, counter2):
+        """
+        Create a map of the ground signal to observe with all detectors
+        """
+        npix = 12 * self._nside ** 2
+        sssmap = random(
```
tskisner: As long as the resolution of the ground signal map is always low this is fine.  If every process was allocating an NSIDE 4096 map this would be a problem.  Is the ground signal always smoothly varying?  Presumably this is a feature of the environment, independent of the telescope aperature and scan speed...

src/toast/tod/sss.py
```diff
@@ -0,0 +1,242 @@
...
+            prefix = "{} : {} : ".format(group, obsname)
+            tod = self._get_from_obs("tod", obs)
+            comm = tod.mpicomm
+            rank = 0
+            if comm is not None:
+                rank = comm.rank
+            site = self._get_from_obs("site_id", obs)
+            # weather = self._get_from_obs("weather", obs)
+
+            # Get the observation time span and initialize the weather
```
tskisner: There are several comments here about the weather object leftover from the atmosphere copy / paste.  In the future, we could imagine passing in a number to the constructor that is based on the site temperature.  Or if we have long observations and changing temperature we could pass in an array of temperatures which could be used during the "observe" step to scale the sampled ground map as a function of time.

src/toast/tod/sss.py
```diff
@@ -0,0 +1,242 @@
...
+                obsname = obs["name"]
+            except Exception:
+                obsname = "observation"
+            prefix = "{} : {} : ".format(group, obsname)
+            tod = self._get_from_obs("tod", obs)
+            comm = tod.mpicomm
+            rank = 0
+            if comm is not None:
+                rank = comm.rank
+            site = self._get_from_obs("site_id", obs)
```
tskisner: flake warning:  "site is assigned to but never used"  (leftover from atm copy / paste).

src/toast/tod/sss.py
```diff
@@ -0,0 +1,242 @@
...
+
+            # Get the observation time span and initialize the weather
+            # object if one is provided.
+            times = tod.local_times()
+            tmin = times[0]
+            tmax = times[-1]
+            tmin_tot = tmin
+            tmax_tot = tmax
+            if comm is not None:
+                tmin_tot = comm.allreduce(tmin, op=MPI.MIN)
```
tskisner: tmin_tot and tmax_tot are never used, I think we can wipe all of the tmin / tmax stuff.

keskitalo:
I'm merging this branch. It is now used in sotodlib.

PR #279, opened 2019-07-16
Fix calls to atmosphere utilities

tskisner:
This updates the ground sim pipeline to use the new interface to the absorption coefficient helpers.  This has not been tested.


keskitalo:
Solves the problem.

PR #280, opened 2019-07-31
Pipeline tools

keskitalo:
[WIP] Not ready for merge until the examples are updated.


pipelines/toast_ground_sim.py
```diff
@@ -508,28 +135,16 @@ def parse_arguments(comm):
...
         "--freq",
         required=True,
-        help="Comma-separated list of frequencies with " "identical focal planes",
-    )
-    parser.add_argument(
-        "--tidas", required=False, default=None, help="Output TIDAS export path"
-    )
-    parser.add_argument(
-        "--spt3g", required=False, default=None, help="Output SPT3G export path"
+        help="Comma-separated list of frequencies with identical focal planes",
```
zonca: is `--freq` asking TOAST to only consider a subset of the frequencies available in the focalplane?

pipelines/toast_ground_sim.py
```diff
@@ -508,28 +135,16 @@ def parse_arguments(comm):
...
     )
 
     try:
         args = parser.parse_args()
     except SystemExit:
-        return
+        sys.exit(0)
```
zonca: should we exit with 1? or what about just leaving the exception be raised to the top?

pipelines/toast_ground_sim.py
```diff
@@ -547,415 +162,173 @@ def parse_arguments(comm):
...
     """ Attach a focalplane to each of the schedules.
 
+    Args:
+        schedules (list) :  List of tuples of the form
+            (`site`, `all_ces`) where `site` is a Site
+            instansce with the `telescope` field filled
+            and `all_ces` is a list of CES objects.
+    Returns:
+        detweights (dict) : Inverse variance noise weights for every
+            detector across all focal planes.
```
zonca: are `detweights` unitless? if so let's state it

pipelines/toast_ground_sim.py
```diff
@@ -547,415 +162,173 @@ def parse_arguments(comm):
...
 
+    # Append a focal plane and telescope to each entry in the schedules list
     detweights = {}
-    for schedule, focalplane in zip(schedules, focalplanes):
+    for schedule, focalplane, telescope in zip(schedules, focalplanes, telescopes):
         schedule.append(focalplane)
-        for detname, det in focalplane.items():
-            net = det["NET"]
-            detweight = 1.0 / (args.samplerate * net * net)
+        schedule.append(Telescope(telescope))
```
zonca: `schedule` is made up of both `focalplane` and `Telescope` objects?

zonca: interleaved?

pipelines/toast_ground_sim.py
```diff
@@ -547,415 +162,173 @@ def parse_arguments(comm):
...
+        for detname, detdata in focalplane.items():
+            net = detdata["NET"]
+            detweight = 1.0 / (args.sample_rate * net * net)
             if detname in detweights and detweights[detname] != detweight:
                 raise RuntimeError("Detector weight for {} changes".format(detname))
             detweights[detname] = detweight
 
-    tmr.stop()
+    timer.stop()
     if comm.world_rank == 0:
```
zonca: can we have the check of rank 0 be inside `timer`?

pipelines/toast_ground_sim.py
```diff
@@ -547,415 +162,173 @@ def parse_arguments(comm):
...
-
-
-@function_timer
-def create_observation(args, comm, all_ces_tot, ices, noise):
+def create_observation(args, comm, all_ces_tot, ices, noise, verbose=True):
     """ Create a TOAST observation.
 
     Create an observation for the CES scan defined by all_ces_tot[ices].
 
     """
```
zonca: should we have also a description of all the arguments in the docstring?

pipelines/toast_ground_sim.py
```diff
@@ -547,415 +162,173 @@ def parse_arguments(comm):
...
-            firsttime=CES_start,
-            rate=args.samplerate,
-            site_lon=site_lon,
-            site_lat=site_lat,
-            site_alt=site_alt,
-            azmin=azmin,
-            azmax=azmax,
-            el=el,
-            scanrate=args.scanrate,
+            detranks=ndetrank,
```
zonca: `ndetrank` survived the renaming to more human understandable names

pipelines/toast_ground_sim.py
```diff
@@ -1896,129 +470,136 @@ def main():
...
+            tod = ob["tod"]
+            tod.free_radec_quats()
 
     # Prepare auxiliary information for distributed map objects
 
     _, localsm, subnpix = get_submaps(args, comm, data)
 
-    if args.input_pysm_model:
+    if args.pysm_model:
+        focalplanes = [s[3] for s in schedules]
```
zonca: using indices is kind of error prone, can we create a named tuple or simple class (dataclasses unfortunately are python 3.7 only)?

pipelines/toast_ground_sim.py
```diff
@@ -1896,129 +470,136 @@ def main():
...
                     )
                 )
 
-            copy_atmosphere(args, comm, data, mem_counter, totalname, totalname_freq)
+            # Make a copy of the atmosphere so we can scramble the gains and apply
+            # frequency-dependent scaling.
+            copy_signal(args, comm, data, totalname, totalname_freq)
 
-            scale_atmosphere_by_frequency(args, comm, data, freq, totalname_freq, mc)
+            scale_atmosphere_by_frequency(args, comm, data, freq, mc, totalname_freq)
```
zonca: switch order looks dangerous, what about using keyword arguments?

pipelines/toast_ground_sim.py
```diff
@@ -1896,129 +470,136 @@ def main():
...
 
             update_atmospheric_noise_weights(args, comm, data, freq, mc)
 
-            add_sky_signal(data, totalname_freq, signalname)
+            # Add previously simulated sky signal to the atmospheric noise.
+
+            add_signal(args, comm, data, totalname_freq, signalname, purge=(nmc == 1))
```
zonca: rename `nmc` to easier to understand name?

src/toast/pipeline_tools/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+        "--MC-count",
+        required=False,
+        default=1,
+        type=np.int,
+        help="Number of Monte Carlo noise realizations",
+    )
+    return
+
+
+@function_timer
```
zonca: what is the difference between the `@function_timer` decorator and the timers you explicitly instantiate inside?

src/toast/pipeline_tools/binning.py
```diff
@@ -0,0 +1,258 @@
...
+    """construct distributed maps to store the covariance,
+    noise weighted map, and hits
+    """
+    log = Logger.get()
+    timer = Timer()
+
+    if subnpix is None or localsm is None:
+        localpix, localsm, subnpix = get_submaps(args, comm, data, verbose=verbose)
+
+    npix = 12 * args.nside * args.nside
```
zonca: use `hp.nside2npix` instead?

src/toast/pipeline_tools/binning.py
```diff
@@ -0,0 +1,258 @@
...
+    data,
+    invnpp,
+    zmap,
+    detweights,
+    outpath,
+    cache_prefix=None,
+    prefix="binned",
+    verbose=True,
+):
+    """ Bin the signal in `cache_prefix` onto `zmap`
```
zonca: what is a `zmap`?

src/toast/pipeline_tools/madam.py
```diff
@@ -0,0 +1,461 @@
...
+from ..map import OpMadam
+
+
+def add_madam_args(parser, ground_data=True):
+    """ Add libmadam arguments
+
+    Args:
+        ground_data (bool) :  If true, assume that Madam will be used to
+            process ground experiment data and every process will have
+            approximately equal sky coverage.
```
zonca: explain quickly what this implies

src/toast/pipeline_tools/madam.py
```diff
@@ -0,0 +1,461 @@
...
+    else:
+        info = 3
+
+    if time_comms is None:
+        time_comms = [("all", comm.comm_world)]
+
+    if telescope_data is None:
+        telescope_data = [("all", data)]
+
+    ttimer = Timer()
```
zonca: what does `ttimer` stands for? maybe add only comment if variable name gets too long

src/toast/pipeline_tools/pointing.py
```diff
@@ -0,0 +1,156 @@
...
+        required=False,
+        default=0,
+        type=np.float,
+        help="The rate (in RPM) of the HWP rotation",
+    )
+    parser.add_argument(
+        "--hwp-step",
+        required=False,
+        default=None,
+        help="For stepped HWP, the angle in degrees of each step",
```
zonca: what about adding unit to argument name? i.e. `--hwp-step-deg`?

src/toast/pipeline_tools/todground.py
```diff
@@ -0,0 +1,496 @@
...
+
+
+class Telescope(object):
+    def __init__(self, name):
+        self.name = name
+        self.id = name2id(name)
+        return
+
+
+class Site(object):
```
zonca: no need to subclass `object` in Python 3

pipelines/toast_ground_sim.py
keskitalo: No, the frequency is used for scaling the atmospheric signal. It overrides the bandpass read from the focal plane.  In the SO sims we use the actual bandpasses.

pipelines/toast_ground_sim.py
keskitalo: The main() has special exception handlers that I don't want to trigger in this case. The argument parser already prints out all of the useful information.

pipelines/toast_ground_sim.py
keskitalo: In K_CMB^-2. I revised the docstring.

pipelines/toast_ground_sim.py
keskitalo: Telescope is just a short-hand to store the name and ID of a telescope. I think TOAST would benefit from a fully-fledged Telescope class (with Focalplane as a member) but that is beyond the scope of the pipeline refactor.

pipelines/toast_ground_sim.py
keskitalo: I suppose we could, but we might also be interested in the timings on individual processes.

pipelines/toast_ground_sim.py
zonca: can you add that to the argument description?

pipelines/toast_ground_sim.py
zonca: I meant for printing purposes, it automatically prints only on rank 0, so you can use it without if

pipelines/toast_ground_sim.py
keskitalo: Sure.

pipelines/toast_ground_sim.py
keskitalo: Should the Timer.report() method then take the process rank as an argument?

zonca: I was thinking about the constructor instead. How does `Timer` know which rank is it reporting about now?

src/toast/pipeline_tools/todground.py
keskitalo: Thanks. Doesn't hurt either.

src/toast/pipeline_tools/pointing.py
keskitalo: I like it. Also `--hwp-step-time` --> `--hwp-step-time-s`

src/toast/pipeline_tools/madam.py
keskitalo: Ted likes very short variable names. I replaced `timer` -> `total_timer` and `ttimer` with `timer`.

src/toast/pipeline_tools/madam.py
keskitalo: Done.

pipelines/toast_ground_sim.py
keskitalo: It does not. I would prefer deferring modifications to the Timer class to some other time.

zonca: Ok

src/toast/pipeline_tools/todground.py
zonca: It is a easy "replace with empty string" fix that removes code, I would do it

src/toast/pipeline_tools/binning.py
keskitalo: The distributed map. I revised the docstrings.

src/toast/pipeline_tools/__init__.py
keskitalo: The function timers can be collected and output in the end. The explicit timers only report when asked.

src/toast/pipeline_tools/binning.py
keskitalo: Sure

pipelines/toast_ground_sim.py
keskitalo: `nmc` -> `nsimu`

pipelines/toast_ground_sim.py
keskitalo: I like to prefix integer count variables with `n`. `detrank` is a well-established variable name in TOAST. The alternative would be `detranks` (as is used in the constructor) but, to me, that looks like a container of detranks, not their total number. For more verbosity, we could use `n_detrank` but it looks clunky. Anyway, short variable names are more acceptable when their lifetime is short (fits on a single terminal window).

pipelines/toast_ground_sim.py
keskitalo: Sure.

pipelines/toast_ground_sim.py
keskitalo: I completely reworked the Schedule, CES, Telescope, Site, Focalplane mess. It had evolved over a long time. Much cleaner now, thanks!

src/toast/pipeline_tools/binning.py
```diff
@@ -0,0 +1,290 @@
...
+        pass
+
+    # `nside` may already be added
+    try:
+        parser.add_argument(
+            "--nside", required=False, default=512, type=np.int, help="Healpix NSIDE"
+        )
+    except argparse.ArgumentError:
+        pass
+    return
```
zonca: empty return

src/toast/pipeline_tools/classes.py
```diff
@@ -0,0 +1,146 @@
...
+    return value % maxval
+
+
+class Focalplane:
+    detector_data = {}
+    _detweights = None
+    sample_rate = None
+    _radius = None
+    _detquats = None
+    _noise = None
```
zonca: not sure why there are class attributes, they are generally used only in very special circumstances.

src/toast/pipeline_tools/classes.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Args:
+            fname_pickle (str) :  Pickle file containing the focal
+                 plane dictionary
+            sample_rate (float) :  Default sampling rate for all
+                detectors.  Will be overridden by 'fsample' fields
+                if they exist for the detectors in the dictionary.
+            radius_deg (float) :  force the radius of the focal plane.
+                otherwise it will be calculated from the detector
+                offsets.
```
zonca: missing description of `detector_data`

src/toast/pipeline_tools/classes.py
```diff
@@ -0,0 +1,146 @@
...
+        """
```
zonca: do `self.detector_data={}` here

src/toast/pipeline_tools/classes.py
```diff
@@ -0,0 +1,146 @@
...
+        if detector_data is not None:
+            self.detector_data.update(detector_data)
+        if fname_pickle is not None:
+            with open(fname_pickle, "rb") as picklefile:
+                self.detector_data.update(pickle.load(picklefile))
+        self.sample_rate = sample_rate
+        self._radius = radius_deg
+        return
+
+    def refresh(self):
```
zonca: what about `reset_properties` instead of `refresh`?

src/toast/pipeline_tools/binning.py
```diff
@@ -0,0 +1,290 @@
...
+    """construct distributed maps to store the covariance,
+    noise weighted map, and hits
+
+    Args:
+        detweights (dict) :  A dictionary of the dimensional, inverse variance
+             detector weights that form the diagonal of N^-1.
+    Returns:
+        npp (DistPixels) :  The white noise covariance matrices needed to
+            bin signal onto the distributed map `zmap`.
+        zmap (DistPixels) :  An empty distributed map compatible with `npp`
```
zonca: what if we name `zmap` -> `dist_map` and `npp` -> `white_noise_cov_matrix`

src/toast/pipeline_tools/todground.py
keskitalo: Agreed.

src/toast/pipeline_tools/classes.py
```diff
@@ -0,0 +1,146 @@
...
+class Telescope(object):
+    focalplane = None
+    site = None
+
+    def __init__(self, name, focalplane=None, site=None):
+        self.name = name
+        self.id = name2id(name)
+        self.focalplane = focalplane
+        self.site = site
+        return
```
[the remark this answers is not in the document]
keskitalo: Removed

src/toast/pipeline_tools/classes.py
```diff
@@ -0,0 +1,146 @@
...
+                fknee=fknee,
+                alpha=alpha,
+                NET=NET,
+            )
+        return self._noise
+
+
+class Telescope(object):
+    focalplane = None
+    site = None
```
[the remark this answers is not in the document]
keskitalo: No good reason. Removed.

src/toast/pipeline_tools/binning.py
keskitalo: Removed

src/toast/pipeline_tools/classes.py
keskitalo: Removed.

src/toast/pipeline_tools/classes.py
keskitalo: Added.

src/toast/pipeline_tools/classes.py
keskitalo: Yes.

src/toast/pipeline_tools/classes.py
keskitalo: Works for me.

src/toast/pipeline_tools/binning.py
keskitalo: Looks a little clunky but definitely more descriptive!

zonca:
if useful I can give another pass of review on Tuesday

pipelines/toast_ground_sim.py
tskisner: Just a comment here about timers- you can also look at the C++ source for more details:

1.  The "Timer" class is a serial object.  If you only want to time events on rank zero, one could just create the timer on that rank.  Standalone "Timer" instances basically replace the functionality of using MPI_Wtime to get elapsed time on one process and print a message.

2.  Additionally each process has a singleton dictionary of timers (GlobalTimers https://github.com/hpc4cmb/toast/blob/refactor_2.3/src/libtoast/include/toast/sys_utils.hpp#L104) which can be started / stopped anywhere in the code.  The @function_timer decorator adds entries to this dictionary.  This global timer dictionary is also serial with no MPI awareness.

3.  When collecting timings across multiple processes, there are functions to do that:  https://github.com/hpc4cmb/toast/blob/refactor_2.3/src/toast/timing.py#L103

4.  Although one could define an "MPITimer" or "SynchronousTimer" class which took a communicator and then did a barrier and a "report" on only a single process, that is not currently implemented.  This would need to be a new additional class, since there are definitely use cases for plain serial timers even if MPI is enabled (for example if different processes are following different code paths and doing different things).  I don't object to defining such a class if you want.  This would be defined in libtoast_mpi with bindings in _libtoast_mpi.cpp.  You could also consider a thin wrapper statement in toast.timing that fell back to using a plain Timer in the case where MPI is disabled.

PR #281, opened 2019-08-06
Add elapsed_seconds method to Timer

keskitalo:
Certain use cases of the timer call for repeatedly checking if a given length of time has passed since the start of the timer. Currently this would require stopping the timer, calling Timer.seconds() and then starting the timer again. The stops-start would also increment the calls_ member of the timer.

Timer.elapsed_seconds() allows checking the elapsed time from a running timer without disturbing its state.


tskisner:
These changes look fine, and it's great to add methods like this for common use cases (like "report_clear").  Merge away after tests pass.

PR #283, opened 2019-08-16
Add utility function to set numba threading layer to OpenMP

tskisner:
Set the numba threading layer to OpenMP prior to importing PySM.

### DIFF
--- a/src/toast/utils.py
+++ b/src/toast/utils.py
@@ -36,7 +36,23 @@ from ._libtoast import (
 )
 
 
-def set_numba_openmp():
+def set_numba_threading():
+    """Set the numba threading layer.
+
+    For parallel numba jit blocks, the backend threading layer is selected at runtime
+    based on an order set inside the numba package.  We would like to change the
+    order of selection to prefer the OpenMP backend in order to be compatible with
+    other pieces of compiled code that use the OpenMP thread pool.  If OpenMP is not
+    supported, then we next try to use TBB for threading and finally fall back to the
+    default, which uses a multiprocessing workqueue.
+
+    Args:
+        None
+
+    Returns:
+        None
+
+    """
     have_numba_omp = False
     have_numba_tbb = False
     try:
--- a/src/toast/utils.py
+++ b/src/toast/utils.py
@@ -53,33 +53,25 @@ def set_numba_threading():
         None
 
     """
-    have_numba_omp = False
-    have_numba_tbb = False
-    try:
-        from numba.npyufunc import tbbpool
-
-        have_numba_tbb = True
-    except ImportError:
-        # no TBB
-        pass
-
+    log = Logger.get()
+    threading = "default"
     try:
         from numba.npyufunc import omppool
 
-        have_numba_omp = True
+        threading = "omp"
     except ImportError:
         # no OpenMP support
-        pass
+        try:
+            from numba.npyufunc import tbbpool
 
+            threading = "tbb"
+        except ImportError:
+            # no TBB
+            pass
     try:
         from numba import config, njit, threading_layer
 
-        if have_numba_omp:
-            config.THREADING_LAYER = "omp"
-        elif have_numba_tbb:
-            config.THREADING_LAYER = "tbb"
-        else:
-            config.THREADING_LAYER = "default"
+        config.THREADING_LAYER = threading
 
         # In order to get numba to actually select a threading layer, we must
         # trigger compilation of a parallel function.
@@ -89,17 +81,17 @@ def set_numba_threading():
 
         x = np.arange(10.0)
         y = x.copy()
-        foo(x, y)
-
-        # Log the layer that was selected
-        layer = threading_layer()
-        log = Logger.get()
-        log.info("Numba threading layer set to:  {}".format(layer))
+        out_py = x + y
+        out_numba = foo(x, y)
+        if not np.allclose(out_py, out_numba):
+            log.info("Numba results do not match python, not setting threading")
+        else:
+            # Log the layer that was selected
+            layer = threading_layer()
+            log.info("Numba threading layer set to:  {}".format(layer))
     except ImportError:
         # Numba not available at all
-        log = Logger.get()
         log.info("Numba not available, not setting threading layer")
-    return
 
 
 try:
--- a/src/toast/utils.py
+++ b/src/toast/utils.py
@@ -36,6 +36,29 @@ from ._libtoast import (
 )
 
 
+def set_numba_openmp():
+    try:
+        from numba import config, threading_layer
+
+        layer = None
+        try:
+            config.THREADING_LAYER = "omp"
+            layer = threading_layer()
+        except ValueError:
+            try:
+                config.THREADING_LAYER = "threadsafe"
+                layer = threading_layer()
+            except ValueError:
+                # Just give up at this point
+                config.THREADING_LAYER = "default"
+                layer = threading_layer()
+        log = Logger.get()
+        log.info("Numba threading layer set to:  {}".format(layer))
+    except ImportError:
+        pass
+    return
+
+
 try:
     import psutil
 
@@ -116,6 +139,7 @@ try:
         if comm is not None:
             comm.Barrier()
         return
+
 
 except ImportError:
 
--- a/src/toast/utils.py
+++ b/src/toast/utils.py
@@ -36,6 +36,8 @@ from ._libtoast import (
     vfast_erfinv,
 )
 
+numba_threading_layer = "NA"
+
 
 def set_numba_threading():
     """Set the numba threading layer.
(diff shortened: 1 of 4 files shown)


tskisner:
This work checks for supported numba threading backends and selects them in order of "omp", "tbb" and then "default".  conda packages for numba on linux support all 3, but pip wheels do not support omp.  The "default" threading layer uses multiprocessing workqueue constructs.  Note that this function needs to be called before the first numba jit function in the code.  Currently it is being called right before the pysm import, but if we ever decide to use numba elsewhere we could put the call in some more generic location.

zonca:
Thanks, can you add the explanation above into the docstring of the function?

tskisner:
Good point, and actually the function name is not quite right, since it selects the backend in a particular order, but not just the omp backend.

src/toast/utils.py
```diff
@@ -36,6 +36,56 @@
 )
 
 
+def set_numba_openmp():
+    have_numba_omp = False
```
zonca: what about doing `threading = "default"`,

then try omp and and set `threading="omp"` if it is available, in the expect block try with tbb.

finally `config.THREADING_LAYER = layer`

src/toast/utils.py
```diff
@@ -36,6 +36,56 @@
...
+
+        # In order to get numba to actually select a threading layer, we must
+        # trigger compilation of a parallel function.
+        @njit(parallel=True)
+        def foo(a, b):
+            return a + b
+
+        x = np.arange(10.0)
+        y = x.copy()
+        foo(x, y)
```
zonca: as we are running this anyway, we could as well throw an `assert` to make a quick test of the result.

src/toast/utils.py
```diff
@@ -36,6 +36,56 @@
...
+
+        # Log the layer that was selected
+        layer = threading_layer()
+        log = Logger.get()
+        log.info("Numba threading layer set to:  {}".format(layer))
+    except ImportError:
+        # Numba not available at all
+        log = Logger.get()
+        log.info("Numba not available, not setting threading layer")
+    return
```
zonca: empty returns are not necessary, they look pretty awkward for python programmers. even though I admit they serve the purpose of more easily understand where the function ends beside indentation.

src/toast/utils.py
tskisner: I was just copying from the numba unit tests, but sure, I can do a nested try/except if you want.

src/toast/utils.py
tskisner: Ok, and I guess in that case it means that numba is broken and we should just exit the function

src/toast/utils.py
tskisner: I agree empty returns are not necessary, and I have a habit of using them anyway since it makes it easier for my eyes to catch the end of the function.  I'll remove it.

tskisner:
Before merging this, I am investigating a larger issue.  I have small, standalone numba test that replicates the nested njit blocks used by PySM and calls them inside an mpi4py-enabled main.  I see the same OpenMP errors that @keskitalo saw in some toast runs.  I want to do more tests on this small program to isolate what is going on.

We know from experience over the past 5-6 years that one cannot mix MPI and python multiprocessing in the same job (it is implementation dependent and fails randomly sometimes when processes are forked from an MPI process).  However, I had assumed that if numba was using OpenMP for parallelism then everything would "play nice".

Currently PySM has a numba parallel njit that calls a non-parallel njit block, which in turn calls numpy.interp (which is probably in threaded compiled code internally).  This is what my small test script is exploring.  However, even if I remove the numpy calls it fails, so there is more to understand.

This problem is larger than PySM, since other dependencies may use numba in the future.  So we need to understand the underlying mechanisms and figure out how to configure numba to work in a way that is compatible with an MPI/OpenMP environment.

zonca:
Numba never calls numpy, np.interp is replaced by LLVM equivalents

tskisner:
Ah, ok, good to know!

src/toast/utils.py
```diff
@@ -36,6 +36,64 @@
...
+            # no TBB
+            pass
+    try:
+        from numba import config, njit, threading_layer
+
+        config.THREADING_LAYER = threading
+
+        # In order to get numba to actually select a threading layer, we must
+        # trigger compilation of a parallel function.
+        @njit(parallel=True)
```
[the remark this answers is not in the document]
stuartarchibald: This might help, `@vectorize(target='parallel')` is supported on more platforms than `@njit(parallel=True)`:
```python
import numba
numba.config.THEADING_LAYER='openmp'
from numba import vectorize
import numpy as np

@vectorize('float64(float64)', target='parallel')
def force_thread_launch(x):
    return x + 1

force_thread_launch(np.zeros(1))

print(numba.threading_layer())
```
it also feels like finding out what threading layer will be used should be easier... if that would be useful please open a feature request on the Numba issue tracker.

src/toast/utils.py
```diff
@@ -36,6 +36,64 @@
...
+        def foo(a, b):
+            return a + b
+
+        x = np.arange(10.0)
+        y = x.copy()
+        out_py = x + y
+        out_numba = foo(x, y)
+        if not np.allclose(out_py, out_numba):
```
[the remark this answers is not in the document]
stuartarchibald: It's unusual to find a case where Numba successfully compiled something (via a specified signature or otherwise) and the resulting compiled function returns an incorrect result. If this is the case it's a Numba bug, please file a bug report!

stuartarchibald:
> The [doc](http://numba.pydata.org/numba-doc/latest/user/threading-layer.html#selecting-a-threading-layer-for-safe-parallel-execution) doesn't seem to mention which one is default explicitly (by searching 'default'.)

The threading layer that corresponds to `"default"` depends on what is installed/available and what actually works in practice on your platform. In the `"default"` case Numba simply tries to load each of `tbb, omp, workqueue` in turn and settles on the first that loads successfully.
 
> IIRC, in the beginning there's only tbb available in Numba and so that probably remains the default.

I don't think that this is correct, TBB support was added much later than the parallel targets themselves. Numba ships with a reference threading backend "workqueue", this is always present and has existed for a very long time. `TBB` was added later and needs the `tbb` package. `OpenMP` was added at the same time as `TBB` and depending on your platform it will either work out of the box or require an additional package. See [here](http://numba.pydata.org/numba-doc/latest/user/threading-layer.html#selecting-a-named-threading-layer) for the requirements.

ickc:
Following up the call just now, I have some experience mixing different kind of threadings in Python. e.g. in one job from other project, I have:

```bash
export MKL_NUM_THREADS=16
export MKL_DOMAIN_NUM_THREADS="MKL_BLAS=16"
export MKL_DYNAMIC=TRUE

export OMP_NUM_THREADS=16
export OMP_PLACES=threads
export OMP_PROC_BIND=spread
export OMP_DYNAMIC=TRUE

export NUMEXPR_NUM_THREADS=16
export OPENBLAS_NUM_THREADS=16
export NUMBA_NUM_THREADS=16

export MPICH_GNI_FORK_MODE=FULLCOPY
```

(This is quite an "abomination", even uses multiprocessing with MPI, but it works in my case...)

**In the case that we can guarantee no 2 different threading is running at the same time, may be we don't even need to specify the threading layer in Numba and just let it choose according to what's available. In that case TBB would be chosen if it is available** (but fall back to OpenMP/workqueue in that order.)

tskisner:
There was a bunch of offline discussion recently about this, but I don't want to rehash it all here on this github issue...  For pure conda systems when toast is installed through conda as well, numba and toast can use the same libgomp / libiomp thread pool.  For systems where numba is using one thread pool and toast another, we know that they will never be running concurrently and we can hopefully find the specific settings at NERSC to allow these threadpools to exist simultaneously (oversubscribe the cores).  Will test some things soon.

tskisner:
Rebased to latest master, and I tried to address the above comments:

1.  use vectorize rather than njit
2.  remove consistency test

I also now use the NUMBA environment variables to control the threading layer.  I tested this against all the different combinations of backends.  With omp and tbb backends the issues we saw at NERSC seem resolved.

In the future, if more of our dependencies start using numba, then we could consider running this function once from the top-level module import.

@ickc , I commented on your numba issue about caching:  I agree it is a problem, and I feel what we really need is a way to disable caching globally.  We can continue discussion on that issue.  I'll merge this as soon as tests pass.

ickc:
The tests are failing because NERSC is down for maintenance and wget failed because of that.

tskisner:
Yes indeed.  A temporary problem.

tskisner:
On further consideration, it is probably a good idea for us to call this function on import rather than guess what dependencies may be using numba.  Will push that small change shortly.

tskisner:
The travis tests will fail until the NERSC project filesystem is back online (from where we download our bundle of build dependencies to save time when running travis).  The unit tests run locally for me with 4 MPI processes and I also tested that I could move different numba threading layer compiled extensions out of the way and that the logic behaves as expected.  Merging this now.

PR #287, opened 2019-09-11
Fix helper function and add a check to qarray.rotation

keskitalo:


tskisner:
Thanks for catching this!  Merge when ready

PR #288, opened 2019-09-17
Functional toast mapmaker; reorganize tod, map and todmap; run uncrus…

keskitalo:
…tify and black on the source tree

The mapmaker itself is still rough and will have many more features. It is important to merge the current state of the work so the branch does not diverge so much that it becomes difficult to merge.

PR #291, opened 2019-10-03
Add debug-level logging to the code that sets the numba threading layer.

tskisner:
Log the threading layer selection process if TOAST_LOGLEVEL=DEBUG.

### DIFF
--- a/etc/numba_test/numba_threads.py
+++ b/etc/numba_test/numba_threads.py
@@ -2,8 +2,6 @@
 
 from toast.mpi import MPI
 
-import os
-
 import numpy as np
 
 from numba import njit, config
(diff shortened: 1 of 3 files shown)


tskisner:
Note that the travis tests will fail pending travis changes in the toast_mapmaker branch.  The tests run locally on this branch.

tskisner:
This PR adds some more cosmetic logging and sets values in both os.environ and the numba.config structure.  Unfortunately, the environment is parsed on every numba compilation and our changes don't seem to "stick".  The only reliable way to force the number of threads seems to be setting NUMBA_NUM_THREADS in the parent environment where python is launched.

The travis tests won't pass in this branch since it does not include the fixes just merged to master in #288.  Merging this now since it does not affect anything (just increases the logging).

PR #292, opened 2019-10-06
Add fix for a race condition

keskitalo:


keskitalo:
self-merging a simple fix to unit test issue.

PR #293, opened 2019-10-08
sim_focalplane.py: plot_focalplane return fig instead of None

ickc:
Useful if people want to manipulate fig itself, and should be backward compatible.


tskisner:
Looks good, thanks.

Issue #298, opened 2019-10-17
Unable to compile TOAST with OpenMPI


ziotom78:
I am trying to compile TOAST within the following environment:

- GCC 9.2.0
- CMake 3.15.4
- OpenMPI 4.0.2

Running `mkdir build && cd build && cmake .. && make` produces the following error:

```
In file included from /usr/include/c++/9.2.0/bits/stl_iterator_base_types.h:67,
                 from /usr/include/c++/9.2.0/bits/forward_list.h:36,
                 from /usr/include/c++/9.2.0/forward_list:38,
                 from /home/tomasi/Projects/toast/src/toast/pybind11/include/pybind11/detail/common.h:140,
                 from /home/tomasi/Projects/toast/src/toast/pybind11/include/pybind11/pytypes.h:12,
                 from /home/tomasi/Projects/toast/src/toast/pybind11/include/pybind11/cast.h:13,
                 from /home/tomasi/Projects/toast/src/toast/pybind11/include/pybind11/attr.h:13,
                 from /home/tomasi/Projects/toast/src/toast/pybind11/include/pybind11/pybind11.h:43,
                 from /home/tomasi/Projects/toast/src/toast/_libtoast_common.hpp:9,
                 from /home/tomasi/Projects/toast/src/toast/_libtoast_mpi.hpp:9,
                 from /home/tomasi/Projects/toast/src/toast/_libtoast_mpi.cpp:6:
/usr/include/c++/9.2.0/type_traits: In instantiation of ‘struct std::is_base_of<pybind11::detail::pyobject_tag, ompi_communicator_t>’:
[snip…]
/home/tomasi/Projects/toast/src/toast/_libtoast_mpi.cpp:82:10:   required from here
/usr/include/c++/9.2.0/type_traits:1301:12: error: invalid use of incomplete type ‘struct ompi_communicator_t’
 1301 |     struct is_base_of
      |            ^~~~~~~~~~
In file included from /home/tomasi/Projects/toast/src/toast/../libtoast_mpi/include/toast_mpi.hpp:9,
                 from /home/tomasi/Projects/toast/src/toast/../libtoast_mpi/include/toast_mpi_internal.hpp:9,
                 from /home/tomasi/Projects/toast/src/toast/_libtoast_mpi.hpp:11,
                 from /home/tomasi/Projects/toast/src/toast/_libtoast_mpi.cpp:6:
[...]

PR #299, opened 2019-10-17
Fix for bug #298

ziotom78:


tskisner:
Thanks @ziotom78 , the one travis failure was due to an unrelated astropy file download issue.

PR #300, opened 2019-10-17
Update installation instructions

ziotom78:
I successfully managed to compile TOAST on my laptop, and I have written a step-by-step guide in file `docs/install.rst`, which still contained instructions for the old Autotools-based installation.

I have added instructions about how to build the documentation too.


docs/install.rst
```diff
@@ -87,28 +88,67 @@ install the dependencies that are available through conda::
...
+Using CMake
 -----------------------
 
-TOAST uses autotools to configure, build, and install both the compiled code
-and the python tools.  If you are running from a git checkout (instead of a
-distribution tarball), then first do::
+TOAST uses CMake to configure, build, and install both the compiled code
+and the python tools.  The suggested way to run TOAST is to install it in a
+directory within your home folder, and then make it visible using the
+``PYTHONPATH`` environment variable.
```
tskisner: In order to have a functioning install, we also need to add the bin directory to PATH.  I see you added that below, so that is good.

docs/install.rst
```diff
@@ -117,4 +157,16 @@ Testing the Installation
 After installation, you can run both the compiled and python unit tests.
 These tests will create an output directory in your current working directory::
 
-    %> python -c "import toast.tests; toast.tests.run()"
```
tskisner: This python command is still the way to run the unit tests.  It runs the python tests as well as the compiled tests.  The `toast_test` executable only runs the compiled tests.

ziotom78: Done, I have reverted the old line of code.

PR #302, opened 2019-10-28
global_to_local was failing unless pixels were int64.

keskitalo:
This is a short term fix to a recent bug in `DistPixels.global_to_local()`.  Long term, we need a good way of checking for `NDarray` data type and write functions that can handle multiple types without unnecessarily repeating code.

### DIFF
--- a/src/toast/_libtoast_pixels.cpp
+++ b/src/toast/_libtoast_pixels.cpp
@@ -15,8 +15,8 @@ py::tuple global_to_local(
     // Get raw pointers to input.
     py::buffer_info gpinfo = global_pixels.request();
     py::buffer_info glinfo = global2local.request();
-    T * rawgpdata = reinterpret_cast <T *> (gpinfo.ptr);
-    int64_t * rawgldata = reinterpret_cast <int64_t *> (glinfo.ptr);
+    T * global_pixels_raw = reinterpret_cast <T *> (gpinfo.ptr);
+    int64_t * global_to_local_raw = reinterpret_cast <int64_t *> (glinfo.ptr);
 
     size_t nsamp = gpinfo.size;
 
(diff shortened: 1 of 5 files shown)


src/toast/map/pixels.py
```diff
@@ -16,6 +16,8 @@
 
 from ..cache import Cache
 
+from .._libtoast import global_to_local as libtoast_g2l
```
keskitalo: Renaming `global_to_local` as `g2l` seems less than optimal. Anyone seeing that will be forced to scroll to the start of the file to figure out what `g2l` does. Can we either keep `global_to_local` or call it `libtoast_global_to_local`?

tskisner: I was avoiding importing this as "global_to_local" since there is already a class method by the same name.  Trying to distinguish between the function from the bindings and the DistPixels member function.  As long as they are different, I don't care what name is used for the imported function.

src/toast/_libtoast_pixels.cpp
```diff
@@ -5,37 +5,61 @@
...
+        Args:
+            global_pixels (array):  The global pixel indices.
+            npix_submap (int):  The number of pixels in each submap.
+            global2local (array, int64):  The local submap for each global submap.
+
+        Returns:
+            (tuple):  The (local submap, pixel within submap) for each global pixel.
+
+    )");
+    m.def("global_to_local", &global_to_local <int32_t> );
```
keskitalo: This is a nice, clean solution to exposing templated functions in `libtoast` to Python. I'm sure I'll have much use for this. Thanks!

keskitalo: Is there any reason this technique would not work for https://github.com/hpc4cmb/toast/blob/master/src/toast/_libtoast_todmap_scanning.cpp?

Right now the Python code has to figure out which version to `scan_map` and `fast_scanning` to call. If we can make that transparent, it would clean the Python code considerably.

tskisner: As I mentioned in an email, there are 2 ways of binding templated code:

1.  Pass arguments like array_t<T> so that different overloads can be detectected from the argument type.  Then register all bindings under the same name.  Or

2.  Pass arguments as py::buffer, so that they can accept any buffer provider (not just numpy arrays).  This means there are not templated arguments to deduce which overload to call.  So the bindings for different types must have different names.

So, if we only ever want to support numpy arrays and not memory from other types of objects, we can use (1) above.  Most of the bindings currently use technique (2).

keskitalo:
@tskisner this branch passes the unit tests on my system. Are we good to merge?

PR #303, opened 2019-10-29
Elevation noise

keskitalo:
We lost this feature from the ground simulation pipeline during the refactor. Here is it re-introduced along with some clean-up of the ground simulation pipeline.


tskisner:
These changes look fine.  I just restarted the travis build now that NERSC is back and the tarballs are available again.  Merge any time after they finish.

keskitalo:
Tests pass. Merging.

PR #304, opened 2019-11-01
Conviqt operator

keskitalo:
Modernize `OpSimConviqt`:
1.  works with arbitrary data distribution
2. can use an arbitrary communicator. Most importantly the rank and group communicators. 
3. add `pipeline_tools` helper functions: `add_conviqt_args` and `apply_conviqt`
4. integrate into `toast_satellite_sim.py`
5. gets the polarization efficiency and angle from a "focalplane" entry in the observation. It is an error to call `OpSimConviqt.exec()` with a data object that has observations without a "focalplane" entry

### DIFF
--- a/src/toast/pipeline_tools/pointing.py
+++ b/src/toast/pipeline_tools/pointing.py
@@ -90,12 +90,6 @@ def expand_pointing(args, comm, data):
     timer = Timer()
     timer.start()
 
-    hwprpm = args.hwp_rpm
-    hwpstep = None
-    if args.hwp_step_deg is not None:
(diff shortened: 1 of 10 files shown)


src/toast/pipeline_tools/pointing.py
```diff
@@ -90,12 +90,6 @@ def expand_pointing(args, comm, data):
     timer = Timer()
     timer.start()
 
-    hwprpm = args.hwp_rpm
```
tskisner: Thanks for catching this stale bit of code.

PR #306, opened 2019-11-02
Store local submap information in the data object

keskitalo:
This PR moves the identification of local submaps from a separate operator into the pixelization operator, such as `OpPointingHpix`. Since the submap information is not specific to any single observation, it establishes new, dictionary-like storage at the `data` object level for pixelization metadata, such has total number of pixels, size of a submap and the locally hit submaps. This information is stored with the same prefix as the pixel vectors in the TOD caches so multiple parallel pixelizations can be supported.

The PR removes some boiler plate from the pipelines as the locally hit submaps are now transparent to the user. This is also reflected in changes to the `DistPixels` class which can now be instantiated with the `data` object, without specifying the submap metadata explicitly.

### DIFF
--- a/src/toast/todmap/pointing.py
+++ b/src/toast/todmap/pointing.py
@@ -56,6 +56,7 @@ class OpPointingHpix(Operator):
             with the common_flag_mask, and then flag the pointing matrix.
         single_precision (bool):  Return the pixel numbers and pointing
              weights in single precision.  Default=False.
+        nside_submap (int):  Size of a submap is 12 * nside_submap ** 2
     """
 
     def __init__(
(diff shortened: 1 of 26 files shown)


src/toast/todmap/pointing.py
```diff
@@ -117,6 +123,16 @@ def mode(self):
...
 
+    @property
+    def local_submaps(self):
+        """(list): Indices of locally hit submaps
+        """
+        if self._single_precision:
+            dtype = np.int32
+        else:
+            dtype = np.int64
+        return np.arange(self._nsubmap, dtype=dtype)[self._hit_submaps]
```
tskisner: Even at NSIDE=8192, with the default nside_submap=16, the length of the np.arange (before slicing, with int64) would still only be a couple of MB, which is fine.

PR #307, opened 2019-11-05
new libaatm uses 64bit unsigned integers (size_t)

keskitalo:
Fix the `aatm` interface to work with https://github.com/hpc4cmb/libaatm

The modified source is not compatible with the old `aatm` from http://www.mrao.cam.ac.uk/~bn204/alma/atmomodel.html


tskisner:
I uploaded new travis tarballs and retriggered this build, which still fails.  Looks like an issue linking to the new libaatm.  I'll investigate further and try to resolve this today, since currently other PRs will fail as well until this is sorted out.

tskisner:
Ok, fixed by new travis tarballs that have libaatm 1.0.8.  Merging.

PR #309, opened 2019-11-16
Change the way that MPI communicators are passed to C++

tskisner:
Change the way that the MPI communicator is passed frompython to C++.  Explicitly get the mpi4py pointer to the MPI_Comm and then duplicate it so that it is independent of the python object lifetime.  This fixes a remaining OpenMPI issue which was seen bulding the conda package (even after the earlier "opaque type declaration" fix).

This should work with any MPI implementation used by mpi4py.


tskisner:
The travis failure was a sporadic network failure.  Merging now.

PR #313, opened 2019-11-25
allow disabling sky signal from the command line

keskitalo:
Most pipeline_tools packages allow disabling the module from the command line. This PR adds that functionality to PySM, conviqt and map sampling: regardless of what arguments are specified before, if the user extends the command line with `--no-sky` or `--no-simulate-sky`, then the sky simulation is disabled.

### DIFF
--- a/src/toast/pipeline_tools/sky_signal.py
+++ b/src/toast/pipeline_tools/sky_signal.py
@@ -14,6 +14,43 @@ from ..map import DistPixels
 from ..todmap import OpSimPySM, OpSimScan, OpMadam, OpSimConviqt
 
 
+def add_sky_signal_args(parser):
+    # The sky signal flag may be already added
+    try:
+        parser.add_argument(
+            "--sky",
+            required=False,
+            action="store_true",
+            help="Add simulated sky signal",
+            dest="simulate_sky",
+        )
+        parser.add_argument(
+            "--simulate-sky",
+            required=False,
+            action="store_true",
+            help="Add simulated sky",
+            dest="simulate_sky",
+        )
+        parser.add_argument(
+            "--no-sky",
+            required=False,
+            action="store_false",
+            help="Do not add simulated sky",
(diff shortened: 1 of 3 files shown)


src/toast/pipeline_tools/sky_signal.py
```diff
@@ -284,12 +328,12 @@ def scan_sky_signal(
...
 def simulate_sky_signal(
-    args, comm, data, focalplanes, cache_prefix, verbose=False, pixels="pixels", mc=0,
+    args, comm, data, focalplanes, cache_prefix, verbose=False, pixels="pixels", mc=0
 ):
     """ Use PySM to simulate smoothed sky signal.
 
     """
-    if not args.pysm_model:
+    if not args.pysm_model or not args.simulate_sky:
         return None
```
zonca: As You are modifying this, what about replace return None with return only?

keskitalo: Explicit is better than implicit (https://www.python.org/dev/peps/pep-0020) especially when the method usually has a return value. Now that you brought this up,  I'll go add `None` to https://github.com/hpc4cmb/toast/blob/sky_signal_args/src/toast/pipeline_tools/sky_signal.py#L252-L255

keskitalo: Conviqt and PySM args already add these: https://github.com/hpc4cmb/toast/blob/sky_signal_args/src/toast/pipeline_tools/sky_signal.py#L136 and https://github.com/hpc4cmb/toast/blob/sky_signal_args/src/toast/pipeline_tools/sky_signal.py#L245 respectively.

zonca: I agree with the first answer to https://stackoverflow.com/questions/15300550/return-return-none-and-no-return-at-all.
Here we are breaking out of a func so I would use return.

PR #315, opened 2019-12-09
Cleanup and fix detection of BLAS/LAPACK and name mangling.

tskisner:
Now works on OS X with both accelerate framework and openblas.


tskisner:
The travis failure was caused by another stupid-astropy-downloading-things-any-time-we-feel-like-it issue.  Merging now.

PR #316, opened 2020-01-03
High cadence scans

keskitalo:
This PR adds a new ground constant elevation scan (CES) mode. With `--scan-sinc-modulate`, the user can instruct `TODGround` object to simulate CES:s that modulate the scan rate proportional to `1/sin(Azimuth)`. This modulation leads to uniform integration depth even when sweeping across wide ranges in Declination. The downside is that the modulation requires increasing the scan rate as the telescope scans towards North or South. It is typically necessary to exclude azimuths within 20-30 degrees of the North or South directions.

This PR also fixes a bug in the turnaround simulation that sped up the telescope acceleration and lead to unrealistically short turnaround times.

### DIFF
--- a/src/toast/todmap/sim_tod.py
+++ b/src/toast/todmap/sim_tod.py
@@ -739,11 +739,14 @@ class TODGround(TOD):
             C (Equatorial), E (Ecliptic) or G (Galactic)
         report_timing (bool):  Report the time spent simulating the scan
             and translating the pointing.
-        hwprpm: if None, a constantly rotating HWP is not included.  Otherwise
-            it is the rate (in RPM) of constant rotation.
-        hwpstep: if None, then a stepped HWP is not included.  Otherwise, this
-            is the step in degrees.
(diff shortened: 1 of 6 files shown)


src/toast/todmap/sim_tod.py
```diff
@@ -1013,140 +1018,166 @@ def scan_range(self):
...
+
+        azmin, azmax = [self._azmin, self._azmax]
+        if self._sinc_modulation:
+            # We always simulate a rising sinc scan and then
+            # mirror it if necessary
+            azmin %= np.pi
+            azmax %= np.pi
+            if azmin > azmax:
+                raise RuntimeError(
+                    "Cannot scan across zero meridian with sinc-modulated scan"
```
tskisner: How much of a limitation is this?

keskitalo: It is a fundamental limitation of the sinc-modulated strategy. 1/sin(az) diverges on the zero meridian so one cannot scan across.

PR #317, opened 2020-01-09
Handle the situation where the "intervals" key is not in the obs dictionary

tskisner:
Also explicitly check for "None" (rather than simply a false value {False, 0, [], None}) since None is used as the default value and we are testing whether the variable has been set.

### DIFF
--- a/src/toast/todmap/mapmaker.py
+++ b/src/toast/todmap/mapmaker.py
@@ -1585,9 +1585,9 @@ class OpMapMaker(Operator):
         self.detweights = []
         for obs in data.obs:
             tod = obs["tod"]
-            try:
+            if "noise" in obs:
                 noise = obs["noise"]
-            except:
+            else:
                 noise = None
             detweights = {}
             for det in tod.local_dets:
--- a/src/toast/todmap/mapmaker.py
+++ b/src/toast/todmap/mapmaker.py
@@ -144,7 +144,7 @@ class SubharmonicTemplate(TODTemplate):
             tod = obs["tod"]
             common_flags = tod.local_common_flags(self.common_flags)
             common_flags = (common_flags & self.common_flag_mask) != 0
-            if self.intervals and self.intervals in obs:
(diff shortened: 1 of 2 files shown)


tskisner:
Don't merge this yet- I'm going to add another fix in a moment.

src/toast/todmap/mapmaker.py
```diff
@@ -1585,9 +1585,9 @@ def get_detweights(self, data):
         self.detweights = []
         for obs in data.obs:
             tod = obs["tod"]
-            try:
+            if "noise" in obs:
```
keskitalo: Do you disapprove of the "easier ask for forgiveness than permission" programming or the fact that the classes of acceptable exceptions were not listed?

src/toast/todmap/mapmaker.py
```diff
@@ -144,7 +144,7 @@ def get_steps_and_preconditioner(self):
             tod = obs["tod"]
             common_flags = tod.local_common_flags(self.common_flags)
             common_flags = (common_flags & self.common_flag_mask) != 0
-            if self.intervals and self.intervals in obs:
+            if (self.intervals is not None) and (self.intervals in obs):
```
keskitalo: I can't decide if redundant parentheses improve readability or add clutter.

src/toast/todmap/mapmaker.py
tskisner: Obviously try / except is a standard python idiom.  If you would prefer to use that and list the key error exception that is fine.  I guess it is a matter of preference.  We already know exactly what we want to test (whether the key is in the dictionary), so why not just check that?

src/toast/todmap/mapmaker.py
tskisner: I find long logical expressions without parenthesis difficult to parse visually.  That may just be a personal issue I have.  Feel free to remove them.  The main thing I care about is using 
```
if A is not None:
```
rather than
```
if A:
```
Since those are actually different things.

PR #324, opened 2020-03-04
Boresight offset

keskitalo:
This PR adds a feature in the scheduler to add an offset to the scheduled scan. An application of this is that one may create observing schedules for an individual tube of the focal plane. It is often necessary for beam campaigns.


tskisner:
The changes are trivial and look fine.  Feel free to merge after tests pass.

PR #329, opened 2020-04-08
Boresight offset

keskitalo:
Re-issuing this PR after the changes somehow got lost.


keskitalo:
Self-merging as these changes were already reviewed once.

Issue #330, opened 2020-04-14
PySM astropy.utils.data cache locking


tskisner:
When running PySM, data files are downloaded using astropy.utils.data, which caches files in ~/.astropy.  When multiple MPI groups are used in toast, there are multiple processes trying to download and read files.  Even if the maps exist in the cache (say from a previous run), then if multiple processes try to read these files at the same time it produces an astropy exception that the cache was locked for more than 5 seconds (the default locking time).

Here is a tarball of a self-contained example on cori.nersc.gov.  

[job_nersc-cori-knl_satellite_small.tar.gz](https://github.com/hpc4cmb/toast/files/4475753/job_nersc-cori-knl_satellite_small.tar.gz)

A typical traceback from master branch is:
```
Proc 45: Traceback (most recent call last):
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/bin/toast_satellite_sim.py", line 477, in <module>
    main()
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/bin/toast_satellite_sim.py", line 341, in main
    args, comm, data, [focalplane], "signal"
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/lib/python3.6/site-packages/toast/timing.py", line 43, in df
    return f(*args, **kwargs)
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/lib/python3.6/site-packages/toast/pipeline_tools/sky_signal.py", line 358, in simulate_sky_signal
    pixels=pixels,
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/lib/python3.6/site-packages/toast/timing.py", line 43, in df
    return f(*args, **kwargs)
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/lib/python3.6/site-packages/toast/todmap/sim_det_pysm.py", line 142, in __init__
    map_dist=map_dist,
Proc 45:   File "/global/common/software/cmb/cori/cmbenv-intel_20200115/cmbenv_aux/lib/python3.6/site-packages/toast/todmap/pysm.py", line 76, in __init__
    if init_sky
[...]

PR #331, opened 2020-04-14
Attempt serialize the data caching and loading done by PySM

tskisner:
`astropy.utils.data` is super convenient, but seems fragile on distributed systems.  See #330.  This branch attempts to "pre-initialize" the sky model on a single process, which should download the files.  Then it instantiates each Sky object one at a time, to avoid multiple processes reading the files at the same time.  Unfortunately this technique does not yet work.

### DIFF
--- a/src/toast/pipeline_tools/dist.py
+++ b/src/toast/pipeline_tools/dist.py
@@ -10,7 +10,7 @@ import numpy as np
 
 from ..mpi import get_world, Comm
 from ..timing import function_timer, Timer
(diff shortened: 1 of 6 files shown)


tskisner:
This issue was resolved by upstream options to PySM and healpy.  When we start running the bandpass integration on-the-fly with multiple process groups, we will need to add logic to load the pixel weight files and pass them to the call to healpy coordinate transform.  We can open a separate issue (and PR) at that time.

PR #332, opened 2020-04-14
Update examples for use in benchmarking

tskisner:
This PR greatly simplifies the generation of toast examples.  Instead of directories parameter files and a mix of cmake and bash, we now have a single TOML config file and a python script.

This PR is still a work in progress.  The old files are still in place and the new setup is still be expanded and tested on cori.


tskisner:
Ok, the generation of the example jobs is working well.  However, we definitely still need to tweak options in `config.toml`.  I suggest we merge this branch now, so that @keskitalo can use these for testing the example script updates.  After those updates, we can have a separate PR to change options in `config.toml` (which may have changed anyway after the work on the example pipelines).

Note:  I did make some small changes to the example pipelines in this branch, but those may be irrelevant (or conflicting with) the changes that @keskitalo is making and should be removed as needed.

examples/config.toml
```diff
@@ -0,0 +1,546 @@
+
```
keskitalo, line 1: This is a really clean way of defining the example jobs.

examples/generate.py
```diff
@@ -0,0 +1,522 @@
...
+
+    outstr = "{}\n".format(outstr)
+
+    return outstr
+
+
+def ceil_div(a, b):
+    """Handy, from: https://stackoverflow.com/questions/14822184/
+    """
+    return -(-a // b)
```
keskitalo, line 114: Where was this function all my life?

examples/trigger_astropy.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
```
keskitalo, line 1: This script should deploy with `healpy`. Have you proposed it?

tskisner: An interesting idea.  Perhaps someday...

PR #341, opened 2020-04-24
Simulation tweaks

keskitalo:
This PR contains the latest modifications for SO and CMB-S4 simulations.


keskitalo:
Here it is. Apologies in advance for this massive PR.

src/libtoast/src/toast_atm.cpp
```diff
@@ -3,9 +3,9 @@
 // All rights reserved.  Use of this source code is governed by
 // a BSD-style license that can be found in the LICENSE file.
 
-// #if !defined(DEBUG)
-// #   define DEBUG
-// #endif
+#if !defined(DEBUG)
```
tskisner, line 6: Did you intend to always define "DEBUG"?

src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -3,6 +3,9 @@
 // All rights reserved.  Use of this source code is governed by
 // a BSD-style license that can be found in the LICENSE file.
 
+#include <string.h>
+#include <omp.h>
```
tskisner, line 7: This needs to be guarded, like:
```

#ifdef _OPENMP
# include <omp.h>
#endif // ifdef _OPENMP
```

src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -52,108 +58,228 @@ void toast::filter_polynomial(int64_t order, size_t n, uint8_t * flags,
...
+        }
+    }
 
-                // For whatever reason, trmv refused to yield the right
-                // answer...
-                // Use general matrix vector multiply instead
+#pragma omp parallel \
+    default(none) shared(invcov, templates, signal, good, ntemplate, nsample)
+    {
+        int nthread = omp_get_num_threads();
```
tskisner, line 195: These calls to `omp_*` functions need to be guarded like:
```
int nthread = 1;
int id_thread = 0;
#ifdef _OPENMP
nthread = omp_get_num_threads();
id_thread = omp_get_thread_num();
#endif
```

src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -52,108 +58,228 @@ void toast::filter_polynomial(int64_t order, size_t n, uint8_t * flags,
...
+                    #pragma omp simd
+                    for (size_t i = 0; i < scanlen; ++i) {
+                        const double x = xstart + i * dx;
+                        current[i] = x;
+                    }
+                } else {
+                    last = &full_templates[(iorder - 1) * scanlen];
+                    lastlast = &full_templates[(iorder - 2) * scanlen];
+                    double orderinv = 1. / iorder;
+                    #pragma omp simd
```
tskisner, line 78: Note that `omp simd` assumes all allocated memory buffers are aligned, and will return incorrect results if not.  In this case the `last` and `lastlast` are pointers to aligned memory, so it is ok.

src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -52,108 +58,228 @@ void toast::filter_polynomial(int64_t order, size_t n, uint8_t * flags,
...
+                                 proj.data(), &norder,
+                                 singular_values.data(), &rcond_limit,
+                                 &rank, work.data(), &lwork, &info);
+
+            for (int iorder = 0; iorder < norder; ++iorder) {
+                double * temp = &full_templates[iorder * scanlen];
+                for (int isignal = 0; isignal < nsignal; ++isignal) {
+                    double * signal = &signals[isignal][start];
+                    double amp = proj[iorder + isignal * norder];
+                    #pragma omp simd
```
tskisner, line 163: Here is a case where signal points to the `signals` variable which is passed into the function and of unknown alignment.  The temp variable points to `full_templates` which is allocated in aligned memory.  There is a helper function for checking alignment:
```
if (toast::is_aligned(signals)) ...
```
See the healpix or qarray source files for examples.

src/libtoast_mpi/src/toast_mpi_atm.cpp
```diff
@@ -3,9 +3,9 @@
 // All rights reserved.  Use of this source code is governed by
 // a BSD-style license that can be found in the LICENSE file.
 
-// #if !defined(DEBUG)
-// #   define DEBUG
-// #endif
+#if !defined(DEBUG)
```
tskisner, line 6: Should DEBUG always be enabled?

src/libtoast/src/toast_atm.cpp
keskitalo: I meant for all the code in the current DEBUG blocks to be enabled. I changed the preprocessed variable to NO_ATM_CHECKS and the code can be disabled by setting it.

src/libtoast/src/toast_tod_filter.cpp
keskitalo: Done.

src/libtoast/src/toast_tod_filter.cpp
keskitalo: Great!

src/libtoast/src/toast_tod_filter.cpp
keskitalo: Thanks. Done.

src/libtoast/src/toast_tod_filter.cpp
keskitalo: Done.

src/libtoast_mpi/src/toast_mpi_atm.cpp
keskitalo: Switched to a more appropriate name.

keskitalo:
@tskisner The OpenMP fixes are now in place. Good to merge?

tskisner:
Yes please :-)

PR #343, opened 2020-05-05
adding HWP support to OpSimConviqt.

giuspugl:
The approach that we are following here is  to slightly modify the convolution function and add the extra angle from the HWP to the polarization angle prior to the interpolation of the data cube, to get the  observed signal in a given TOD sample.
we convolve three streams, one for intensity, one for perfectly co-polarized beam and one for a perfectly cross-polarized beam and finally we coadd them.

### DIFF
--- a/src/toast/pipeline_tools/sky_signal.py
+++ b/src/toast/pipeline_tools/sky_signal.py
@@ -295,6 +295,7 @@ def apply_conviqt(args, comm, data, cache_prefix="signal", verbose=True):
 
     return cache_prefix
 
+
 @function_timer
 def apply_weighted_conviqt(args, comm, data, cache_prefix="signal", verbose=True):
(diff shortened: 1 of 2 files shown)


giuspugl:
I have updated the weights for the polarization angles consistently as it is now in 
https://github.com/hpc4cmb/toast/blob/2a24945db55dea355f2c7906992b47ca40b18499/src/toast/todmap/pointing.py#L36 
see https://github.com/giuspugl/toast/blob/04957d2bee54172f0d8e679947876daeb5b2b2ec/src/toast/todmap/conviqt.py#L529 
and 
https://github.com/giuspugl/toast/blob/04957d2bee54172f0d8e679947876daeb5b2b2ec/src/toast/todmap/conviqt.py#L578

PR #345, opened 2020-05-06
Support passing None for HWP angles to pointing matrix calculation.

tskisner:
Rather than always passing an array of zeros if there is no HWP,
support passing None value to the underlying C++ code.  Fix an
error in the docstring.  Add a unit test to verify that no HWP
and HWP angles of all zero give matching results.

We should rebase this after #343 is merged.

### DIFF
--- a/src/toast/_libtoast_tod_pointing.cpp
+++ b/src/toast/_libtoast_tod_pointing.cpp
@@ -41,6 +41,13 @@ void init_tod_pointing(py::module & m) {
                   auto hwpbuf = py::cast <py::buffer> (hwpang);
                   pybuffer_check_1D <double> (hwpbuf);
                   py::buffer_info info_hwpang = hwpbuf.request();
+                  if (info_hwpang.size != n) {
+                      auto log = toast::Logger::get();
+                      std::ostringstream o;
+                      o << "HWP buffer size is not consistent.";
+                      log.error(o.str().c_str());
+                      throw std::runtime_error(o.str().c_str());
+                  }
                   rawhwpang = reinterpret_cast <double *> (info_hwpang.ptr);
               }
               toast::pointing_matrix_healpix(hpix, nest, eps, cal, mode, n,
--- a/src/toast/_libtoast_tod_pointing.cpp
+++ b/src/toast/_libtoast_tod_pointing.cpp
@@ -9,15 +9,13 @@
 void init_tod_pointing(py::module & m) {
     m.def("pointing_matrix_healpix",
           [](toast::HealpixPixels const & hpix, bool nest, double eps, double cal,
-             std::string const & mode, py::buffer pdata, py::buffer hwpang,
+             std::string const & mode, py::buffer pdata, py::object hwpang,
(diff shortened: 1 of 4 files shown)


src/toast/_libtoast_tod_pointing.cpp
```diff
@@ -36,16 +33,23 @@ void init_tod_pointing(py::module & m) {
...
-              double * rawhwpang = reinterpret_cast <double *> (info_hwpang.ptr);
               uint8_t * rawflags = reinterpret_cast <uint8_t *> (info_flags.ptr);
               double * rawweights = reinterpret_cast <double *> (info_weights.ptr);
               int64_t * rawpixels = reinterpret_cast <int64_t *> (info_pixels.ptr);
+              double * rawhwpang = NULL;
+              if (!hwpang.is_none()) {
+                  auto hwpbuf = py::cast <py::buffer> (hwpang);
+                  pybuffer_check_1D <double> (hwpbuf);
+                  py::buffer_info info_hwpang = hwpbuf.request();
+                  rawhwpang = reinterpret_cast <double *> (info_hwpang.ptr);
```
keskitalo, line 44: Shouldn't you check for the length of `hwpang` just like other vectors are checked?

src/toast/todmap/pointing.py
```diff
@@ -33,7 +33,7 @@ class OpPointingHpix(Operator):
     the total response is:
 
     .. math::
-        d = cal \\left[\\frac{(1+eps)}{2} I + \\frac{(1-eps)}{2} \\left[Q \\cos{4(a+w)} + U \\sin{4(a+w)}\\right]\\right]
+        d = cal \\left[\\frac{(1+eps)}{2} I + \\frac{(1-eps)}{2} \\left[Q \\cos{2a+4w} + U \\sin{2a+4w}\\right]\\right]
```
keskitalo, line 36: @giuspugl Is this definition of the pointing weights consistent with what Davide used for the LB simulations?

src/toast/todmap/pointing.py
```diff
@@ -165,8 +165,6 @@ def exec(self, data):
                 hwpang = tod.local_hwp_angle()
             except:
                 pass
```
keskitalo, line 167: Seems like assigning `hwpang = None` in the except clause would be prudent. I have been burned by overly general try-catch segments before. Perhaps it would make sense to list the exceptions we anticipate?

tskisner: Going recursively through the code, it looks like the call to `local_hwp_angle()` can raise `RuntimeError`, `ValueError`, and `NotImplementedError`.  I guess we need a cleaner way to determine if a TOD instance has HWP angles...

tskisner:
I added the check for the buffer length, and explicit reset of `hwpang` to None in the except clause.  I am going to leave it as a generic `except` for now, since cleaning that up would require changing code that will soon be deprecated.

keskitalo:
Great!

tskisner:
This has been rebased after #341.  Merging now.

PR #348, opened 2020-05-12
Build system improvements for pip wheel compatibility

tskisner:
This is still a work in progress.  So far I have:

- Simplified the cmake build system (fewer internal convenience libs)

- Improved the setup.py (which calls cmake).  This now uses mpi4py to attempt to find the MPI compiler wrappers used at install and then pass that to cmake.  This will help people pip-installing from github who also want to use MPI.


tskisner:
In the spirit of having more frequent, small PRs, I would like to just merge this work as-is in terms of the build system improvements.  Actually building pip packages will be in a later PR.

CMakeLists.txt
```diff
@@ -44,9 +44,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+if(NOT TOAST_STATIC_DEPS AND ENV{TOAST_STATIC_DEPS})
+  set(TOAST_STATIC_DEPS $ENV{TOAST_STATIC_DEPS})
+endif()
+
 find_package(OpenMP)
 
-find_package(BLAS)
+if(TOAST_STATIC_DEPS)
+    set(BLA_STATIC TRUE)
+    find_package(BLAS)
```
keskitalo, line 60: Is it necessary to have `find_package(BLAS)` in both branches? Why not after the `if`?

tskisner: Good point, will fix that.

tskisner:
Not sure why the travis status is not being displayed here on the PR, I just checked the build on travis-ci.org and it passed.  Merging now.

PR #349, opened 2020-05-16
Add github action to run unit tests

tskisner:
This PR changes the continuous integration service from travis to github actions / workflows.  The main benefits are larger allowed resource usage, including a 6 hour run time per job and no need to print things to stdout to keep the job alive like on travis.  Also the unit tests on github run faster than on travis.

The units test are also converted to run in pre-built docker containers with all dependencies.  These are generated automatically for python 3.5-3.8 whenever a new tag is made of the cmbenv repo:

https://github.com/hpc4cmb/cmbenv/blob/master/.github/workflows/deploy.yml

Hopefully this will streamline the process of updating the testing whenever a dependency is updated.


tskisner:
Ok @zonca and @keskitalo , please look at this when you have time.  It removes all the travis dependency stuff in favor of using the containers here:

https://hub.docker.com/orgs/hpc4cmb/repositories

And runs the serial / MPI unit tests and builds the docs directly on github.  The workflow status looks like this:

https://github.com/hpc4cmb/toast/pull/349/checks?check_run_id=681613869

I think this will be much easier to maintain and extend to deploying docs and pip wheels going forward.

.github/workflows/test.yml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
+  py35:
+    name: Python 3.5
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Pull Dependency Image
+        run: docker pull hpc4cmb/toast-deps-py35:latest
```
zonca, line 19: is there another repo to build the `toast-deps` docker container?

tskisner: Yes, whenever a new tag is pushed to the cmbenv repo, the toast-deps docker images are built and pushed to docker hub:

https://github.com/hpc4cmb/cmbenv/blob/master/.github/workflows/deploy.yml

So to update the toast-deps image, you make any changes to the package files (updating versions, etc) in the cmbenv repo.  Then make a cmbenv tag and toast-deps will be rebuilt.

tskisner: The toast-deps images are based on the upstream official `python:3.x-slim-buster` images, with compiled dependencies pip packages installed to `/usr/local`.

tskisner:
Ok, thanks for the feedback, merging now.

PR #350, opened 2020-05-21
Remove dependency on MPI compilers

tskisner:
This is still a work in progress. By reorganizing the atmosphere simulation code, we can remove our need for MPI compilers (we just need mpi4py).  This means that the internal toast compiled code can be built with any serial compiler.  There is no need for consistency between the compilers used to build toast and the compilers used to install mpi4py.  This greatly simplifies toast installation and will enable us to distribute binary pip packages.


tskisner:
Ok, I think this branch is ready for more extensive tests on actual simulation pipelines.  Running small local tests seem to indicate consistent results and timing compared to the current master branch.

tskisner:
Ok @keskitalo, this branch now agrees with master on the small test case you shared with me offline.  If you could reconfirm that consistency it would be great, and then hopefully we can merge this.

keskitalo:
The branch is now fully compatible with `master`.

PR #351, opened 2020-05-25
Begin work to integrate cibuildwheel tests.

tskisner:
This is a work in progress.  Add building of Linux and MacOS wheels to test runners.  These will have MPI support disabled until #350 is merged.


tskisner:
Ok, I've squashed all my debugging commits and I think this is ready to go.  The pip wheel tests will run daily and generate wheels that can be downloaded manually for testing as artifacts.  Whenever there is a tag, the wheels will be uploaded to PyPI.  I verified that I could download the generated wheels and install them manually in a virtualenv.  @zonca, no pressure to review this if you are busy, but obviously any comments are welcome.  It feels good to have all this packaging happening as part of our CI process :-)

setup.py
```diff
@@ -202,37 +215,47 @@ def move_output(self, ext):
...
-conf["python_requires"] = ">=3.4.0"
-conf["install_requires"] = [
-    "cmake",
-    "numpy",
-    "scipy",
-    "healpy",
-    "matplotlib",
-    "pyephem",
-]
+conf["python_requires"] = ">=3.5.0"
```
zonca, line 242: what about dropping 3.5? so there is no worries of using f-strings

setup.py
```diff
@@ -202,37 +215,47 @@ def move_output(self, ext):
...
 conf["ext_modules"] = ext_modules
 conf["cmdclass"] = {"build_ext": CMakeBuild}
 conf["zip_safe"] = False
+conf["classifiers"] = [
+    "Development Status :: 5 - Production/Stable",
+    "Environment :: Console",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: POSIX",
+    "Programming Language :: Python :: 3.6",
```
zonca, line 255: it looks like 3.5 is not listed here

src/toast/RELEASE
```diff
@@ -0,0 +1 @@
+2.3.7a0
```
zonca, line 1: is this a version id? I admit I haven't checked at https://semver.org/, but to my lazy self this seems to not be a semver.

wheels/install_deps_linux.sh
```diff
@@ -0,0 +1,146 @@
...
+    CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}" AUTOCC=no \
+    GPU_CONFIG="" CFOPENMP="${OPENMP_CXXFLAGS}" BLAS="-lopenblas -fopenmp -lm" \
+    && make static JOBS=${MAKEJ} \
+    CC="${CC}" CXX="${CXX}" \
+    CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}" AUTOCC=no \
+    GPU_CONFIG="" CFOPENMP="${OPENMP_CXXFLAGS}" BLAS="-lopenblas -fopenmp -lm" \
+    && cp -a ./include/* "${PREFIX}/include/" \
+    && find . -name "*.a" -exec cp -a '{}' "${PREFIX}/lib/" \; \
+    && popd >/dev/null 2>&1
+
```
zonca, line 144: there is so much building from source, isn't there another way? maybe do 1 docker container and update it once in a while? or conda install those?

wheels/install_deps_osx.sh
```diff
@@ -0,0 +1,119 @@
...
+# bundled with our compiled extension.
+#
+
+# Location of this script
+pushd $(dirname $0) >/dev/null 2>&1
+topdir=$(pwd)
+popd >/dev/null 2>&1
+
+# Get newer cmake with pip
+pip install cmake
```
zonca, line 16: pip can install `cmake`?? wow

setup.py
tskisner: I hesitate to drop 3.5, but let me check the default python versions on CentOS and the last couple Ubuntu LTS.

setup.py
tskisner: Good catch.

src/toast/RELEASE
tskisner: This must be a PEP440 compliant version, not a semantic version.  When installing with setup.py (install or bdist_wheel), a version based on the git tags is used.  When running "pip wheel" it looks like git is not available, so this file contains the tag version to use instead.

wheels/install_deps_linux.sh
tskisner: For linux we could do that by taking the cibuildwheel container installing our stuff, and pushing that new container to dockerhub.  For MacOS, I don't think we can do that...

setup.py
tskisner: Ok, CentOS7 (which we are using to build wheels, and so is the oldest glibc version we care about) ships with python 3.6.  Ubuntu 16.04 LTS ships with python 3.5, but that is now 4 years old and there are two LTS releases since then.  So- I am fine restricting to pyton 3.6.  @keskitalo , do you have any opinion on requiring >=3.6?

src/toast/RELEASE
tskisner: I will add a section to the docs about procedure for making a release, which (unfortunately) involves editting this file.

wheels/install_deps_linux.sh
tskisner: Re: conda.  I'm not sure we can mix conda-installed libraries with the system ones.  Conda uses its own compiler toolchain which acts like a cross compiler.  Of course it would also require installing a miniconda base system here and potentially debugging future incompatibilities.  The downside to building these things from source is the time it takes.  The upside is that we can control how they are built and link them statically into our compiled extension.  So our only external dependencies are libc, libm, etc

setup.py
keskitalo: No objections. I use 3.6, 3.7 and 3.8 on different systems.

wheels/install_deps_linux.sh
tskisner: I will leave this as-is for now, since it is working.  If we find a way to reliably install suitesparse / openblas from a package manager then this can be a new PR to this file in the future.

tskisner:
Appoved by @keskitalo in offline conversation.

PR #355, opened 2020-06-11
Support el nods

keskitalo:


src/toast/pipeline_tools/atm.py
```diff
@@ -369,7 +367,10 @@ def scale_atmosphere_by_frequency(
         # Use the entire processing group to sample the absorption
         # coefficient as a function of frequency
         freqmin = 0
-        freqmax = 2 * freq
+        if freq is None:
+            freqmax = 1000
```
tskisner, line 371: Is there any potential surpise if `freq` is not passed in and the upper limit of the frequency band is hardcoded to 1000GHz?  Just wondering if we should keep the exception and force specifying the frequency.

keskitalo: Certain experiments define the frequency (band) as a string and may have non-standard ways of communicating the bandpass. I would revisit this once the TOAST hardware map is at a shape that we know we can retrieve the detector bandpass from it.

PR #358, opened 2020-06-15
Scanning detector maps

keskitalo:
This PR makes it possible to scan a different input map for each detector. Typically each process in a process group has different detectors, so the reads are performed independently on each process.

### DIFF
--- a/src/toast/pipeline_tools/sky_signal.py
+++ b/src/toast/pipeline_tools/sky_signal.py
@@ -55,7 +55,14 @@ def add_sky_map_args(parser):
     """ Add the sky arguments
     """
 
-    parser.add_argument("--input-map", required=False, help="Input map for signal")
+    parser.add_argument(
+        "--input-map",
+        required=False,
+        help="Input map for signal.  You can use Python formatting for Monte "
+        "Carlo realization, e.g.: {:04i}.  If tag DETECTOR is encountered, it "
+        "will be replaced with the detector name and different map is loaded "
+        "for every detector.",
+    )
 
     # The nside may already be added
     try:
@@ -363,7 +370,7 @@ def apply_weighted_conviqt(args, comm, data, cache_prefix="signal", verbose=True
 
 @function_timer
 def scan_sky_signal(
-    args, comm, data, cache_prefix="signal", verbose=True, pixels="pixels", mc=0
+    args, comm, data, cache_prefix="signal", verbose=True, pixels="pixels", nnz=3, mc=0
 ):
     """ Scan sky signal from a map.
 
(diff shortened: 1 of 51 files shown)


src/toast/todmap/sim_det_map.py
```diff
@@ -166,13 +189,51 @@ def exec(self, data):
...
+                    if not os.path.isfile(filename):
+                        raise RuntimeError("Input map not found: {}".format(filename))
+                    detector_map = DistPixels(
+                        data,
+                        comm=comm,
+                        nnz=self._nnz,
+                        dtype=np.float32,
+                        pixels=self._pixels,
+                    )
+                    detector_map.read_healpix_fits(filename)
```
tskisner, line 234: You are looping over local detectors (potentially different on every process), but distributing the map over the communicator used for the whole observation.  If each process had the same number of detectors (loop iterations), then the filename on rank zero would be loaded into detector_map regardless of what detectors existed on other processes.  In the case of different numbers of detectors on each process you would have a hang on some processes waiting for communication.

It seems like some options would be:

1. Loop over the full detector list and only scan the map to timestreams if the detector is locally present on the process.
2.  OR, Loop over local detectors and use the process grid row communicator for the DistPixels object.

The second option would result in each row of the process grid independently scanning their detector maps.

tskisner:
Hello @keskitalo , it would be good to merge this very soon.  I can work on the communicator change if you don't have time in the next day.

keskitalo:
Thanks @tskisner . I'll try to get it done today.

src/toast/pipeline_tools/sky_signal.py
```diff
@@ -55,7 +55,14 @@ def add_sky_map_args(parser):
     """ Add the sky arguments
     """
 
-    parser.add_argument("--input-map", required=False, help="Input map for signal")
+    parser.add_argument(
+        "--input-map",
+        required=False,
+        help="Input map for signal.  You can use Python formatting for Monte "
+        "Carlo realization, e.g.: {:04i}.  If tag DETECTOR is encountered, it "
```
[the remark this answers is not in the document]
keskitalo: Addressed in 5fcca34759e2fa0ffb253545d0360253764ae46f

src/toast/todmap/sim_det_map.py
keskitalo: I don't think that to be the case. Lines 221-224 and 230 are there to ensure that every process loads and samples the map independent of other processes. I didn't want to use the rank communicators because we can't guarantee that same detectors are there for every rank.

tskisner: Ok, sorry I missed that- thanks for correcting my understanding.  In the case of scanning a single map for all detectors, that is done once (collectively) at the beginning of the function and then used for all observations.  In the case of detector-specific maps this code is basically doing:
```
for each observation:
    for each local detector:
        each process reads detector map from a file
            - DistPixels has the locally hit submaps for all detectors
              and all observations on this process 
        project to timestream
        discard map
```
So that means if you have 100 observations then a detector map gets loaded off disk 100 times for every process that has that detector.  I guess we are not running in this mode very much currently, but in the future we could consider each process doing a loop over observations to compute the superset of all detectors and then loop over detectors and populate the observations where it has that detector.  However, I am fine with this implementation if it serves our needs for now.

keskitalo: Yes, this is not optimal but it is a relatively clean solution that requires no blocking communication. Also, parallel reads tend to be efficient, especially when the files are only accessed by less than a hundred processes.

tskisner (approved):

PR #359, opened 2020-06-25
Switch to manylinux2010 image for wheels.

tskisner:
The current manylinux2014 image used for wheels allows us to use
a newer compiler, but seems to be incompatible with some
still in-use OS versions (e.g. Ubuntu 18.04 LTS).  See if this
is resolved by moving to the 2010 image.


tskisner:
Ok @keskitalo , this work is ready.  These changes update the internal pybind11 sources and include some small tweaks to the cmake files and the pip wheel deployment.  I have verified manually that the generated wheels run on:

- Ubuntu 20.04 with python 3.8
- Ubuntu 18.04 with python 3.6
- CentOS 6 with python 3.6
- CentOS 7 with python 3.6
- MacOS with python 3.7 from homebrew

The pip / virtualenv instructions in the docs have been updated to the commands that work reliably across all platforms.  I also verified manually on MacOS that the conda-forge recipe now works when building from the latest commit in this branch.  So this work will resolve the build failure on conda-forge with OS X.

After this is merged I'm going to quickly release 2.3.8 so that these fixes can propagate to PyPI and conda-forge.

PR #360, opened 2020-06-30
Check that plan returned by fftw is not null

mhasself:
Because MKL, when pretending to be fftw, cannot perform some transforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,10 @@ if(TOAST_STATIC_DEPS)
     set(SUITESPARSE_USE_STATIC_LIBS TRUE)
 endif()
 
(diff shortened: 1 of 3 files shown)


tskisner:
Thanks for this, I just pushed one small bit of extra logging.  Do you think we need to be more clever about detecting MKL's FFTW interface?  For example, should our top level CMakeLists.txt check for FFTW first (before BLAS/LAPACK) and make sure that our detected FFTW is appended to linking after any MKL for blas/lapack?

tskisner:
Ok, since it was easy, I changed the order of checking / linking FFTW.  Hopefully that will help avoid ending up with MKL symbols for FFTW.

mhasself:
Oh, nice... the linking order business is eye-opening.  I will give that a try next time... not everyone is as willing as I am to just purge MKL...

PR #361, opened 2020-07-01
Jrs584/fixed elevation

keskitalo:
This branch adds two new features to the TOAST scheduler:
* support for fixed, user-defined observing elevations
* support for performing partial scans when full scans are not available
Many, many thanks for @jrs584 for the original code!

### DIFF
--- a/pipelines/toast_ground_schedule.py
+++ b/pipelines/toast_ground_schedule.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright (c) 2015-2020 by the parties listed in the AUTHORS file.
+# Copyright (c) 2015-2019 by the parties listed in the AUTHORS file.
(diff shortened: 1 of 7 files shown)


pipelines/toast_ground_schedule.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright (c) 2015-2020 by the parties listed in the AUTHORS file.
+# Copyright (c) 2015-2019 by the parties listed in the AUTHORS file.
```
tskisner, line 3: We should change that back to 2020

keskitalo: thanks Ted. Yes, this one is tested thoroughly.

We don't have unit tests for the pipelines. Perhaps we should.

PR #365, opened 2020-10-06
Standalone benchmarking script

tskisner:
This script scales the data volume according to the job size and attempts to use a group size of 4 nodes and a fake focalplane such that every process has a few detectors.  This needs more testing.


tskisner:
Ok @keskitalo , this is ready for review.  In addition to the actual benchmark script this also moves the scheduler into the package and there are some other small tweaks.

src/toast/_libtoast_pixels.cpp
```diff
@@ -60,6 +60,6 @@ void init_pixels(py::module & m) {
             (tuple):  The (local submap, pixel within submap) for each global pixel.
 
     )");
-    m.def("global_to_local", &global_to_local <int32_t>);
-    m.def("global_to_local", &global_to_local <int16_t>);
+    m.def("global_to_local", &global_to_local <int32_t> );
```
keskitalo, line 63: Looks like `uncrustify` has slightly changed.

keskitalo (approved):
Looks great! Merge away.

src/toast/_libtoast_pixels.cpp
tskisner: Yeah, I am using version 0.69 from ubuntu sources.  I see upstream is at version 0.71.  If we become too annoyed in the future we could enforce a particular version...

PR #366, opened 2020-10-12
2D filtering and destriping

keskitalo:


keskitalo:
This work will require a focused study in the future to determine the best prior (C_a) matrix for the destriper.

tskisner (approved):
This is a large body of work, thanks for getting it in to master prior to the toast3 porting effort!

PR #371, opened 2020-11-07
Re-enable MKL support for FFTs

tskisner:
Our internal FFT routines have had support for directly using MKL for years.  However, this was disabled when we moved to our modern CMake build system.  This PR adds checks for MKL and if found does not look for FFTW.  This works around the frequent issue where MKL ships with a partial FFTW interface and so linking to both libraries can break our code.

Tested on `cori.nersc.gov` with Intel compilers and MKL.


tskisner:
Although statistical unit tests pass with both FFT frameworks, I am working on comparing generated timestream values to make sure that they are the same.  However, the build system changes in this PR can be reviewed already.

cmake/FindMKL.cmake
```diff
@@ -0,0 +1,414 @@
+# - Find INTEL MKL library
```
keskitalo, line 1: I assume this is all original code but the file is missing a copyright notice.

Rather surprising that there wasn't a cmake macro already available. Of course this way we get exactly what we need.

keskitalo (approved):

cmake/FindMKL.cmake
tskisner: This was from pytorch, but I am having to rewrite it from scratch now, so disregard.  We only want to link to `libmkl_rt.so`.  And indeed no "official" cmake check exists for MKL, although many people asked on the forums for Intel to make one.

tskisner:
For the record, I ran a comparison using the `toast_satellite_sim.py` example script and the following parameters:
```
--group-size
4
--nside
256
--no-day-maps
--no-season-maps
--no-sky
--simulate-noise
--sample-rate
20.0
--obs-time-h
1.0
--obs-num
10
--hwp-rpm
5.0
--madam
--madam-iter-max
10
--madam-baseline-length
60.0
--no-destripe
--no-wcov
--no-wcov-inv
--binmap
--hits
--focalplane
../fp_7.pkl
```
With a synthetic focalplane made with
```
toast_fake_focalplane.py \
    --minpix 7 \
    --out fp \
    --fwhm 10.0 \
    --fov 10.0 \
    --psd_fknee 0.1 \
    --psd_alpha 1.0 \
    --psd_fmin 1.0e-5 \
    --psd_NET 1.0 \
    --bandcenter_ghz 150.0 \
    --bandwidth_ghz 40.0
```
I used one cori KNL node with 16 processes.  I ran this script with the `cmbenv-gcc/20200630` and `cmbenv-intel/20200630` software stacks (which have toast 2.3.8) and verified that the outputs agreed.  Next I took the cmbenv-intel stack and compiled this branch of toast using just MKL without any linking to FFTW.  However, note that the pipeline above does have other python packages that link to FFTW (Madam).  This compilation of toast passes all unit tests and the outputs of the above pipeline agree with the other two stacks.

cmake/FindMKL.cmake
tskisner: I added a copyright to the new custom MKL check.

PR #374, opened 2020-11-30
Obs matrix

keskitalo:
Add a new operator to filter and bin signal: `OpFilterBin`. It is also able to calculate the matching observation matrix.


keskitalo:
This PR is ready for review and merge. There are features such as caching of intermediate products that we will want to add but the are not critical right now.

src/toast/todmap/filterbin.py
```diff
@@ -0,0 +1,725 @@
...
+                intervals = obs[self._intervals]
+            else:
+                intervals = None
+            local_intervals = tod.local_intervals(intervals)
+            times = tod.local_times()
+            common_flags = tod.local_common_flags(self._common_flag_name)
+
+            phase = self._get_phase(tod)
+            t1 = time()
+            common_templates = self._build_common_templates(
```
tskisner, line 627: More notes for toast3 porting...  It seems like the `FilterBin` operator could be constructed like:

1. operator which builds common templates for every observation and view
2. operator which builds and regresses per-detector templates
3. operator which accumulates observation matrix
4. Pipeline which runs pointing expansion and (2.) - (3.) one detector at a time

Then the `FilterBin` operator can internally run (1.) and then `BinMap` passing the pipeline in (4.) as the `pre_process` option.  Anyway, I'm sure we'll iterate on this along the way.

tskisner (approved):
Impressive new tool!  Going to be some work to port to toast3, but good to merge this sooner rather than later so that the work can begin.

PR #375, opened 2021-01-29
Support for fixed observing elevations broke SSO scheduling.  This fi…

keskitalo:
…xes it.


tskisner:
Looks good, merge when ready.

PR #376, opened 2021-02-10
Scheduler offset fix

keskitalo:
Fix an issue with applying Az/El offsets to boresight pointing in the scheduler. The offsets need to account for the current boresight elevation.


keskitalo:
Self-merging a change that only affects the scheduler.

PR #378, opened 2021-03-17
sim_det_map.py: OpSimScan: fix resolving backward compat of distmap

ickc:
@keskitalo, @tskisner, this is a simple fix on deprecation of distmap arg and transition to input_map in #358. The original logic doesn't handle distmap correctly.


keskitalo:
Looks good. I'm merging this to clear up old pull requests.

PR #379, opened 2021-03-18
Fix weather params

keskitalo:


tskisner (approved):
These changes are fine.  For toast3, we should have a dedicated operator for creating the weather model in each observation.  I added a card to the porting project for that (https://github.com/hpc4cmb/toast/projects/1#card-57236576).

PR #381, opened 2021-04-06
Gain template fitting

giuspugl:
This PR will include in the toast mapmaker a new template related to the gain fluctuations, notice that this implies a slighter difference with respect to the other mapmaker templates. In fact the gain estimate is not _subtracted_ from the TODs (e.g. as the offsets ) but it's instead _divided_ aiming essentially at `calibrating` the signal by estimating the drift with a coarse signal template, see 
https://github.com/hpc4cmb/toast/blob/91c9b4db927c89b78b15099c1c0c7aca510a1192/src/toast/todmap/mapmaker.py#L409
The overall signal template are included as  linear combination of polynomial legendre up to the 4th order and reference signal . 
https://github.com/hpc4cmb/toast/blob/91c9b4db927c89b78b15099c1c0c7aca510a1192/src/toast/todmap/mapmaker.py#L426


giuspugl:
Just notice that before merging this PR we need finalize the way we want to write to disk the gain amplitudes , so far the implementation is that every processor write them into the same file . 
https://github.com/hpc4cmb/toast/blob/91c9b4db927c89b78b15099c1c0c7aca510a1192/src/toast/todmap/mapmaker.py#L386 
This is a feature useful for debugging but ain't sure that's gonna be used for production runs. Any comment/suggestion ?

keskitalo (approved):
Looks good. Please run `src/format_source.sh` before merging.

PR #385, opened 2021-04-16
port the gainscrambler and unit test

keskitalo:


src/toast/ops/gainscrambler.py
```diff
@@ -0,0 +1,124 @@
...
+        "match the pattern are scrambled.",
+    )
+    center = Float(1, allow_none=False, help="Gain distribution center")
+
+    sigma = Float(1e-3, allow_none=False, help="Gain distribution width")
+
+    realization = Int(0, allow_none=False, help="Realization index")
+
+    component = Int(0, allow_none=False, help="Component index for this simulation")
+
```
tskisner, line 47: This operator does not use any flags or data views.  There are some other operators that do this (e.g. `SimNoise`) because they need to generate a contiguous series of samples.  Ok to leave this as-is, but if we find a situation where we want to only scramble gains over specific intervals then we should add a view trait here.

tskisner (approved):
Thanks for porting this!

PR #387, opened 2021-04-22
WIP: Add totalconvolve

mreineck:
This is a first try to integrate the `ducc0.totalconvolve` code. It seems to produce not completely broken results, but tests still fail.

Not done yet:
 - proper MPI suport (currently every task does the full convolution and stores the entire data cube). I'll need help with that.
 - dipole removal
 - sky smoothing is probably still broken.
 
(I also had to switch off the gainscrambler test manually for some reason. Of course this will be re-enabled before merging!)

Please let me know what you think!


mreineck:
To run this, you will need to have the `more_evolution` branch of `ducc0` installed. Sorry, I'll try to get this merged as soon as possible.

mreineck:
Tests now pass, but I think there are still many details to be discussed.

src/toast/ops/totalconvolve.py
```diff
@@ -0,0 +1,654 @@
...
+        timer.start()
+        sky = self.load_alm(skyfile, self.lmax)
+        fwhm = self.fwhm.to_value(u.radian)
+        # Hmm, healpy doesn't support un-smoothing via negative FWHM, so let's do it by hand ...
+        if fwhm != 0:
+            sigma = fwhm / (2.0 * np.sqrt(2.0 * np.log(2.0)))
+            for i in range(sky.shape[0]):            
+                ell = np.arange(self.lmax + 1.0)
+                s = 2 if i >= 1 else 0
+                fact = np.exp(0.5 * (ell * (ell + 1) - s ** 2) * sigma ** 2)
```
keskitalo, line 382: You could just use `healpy.gauss_beam()` to get the window function. I presume you would want to use the proper polarized definition.

src/toast/ops/totalconvolve.py
```diff
@@ -0,0 +1,654 @@
...
+                sky[i] = hp.sphtfunc.almxfl(sky[i], fact, mmax=self.lmax, inplace=True)
+        if self.remove_monopole:
+            sky[:,0] = 0
```
keskitalo, line 385: This is manipulating T,E and B coefficients, correct? I know they should be zero but if they weren't, this would be a side effect.

src/toast/ops/totalconvolve.py
```diff
@@ -0,0 +1,654 @@
...
+    def _provides(self):
+        prov = self.detector_pointing.provides()
+        prob["detdata"].append(self.det_data)
+        return prov
+
+    def _accelerators(self):
+        return list()
+
+
+class SimWeightedTotalconvolve(SimTotalconvolve):
```
keskitalo, line 570: I thought one of the advantages of the totalconvolve code was that it could natively support the HWP without the kind of trickery we have resort to. Could these two operators be merged into one?

src/toast/tests/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-
+  
 # Install the python files
 
 install(FILES
```
keskitalo, line 4: This list is still missing ops_totalconvolve.py.

src/toast/ops/totalconvolve.py
mreineck: Good point, thanks! In the longer term we should probably get this sort of deconvolution directly into `healpy`.
I think I should get the polarization right via the use of `s` (actually this is copy/paste from from the `healpy` sources with the sign of the exponential flipped).

src/toast/ops/totalconvolve.py
mreineck: Correct, it should be sufficient to just set `sky[0,0]` here.
Anyway, this is such an ugly hack that I don't expect it to survive until merging ...

src/toast/ops/totalconvolve.py
mreineck: Yes, I think so. At the moment my goal is just to produce a "drop-in" (as far as possible) replacement for libconviqt, which should make comparison and validation easier.

src/toast/ops/totalconvolve.py
mreineck: Fixed. This version is really much nicer!

src/toast/ops/totalconvolve.py
mreineck: I changed this to only work on sky_T.

src/toast/tests/CMakeLists.txt
mreineck: Done.

keskitalo:
The conviqt unit tests that your unit tests are based on use a symmetric beam to have an easy analytical result to compare against. I added a new unit test, comparing the two operators with asymmetric beams and a sky with point sources. The results are interesting:

map_comparison

Perhaps the problem is ill-formed but I wonder if the exacerbated ringing in the `totalconvolve` case is an indication of a problem?

mreineck:
I think that the ringing actually originates in the way how the input sky and beam a_lm are constructed. If we plot the logarithm of sky and beam, at least the beam has artifacts with a very similar wavelength:

map_comparison

If the inputs had been contructed in harmonic space exclusively (which is possible, but somewhat more complicated), I think the ringing would go away.

The only accurate way (I can think of) of measuring the quality of the convolution is to pick single pointings and compare the results with a brute-force convolution performed via `rotate_alm` and friends. This is extremely expensive and can only be done for a handful of values, though.

mreineck:
Indeed, if I increase the `nside` parameter at which the fake beam is generated in `create_fake_beam_alm` by a factor of 4, the ringing goes away completely for both operators:
map_comparison

mreineck:
Ah, I see that you are already doing something very similar in `create_fake_sky_alm` already :-)

mreineck:
Question: if we merge the two operators into one, how do we tell the operator (either during construction or in `exec`) whether it should run in HWP mode or not? Would it be OK to inspect `psi_pol` and use HWP mode if it contains any nonzero elements?

mreineck:
I just merged the `more_evolution` branch into the default `ducc0` branch, so you longer need a special branch of `ducc0` to test this.
Before this is merged I'll also make a release on PyPI.

mreineck:
Is there any documentation how the `I000`, `0I00` and `00I0` beam files are generated?

keskitalo:
I think checking if there are non-zero HWP values is acceptable, although it would be even more straightforward to just check if the HWP angle exists at all.

The I00, 0I0 and 00I beam files are IQU beams that have only one non-zero component. There really should be no need for you to deal with them since `totalconvolve` can handle the HWP internally.

Very nice catch on the resolution parameter in creating the `b_lm`. I hope the higher resolution does not make the tests too costly to run.

tskisner:
About the combining of operators into one, we could have a boolean trait that the user can switch to change modes.  Actually we could do that for the conviqt case too rather than having two separate operators.  Does `ducc0.totalconvolve` require MPI?  Just looking at the import try / except clause which only gets executed in the MPI case.

tskisner:
@keskitalo, so you are saying to just check `if self.hwp_angle is not None:`?  Is there a reason we don't do that in the conviqt operators as well to merge them into one?

keskitalo:
@kisner We could certainly perform the same checks in the conviqt operators and merge them. However, that would mean having two very different `_exec()` methods married together. I'm not convinced that it would be more elegant than what we have now in place.

If I recall correctly from the example programs @mreineck made available a while back that the specialization in `totalconvolve` is much more subtle. Please correct me if I'm wrong.

Kisner:
@keskitalo You're tagging the wrong user.

mreineck:
>  Does ducc0.totalconvolve require MPI?

No, it is not MPI-aware at all. All parallelization work needs to be done on the user side (which is the part I have not done yet and probably ned your help with). But it supports working on partial problems as far as that is possible.

mreineck:
Concerning the HWP: we have an algorithm which should emulate an ideal HWP, and it only requires the normal TEB beam a_lm. I'll try to implement this soon in the PR (`ducc0.totalconvolve` doesn't do this by itself, it merely provides the necessary tools).
I'm curious how this will compare to your current approach since it looks quite different on paper.

mreineck:
One more question: how should I deal with multithreading? `libconviqt` just uses `OMP_NUM_THREADS` as far as I understand; should I also read this variable in the constructor of the operator, or would you prefer to pass this as a parameter?

src/toast/ops/totalconvolve.py
```diff
@@ -0,0 +1,695 @@
...
+        """
+        my_dets = set()
+        for obs in data.obs:
+            # Get the detectors we are using for this observation
+            obs_dets = obs.select_local_detectors(detectors)
+            for det in obs_dets:
+                my_dets.add(det)
+            # Make sure detector data output exists
+            obs.detdata.ensure(self.det_data, detectors=detectors)
+        all_dets = self.comm.gather(my_dets, root=0)
```
tskisner, line 301: Here is another case where we need to allow running without MPI.
```
all_dets = my_dets
if self.comm is not None:
    all_dets = self.comm.gather(my_dets, root=0)
    if self.comm.rank == 0:
        for some_dets in all_dets:
            my_dets.update(some_dets)
        my_dets = sorted(my_dets)
    all_dets = self.comm.bcast(my_dets, root=0)

File: src/toast/ops/totalconvolve.py [hunk 6]
```diff
@@ -0,0 +1,695 @@
...
+        pnt[:, 1] = phi
+        pnt[:, 2] = psi + np.pi  # FIXME: not clear yet why this is necessary
+        if verbose:
+            timer.report_clear(f"pack input array for detector {det}")
+        return pnt
+
+    def convolve(self, sky, beam, lmax, mmax, pnt, det, verbose):
+        timer = Timer()
+        timer.start()
+        convolver = totalconvolve.Interpolator(
```
tskisner, line 513: I guess here is where we need to do something different if we have an MPI communicator?  In that case, every process in `self.comm` will have some amount of pointing data (possibly zero samples, if we are working on a detector that does not exist on every process).  And also every process will be doing a portion of the convolution.

Alternatively, if implementing the communication needed for that is too much at the moment, it would also be fine to put a check in the `_exec()` method and raise an exception if `self.comm.size` is greater than one.  This would allow us to run certain types of jobs for testing in the mean time.

tskisner (commented):
Great progress, I guess the question is whether we want to try to support the MPI case now (which I think requires doing some communication and book-keeping here?) or save that for future development and test this operator on certain jobs where only one process is working on each detector.

src/toast/ops/totalconvolve.py
```diff
@@ -0,0 +1,695 @@
...
+from .operator import Operator
+from .. import qarray as qa
+from ..timing import function_timer
+from ..traits import trait_docs, Int, Float, Unicode, Bool, Dict, Quantity, Instance
+from ..utils import Logger, Environment, Timer, GlobalTimers, dtype_to_aligned
+
+
+totalconvolve = None
+
+if use_mpi:
```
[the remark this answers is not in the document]
mreineck: Done.

src/toast/ops/totalconvolve.py
```diff
@@ -0,0 +1,695 @@
...
+
+        log = Logger.get()
+
+        timer = Timer()
+        timer.start()
+
+        all_detectors = self._get_all_detectors(data, detectors)
+
+        for det in all_detectors:
+            verbose = self.comm.rank == 0 and self.verbosity > 0
```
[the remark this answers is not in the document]
mreineck: Oops, I skipped over this comment and solved it slightly differently. Let's see if that works.

src/toast/ops/totalconvolve.py
mreineck: Done.

src/toast/ops/totalconvolve.py
mreineck: If we want to be efficient, we ned to do something more clever in the presence of MPI, but the code will also work (somewhat wastefully) as it is. Every task will do the full convolution work then and store the full data cube. The interpolation part will be parallelized automatically, I guess, due to the pointings being distributed over the tasks already. In short, this simple version will be sort of OK as long as `lmax` and `beammmax` are not very high, and you are processing very many pointings with every call.

The ultimate solution would be to use the "ConvolverPlan" class from `ducc` instead of te simpler `Interpolator` class. Then the approach would be:
- distribute the convolution work among the tasks along the m_beam axis
- redistribute the data such that every task has the full m_beam axis, but only parts of theta
- FFT the m_beam axis into a psi axis
- rearrange the pointings such that every task only has pointings falling into its own theta range
- do the interpolation
- move the resulting data back to its original tasks

However, this is a bit more than I have time for at the moment ...

mreineck:
I have now merged the operators and use our own algorithm to compute the signal in presence of an ideal HWP. However, this latter part has not been validated yet, and the relevant tests in Toast only seem to verify that the code doesn't crash (they don't inspect the results as far as I understand), so this may still be quite buggy.

keskitalo:
This is starting to look really mature. It is true that the HWP-enabled unit test is incomplete for both conviqt and totalconvolve. We could compare them against each other use the analytic (but expensive) calculation Martin mentioned. Any thoughts?

mreineck:
I'm trying to run the tests with MPI and more than one task. What is the official invocation for this?
My attempt was
`mpirun -np 2 python3 -c 'import toast.tests; toast.tests.run()'`
but this hangs in SimConviqtTest...

mreineck:
The convolution part of the algorithm is now distributed as much as possible over MPI tasks. I think it does the right thing, but can't really test it, since the tests don't seem to like MPI.

mreineck:
The only missing parallelization feature is the redistribution of pointings in such a way that every MPI task only needs to stoe a certain theta band. I fear this will be complicated since it requires a few big alltoallv() communications, and I just don't manage to find useful documentation for mpi4py ... the best I could manage is
```
In [3]: mpi4py.MPI.COMM_WORLD.Alltoallv?
Docstring:
Comm.Alltoallv(self, sendbuf, recvbuf)

All to All Scatter/Gather Vector, send data from all to all
processes in a group providing different amount of data and
displacements
Type:      builtin_function_or_method
```
which is a bit on the unhelpful side. Do you have any tips?

mreineck:
> It is true that the HWP-enabled unit test is incomplete for both conviqt and totalconvolve. We could compare them against each other use the analytic (but expensive) calculation Martin mentioned. Any thoughts?

I only know how to compare against brute-force convolution when the HWP is not pesent: given s_lm, b_lm, and a (theta,phi,psi) triple, it is something like:

```
def convolve(alm1, alm2, lmax):
    job = ducc0.sht.sharpjob_d()
    job.set_triangular_alm_info(lmax, lmax)
    job.set_gauss_geometry(lmax+1, 2*lmax+1)
    map = job.alm2map(alm1)*job.alm2map(alm2)
    job.set_triangular_alm_info(0, 0)
    return job.map2alm(map)[0]*np.sqrt(4*np.pi)

blm_rot = ducc0.sht.rotate_alm(blm, lmax, psi, theta, phi)
reference_value = convolve(slm, blm_rot, lmax).real
```
(Ths has to be summed up over T, E and B components.)

I'm not absolutely sure how this changes in the presence of an ideal HWP at angle alpha.

mreineck:
Sorry, yet another question ...

Is there a deeper sense in writing sky and beam a_lm to disk in the main program and reading them in again in the convolution module, or was this done to have a simpler interface to libconviqt?

keskitalo:
Reading the slm and blm from file does come from libconviqt. If they are being synthesized on the fly and are large, we should provide methods for instantiating the conviqt Sky and Beam classes from buffers in memory.

keskitalo:
Thanks for noticing the MPI unit test failure. It is fixed in #391 .

mreineck:
Great, I'm merging this directly!

mreineck:
I can run the tests with 4 tasks now. When I use 8, there seems to be some kind of deadlock inside libconviqt. But this is most likely not relevant for this PR ...

mreineck:
I have a rough idea how to do the pointing redistribution; however, this is a fairly big task and will take some time. So I'd suggest that to review and finalize this PR now and open another one once the redistribution is ready. OK?

One more thing: we can save a factor of two in memory (and potentially gain some speed as well) by computing the interpolation in single precision. Do you think this would be worthwhile? I can support this optionally, e.g. via a `single_precision` parameter for the operator.
Anyway, this isn't really urgent either...

mreineck:
I managed to reduce the memory footprint so that the data cube will now only be stored once on every compute node, regardless of how many MPI tasks are running on the node.
However, this is a little bit magic and I'd like to hear your professional opinion about it ...

src/toast/ops/totalconvolve.py
```diff
@@ -527,15 +527,44 @@ def get_buffer(self, theta, phi, psi, det, verbose):
             timer.report_clear(f"pack input array for detector {det}")
         return pnt
 
-    def conv_and_interpol(self, skycomp, beamcomp, lmax, mmax, pnt, nthreads):
+    # simple approach when there is only one task
+    def conv_and_interpol_scalar(self, skycomp, beamcomp, lmax, mmax, pnt, nthreads):
```
keskitalo, line 531: Presumably this should be serial rather than scalar?

src/toast/ops/totalconvolve.py
```diff
@@ -568,6 +597,100 @@ def conv_and_interpol(self, skycomp, beamcomp, lmax, mmax, pnt, nthreads):
...
+                lmax=lmax,
+                kmax=mmax,
+                sigma=self.oversampling_factor,
+                epsilon=self.epsilon,
+                nthreads=intracomm.size * nthreads,
+            )
+            myrank, nranks = intercomm.rank, intercomm.size
+
+            # get shared memory for the data cube
+            shm = shared_memory.SharedMemory(
```
keskitalo, line 619: TOAST uses https://github.com/tskisner/pshmem for shared memory management on node. @tskisner do you know if there are critical differences between `pshmem` and `multiprocessing.shared_memory`?

I also noticed some warnings from `shared_memory` after running the unit tests:
```
/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 10 leaked shared_memory objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/resource_tracker.py:229: UserWarning: resource_tracker: '/psm_8558dd07': [Errno 2] No such file or directory: '/psm_8558dd07'
  warnings.warn('resource_tracker: %r: %s' % (name, e))
/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/resource_tracker.py:229: UserWarning: resource_tracker: '/psm_177d1d75': [Errno 2] No such file or directory: '/psm_177d1d75'
...
```
I wonder if they are cause for concern.

src/toast/ops/totalconvolve.py
mreineck: Yeah, that makes more sense. Thanks!

src/toast/ops/totalconvolve.py
mreineck: I wasn't aware of psmem ... if it is not too hard for me to switch, I'll do that.

Concerning the warnings, I see them too, but only intermittently, so I have no idea what's going on. Given that every MPI test run ends with 
```
 ----------------------------------------------------------------------
Ran 102 tests in 109.361s

[1] FAILED (errors=1)
[0] FAILED (errors=1)
ICE default IO error handler doing an exit(), pid = 27215, errno = 32
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[62616,1],0]
  Exit code:    1
--------------------------------------------------------------------------
```
I wondered whether there is some more fundamental problem with cleaning up after the tests, so I didn't think about this too much.

mreineck: OK, I think with `pshmem` things become even simpler!
I'm only a bit worried because I'm deliberately not using `pshmem`'s auto-sync feature. I'm just using its `data` member as a numpy array and take care of synchronization across nodes myself. Is that proper usage, @tskisner?

keskitalo: Using `pshmem` at least offers the advantage that you don't need to communicate the pieces of the shared memory between nodes. `pshmem` can maintain synchronized copy of the same buffer on all nodes.

mreineck: I think I won't be able to avoid explicit communication, since I pass sub-arrays of the shared memory region to other functions, and they write to them without knowing that they are special. Also, synchronizing every write access to the buffer would be very expensive, and it is not necessary. If it's at all possible, I'd like to stick to the few explicit `Bcast` statements I currently have.

src/toast/ops/totalconvolve.py
mreineck: Done.

keskitalo:
I just pushed a small fix to support serial versions of the unit tests that were failing. @tskisner I believe this branch is ready for merge.

tskisner (approved):
Thanks for all this work @mreineck !  Will be great to use this new functionality soon.

mreineck:
Thanks a lot!
Please let me know a few days in advance before you release Toast 3, so I can upload the latest ducc0 version to PyPI.

I will continue to look at tests and possible future optimizations. In the meantime,please let me know about any problems/strangeness with the new operator!

PR #390, opened 2021-04-26
Fix two unit test failures. The unit test suite does not run automati…

keskitalo:
…cally when filing a pull request against a branch.


tskisner:
Thanks for the catch.  Merging.

PR #391, opened 2021-04-27
Parallel conviqt tests hung during figure output.

keskitalo:


tskisner:
@keskitalo , was this fixed in #387?

keskitalo:
It was.

PR #392, opened 2021-04-29
Updates to CI Tests

tskisner:
Some code now requires healpy-1.14.0.  Run unit tests on toast3 branch for now as well as master, since we are actively using PRs against that branch.  Remove python-3.6 tests since we now require 3.7.  We can enable 3.9 tests as soon as upstream cmbenv docker container is built.


tskisner:
Rather that require healpy-1.14.0, I just guarded the `map2alm(verbose=...)` call with a try / except.  This PR is mostly fixes to the CI files, so I will self merge after the tests pass.

PR #394, opened 2021-04-29
Gaindrift operator toast3 

giuspugl:
With this operator we port the functionalities developed   for satellite experiments to inject calibration drifts .

The drift can be injected into 3 different ways:
    -  `linear_drift`: inject a linear drift   with a random slope  for each detector
    -  `slow_drift`: inject a drift signal with a `1/f` PSD, simulated up to
    the  frequencies<`cutoff_freq`, in case `cutoff_freq< (1/t_obs)`, `cutoff_freq=1/t_obs`.
    -  `thermal_drift`: inject a drift encoding frequencies up to the sampling rate, to simulate
    the thermal fluctuations in the focalplane.
    Both `slow_drift` and `thermal_drift` modes encode the possibility to inject a common mode drifts
    to all the detectors belonging to a group of detectors identified the string `focalplane_group` ( can
    be any string set by the user used to identify the groups in the detector table).
    The amount of common mode contribution is set by setting detector_mismatch to a value `<1`, (with
    0 being the case with only injecting common mode signal).

https://github.com/hpc4cmb/toast/projects/1#card-59557848


src/toast/ops/sim_gaindrifts.py
```diff
@@ -0,0 +1,322 @@
...
+    )
+    thermal_fluctuation_amplitude = Quantity(
+        1 * u.K,
+        help="Amplitude of thermal fluctuation for `thermal_drift` in  Kelvin units ",
+    )
+    focalplane_Tbath = Quantity(
+        100 * u.mK,
+        help="temperature of the focalplane for `thermal_drift` ",
+    )
+    responsivity_function = Callable(
```
tskisner, line 78: This is a cool trait type that we have not used before.  It will be useful in many operators as a mechanism for user-provided callbacks!

src/toast/ops/sim_gaindrifts.py
```diff
@@ -0,0 +1,322 @@
...
+                        samples=ob.n_local_samples,
+                        freq=freq,
+                        psd=psd,
+                        py=False,
+                    )
+                    thermal_fluct.append(gain)
+                thermal_fluct = np.array(thermal_fluct)
+
+                for det in dets:
+                    detindx = focalplane[det]["uid"]
```
tskisner, line 188: Note that this detector unique ID is not necessarily a simple index.  It may be a very large integer generated from a hash of the detector name:  https://github.com/hpc4cmb/toast/blob/57f57d12cce917da9d87a18d78e1919167bb56af/src/toast/instrument.py#L322
Something to keep in mind when using it for RNG state.

src/toast/ops/sim_gaindrifts.py
```diff
@@ -0,0 +1,322 @@
...
+                    # only if the mismatch !=0
+                    if self.detector_mismatch != 0:
+                        key1 = (
+                            self.realization * 429496123345
+                            + telescope * 6512345
+                            + self.component
+                        )
+                        counter1 = 0
+                        counter2 = 0
+                        key2 = obsindx * 12345676 + detindx
```
tskisner, line 199: The starting state of the RNG is determined from `key1`, `key2`, and `counter1`.  Then each sample increments `counter2`.  Since `detindx` and `obsindx` might be a very large 64bit integer, perhaps we could do this:
```
key1 --> (keep the same- based on realization, telescope, component)
key2 --> obsindx
counter1 --> detindx
counter2 --> 0
```

src/toast/tests/ops_sim_gaindrifts.py
```diff
@@ -0,0 +1,348 @@
...
+            obsindx = obs.uid
+
+            key1 = (
+                drifter.realization * 4294967296 + telescope * 65536 + drifter.component
+            )
+            counter1 = 0
+            counter2 = 0
+            for det in obs.local_detectors:
+                detindx = focalplane[det]["uid"]
+                key2 = obsindx * 4294967296 + detindx
```
tskisner, line 69: Obviously any changes to the operator have to be made here as well...

src/toast/tests/ops_sim_gaindrifts.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # inject gain drift  w/ common mode
+
+        drifter = ops.GainDrifter(
+            det_data=key,
+            drift_mode="thermal_drift",
+            detector_mismatch=0.7,
+        )
+        drifter.apply(data)
+        binner2 = ops.BinMap(
```
tskisner, line 138: This operator is using all the same traits as `binner1` (including the key for the output map in data), so you can just apply the original binner again.

src/toast/tests/ops_sim_gaindrifts.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # inject gain drift  w/ common mode
+
+        drifter = ops.GainDrifter(
+            det_data=key,
+            drift_mode="slow_drift",
+        )
+        drifter.apply(data)
+
+        binner2 = ops.BinMap(
```
tskisner, line 220: Same here, you can just apply the original binner again.

tskisner (approved):
Just a couple small changes, otherwise great!  Thanks for this work.

src/toast/ops/sim_gaindrifts.py
giuspugl: Sounds good ! thanks

PR #395, opened 2021-05-03
Fix a several MPI issues and some typos

tskisner:
- Removing stale debugging barrier() inside the Pipeline operator

- In the mapmaker utils, do the check of existing map domain objects
  before looping over observations, so that processes with different
  numbers of detectors still participate in communication

- In equality test of a PixelDistribution, do not use blocking
  communication.

- In the satellite workflow, do not attempt to use madam unless it
  is available.

### DIFF
--- a/src/toast/ops/mapmaker.py
+++ b/src/toast/ops/mapmaker.py
@@ -221,9 +221,10 @@ class MapMaker(Operator):
             raise RuntimeError("You must set the det_data trait before calling exec()")
 
         # Check map binning
+        map_binning = self.map_binning
         if self.map_binning is None:
             # Use the same binning used in the solver.
-            self.map_binning = self.binning
+            map_binning = self.binning
 
         # We use the input binning operator to define the flags that the user has
         # specified.  We will save the name / bit mask for these and restore them later.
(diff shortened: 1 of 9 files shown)


src/toast/ops/mapmaker.py
```diff
@@ -221,9 +221,10 @@ def _exec(self, data, detectors=None, **kwargs):
             raise RuntimeError("You must set the det_data trait before calling exec()")
 
         # Check map binning
+        map_binning = self.map_binning
```
tskisner, line 224: A note on this change:  I am doing this with a local variable so that the original value of the `self.map_binning` trait is preserved instead of being overwritten.

tskisner:
@keskitalo and @nestordemeure , I am requesting reviews from both of you- for Reijo, this touches some of the mapmaker utilities, so wanted you to take a look.  @nestordemeure , it would be good to confirm that this fixes your issue.

nestordemeure (approved):
Good for me, it solves the hang-up I had with the satellite workflow.

src/toast/ops/mapmaker_utils.py
```diff
@@ -120,6 +120,8 @@ def _exec(self, data, detectors=None, **kwargs):
                 "Building hit map with pixel_distribution {}".format(self.pixel_dist)
             )
 
+        pstr = f"Proc ({data.comm.world_rank}, {data.comm.group_rank})"
```
keskitalo, line 123: I don't think this string is used anywhere.

keskitalo (approved):

PR #396, opened 2021-05-11
 [toast3] satellite benchmark

nestordemeure:
This PR adds the sattelite benchmarking script.

It also adds an additional `info` member to the loger.


src/toast/scripts/toast_benchmark_satellite
```diff
@@ -0,0 +1,523 @@
...
+    compute_science_metric(args, global_timer, n_nodes, rank, log)
+
+    # dumps all the timing information
+    timer = toast.timing.GlobalTimers.get()
+    timer.start("toast_benchmark_satellite (gathering and dumping timing info)")
+    alltimers = gather_timers(comm=world_comm)
+    if comm.world_rank == 0:
+        out = os.path.join(args.out_dir, "timing")
+        dump(alltimers, out)
+        with open(os.path.join(args.out_dir, "log"), "a") as f:
```
tskisner, line 502: We can remove this appending of the timing to the log file.  It is historical and actually assumes that the output of this script is being piped to a file called "log" in the output directory, which is not always the case.

src/toast/scripts/toast_benchmark_satellite
```diff
@@ -0,0 +1,523 @@
...
+        msg = f"Science Metric: {prefactor:0.1e} * ({args.n_detector:d}**{det_factor:0.2f}) * ({kilo_samples:0.3e}**{sample_factor:0.3f}) / ({runtime:0.1f} * {n_nodes}) = {metric:0.2f}"
+        log.info("")
+        log.info(msg)
+        log.info("")
+        with open(os.path.join(args.out_dir, "log"), "a") as f:
+            f.write(msg)
+            f.write("\n\n")
+
+def main():
+    env = toast.utils.Environment.get()
```
tskisner, line 407: Since this is a benchmark, we should probably always enable function timers.  At this point in the code we could do:
```
env.enable_function_timers()
```

tskisner (approved):
Looks good, just a couple small changes and running the source formatter.  Thanks for this work!

src/toast/utils.py
```diff
@@ -39,6 +39,17 @@
 
 from .mpi import MPI, use_mpi
 
+from ._libtoast import Logger
+
+def _info_rank0(self, message, comm):
```
keskitalo, line 44: I really like this.

keskitalo (approved):
Looks great! @tskisner the source formatter is not formatting the scripts.

tskisner:
Yes, I asked @nestordemeure offline to fix the `format_source.sh` file as part of this.  We just need to replace the `pipelines` directory with `workflows`.

src/toast/scripts/toast_benchmark_satellite
nestordemeure: keeping line 500-501 I guess? (removing the log part but not the writing of timers to `timing`)

nestordemeure:
I just pushed a new version taking the reviews into account. @tskisner do you see things to do before merging?

tskisner:
Thanks @nestordemeure , merge when ready.

PR #397, opened 2021-05-11
Toast3 Ground simulation support

tskisner:
Not yet ready for review.  Porting of simulated ground scanning.  Elevation noise and interval flagging are extracted as separate operators.

### DIFF
--- a/src/toast/ops/elevation_noise.py
+++ b/src/toast/ops/elevation_noise.py
@@ -209,13 +209,20 @@ class ElevationNoise(Operator):
 
                 el = np.median(np.concatenate(el_view))
 
-                # PSD
+                # Scale the PSD
+
+                # The PSD in a fixed set of units
                 psd_ksq = noise.psd(det).to_value(u.K ** 2 * u.second)
-                rate = noise.rate(det)
+
+                # Use the high-frequency part of the PSD to estimate the NET^2
                 old_net_sq = np.median(psd_ksq[-10:])
+
+                # Compute the new NET in the same units.
                 new_net = self.noise_a.to_value(u.K * (1.0 * u.second)) / np.sin(
                     el
                 ) + self.noise_b.to_value(u.K * (1.0 * u.second))
+
+                # Apply scaling
                 psd[:] *= new_net ** 2 / old_net_sq
         return
 
--- a/src/toast/ops/elevation_noise.py
+++ b/src/toast/ops/elevation_noise.py
@@ -1,0 +1,237 @@
+# Copyright (c) 2015-2020 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+import copy
+
+import numpy as np
+
+import traitlets
+
+from ..utils import Environment, Logger
+
+from ..timing import function_timer, Timer
+
+from ..noise_sim import AnalyticNoise
+
+from ..traits import trait_docs, Int, Unicode, Float, Bool, Instance, Quantity
+
+from .. import qarray as qa
+
+from .operator import Operator
+
+
+@trait_docs
+class ElevationNoise(Operator):
+    """Modify detector noise model based on elevation.
+
+    This adjusts the detector NET in a noise model based on the median elevation of
+    each detector in each observation.
+
+    The new NET value is computed as:
+
+    .. math::
+        NET_{new} =  1.0e-12 x f_{sample} x (a / sin(el) + b)^2 / NET_{old}
+
+    Where the old NET is estimated from the high frequency samples of the PSD.
+
+    NOTE: since this operator generates a new noise model for all detectors, you
+    should specify all detectors you intend to use downstream when calling exec().
+
+    If the view trait is not specified, then this operator will use the same data
+    view as the detector pointing operator when computing the pointing matrix pixels
+    and weights.
+
+    If the output model is not specified, then the input is modified in place.
+
+    """
+
+    # Class traits
+
+    API = Int(0, help="Internal interface version for this operator")
+
+    noise_model = Unicode(
+        "noise_model", help="The observation key containing the input noise model"
+    )
+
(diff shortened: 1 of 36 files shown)


tskisner:
Summary of this work:

- New SimGround operator which combines functionality of old TODGround class and code in the pipeline_tools package.
- Az / El to RA / DEC pointing conversion now in a separate coordinates file.  Currently using pyephem for agreement with old code but an alternate astropy implementation is also there and can be used after more consistency testing.
- Low-level scan motion simulation is split out into a separate source file to keep the file size down.
- Elevation dependent noise model is now a separate operator.
- New FlagIntervals operator that takes intervals and updates bits of a shared flag field.
- New basic ground simulation workflow that is very similar to the satellite workflow for now.  Obviously still need to add ATM simulation and other mapmaking operations.
- Slight tweak to schedule classes to add the telescope and site names, for reference.

tskisner:
@keskitalo , it will be great to have your review on this.  I am not sure what kinds of other unit tests should be done, but we definitely need a more extensive higher-level testing campaign against toast2.  Essentially verifying equivalent hit maps for the same schedule.

src/toast/ops/elevation_noise.py
```diff
@@ -0,0 +1,237 @@
...
+    )
+
+    view = Unicode(
+        None, allow_none=True, help="Use this view of the data in all observations"
+    )
+
+    noise_a = Float(
+        None,
+        allow_none=True,
+        help="Evaluate noise PSD as (a / sin(el) + b) ** 2 * fsample * 1e-12",
```
keskitalo, line 74: This expression is just the NET, not the full PSD. Also, somewhere we should define that `(a / sin(el) + b)` yields the NET in uK_CMB rts.

src/toast/ops/elevation_noise.py
```diff
@@ -0,0 +1,237 @@
...
+from .. import qarray as qa
+
+from .operator import Operator
+
+
+@trait_docs
+class ElevationNoise(Operator):
+    """Modify detector noise model based on elevation.
+
+    This adjusts the detector NET in a noise model based on the median elevation of
```
keskitalo, line 28: We scale the full PSD, not just the NET.

src/toast/ops/elevation_noise.py
```diff
@@ -0,0 +1,237 @@
...
+                    else:
+                        el_view.append(np.pi / 2 - theta[view_flags[vw] == 0])
+
+                el = np.median(np.concatenate(el_view))
+
+                # PSD
+                psd = noise.psd(det)
+                rate = noise.rate(det)
+                old_net = np.median(psd[-10:])
+                new_net = (self.noise_a / np.sin(el) + self.noise_b) ** 2 * rate * 1e-12
```
keskitalo, line 211: The original comment in TOAST2 would make this more readable (paraphrasing):
```
# The (a, b) model evaluates to uK / sqrt(Hz)
# Translate it to K_CMB ** 2 and determine the appropriate scaling of the PSD
```

src/toast/ops/flag_intervals.py
```diff
@@ -0,0 +1,141 @@
...
+
+    view_mask = List(
+        None, allow_none=True, help="List of tuples of (view name, bit mask)"
+    )
+
+    shared_flags = Unicode(
+        None, allow_none=True, help="Observation shared key for flags to modify"
+    )
+
+    shared_flag_bytes = Int(
```
keskitalo, line 43: This is a great improvement over default 8-bit flag fields!

src/toast/ops/sim_ground.py
```diff
@@ -62,17 +71,29 @@ class SimGround(Operator):
...
+        1.0 * u.degree / u.second, help="The sky azimuth scanning rate"
+    )
 
     scan_rate_el = Quantity(
         None, allow_none=True, help="The sky elevation scanning rate"
     )
 
-    scan_accel = Quantity(
+    scan_accel_az = Quantity(
         0.1 * u.degree / u.second ** 2,
```
keskitalo, line 97: Can we make the default 1.0 deg/s^2? Turns out the telescopes turn pretty fast!

src/toast/ops/sim_ground.py
```diff
@@ -239,23 +281,14 @@ def _exec(self, data, detectors=None, **kwargs):
             scan_offsets.append(off)
             off += ns
 
+        # Ensure that astropy IERS is downloaded
+        astropy_control(max_future=self.schedule.scans[-1].stop)
```
keskitalo, line 285: Even if not using astropy for coordinate transforms?

workflows/toast_sim_ground.py
```diff
@@ -0,0 +1,311 @@
...
+
+    ops.sim_ground.telescope = telescope
+    ops.sim_ground.schedule = schedule
+    ops.sim_ground.apply(data)
+
+    # Construct a "perfect" noise model just from the focalplane parameters
+
+    ops.default_model.apply(data)
+
+    # Create the Elevation-weighted noise model
```
keskitalo, line 215: Elevation-modulated?

workflows/toast_sim_ground.py
```diff
@@ -0,0 +1,311 @@
...
+        ops.scan_map.apply(data)
+
+    # Simulate detector noise
+
+    if ops.sim_noise.enabled:
+        ops.sim_noise.apply(data)
+
+    # Build up our map-making operation from the pieces- both operators configured
+    # from user options and other operators.
+
```
keskitalo, line 256: This workflow is still missing all support for filter+bin. That is fine for now but it would be helpful to add comments to where that would go in the file.

keskitalo (approved):
Looks great! I made a number of small observations about the comments to make the source more readable.

src/toast/ops/elevation_noise.py
tskisner: Class docstring is updated.  These traits also now have units.

PR #398, opened 2021-05-11
Gain fluctuation template fitting for  toast3 mapmaker 

giuspugl:
This aims at porting the gain template fitting approach ( already implemented in #381 )  into toast3. 
It presents minor  differences to the implementation shown in #381 . To be further debugged,  we might need to merge firstly the #394  so that we can try to inject a gain fluctuation and then mitigate it with the template fitting approach.  
Notice that this template has an extra method 
https://github.com/hpc4cmb/toast/blob/8dd5b4ba15ee75ce2a7d6fc50da91c7278b046f0/src/toast/templates/gaintemplate.py#L219 
which aims exactly at recalibrating the detector data once the gain fluctuation amplitudes are estimated, this template fitting is therefore advised to be run **before**  other templates are accounted for (e.g. offset baselines, 2d polynomials, etc.. ).

### DIFF
--- a/src/toast/ops/__init__.py
+++ b/src/toast/ops/__init__.py
@@ -55,7 +55,7 @@ from .mapmaker_utils import (
 
 from .mapmaker_binning import BinMap
 
-from .mapmaker_templates import TemplateMatrix
+from .mapmaker_templates import TemplateMatrix, SolveAmplitudes
 
 from .mapmaker import MapMaker, Calibrate
 
--- a/src/toast/templates/gaintemplate.py
+++ b/src/toast/templates/gaintemplate.py
@@ -24,12 +24,11 @@ import scipy
 @trait_docs
 class GainTemplate(Template):
     """This class aims at fitting and mitigating gain fluctuations in the data.
-    The fluctuations  are modeled as a linear combination of Legendre polynomials (up to a given order, commonly
-    `n<5` ) weighted by the so called _gain amplitudes_. The gain template is therefore obtained by estimating
-    the polynomial amplitudes by assuming a _signal estimate_ provided by a template map (encoding a coarse
-    estimate of the underlying signal. )
-
-
+    The fluctuations  are modeled as a linear combination of Legendre polynomials (up
+    to a given order, commonly `n<5` ) weighted by the so called _gain amplitudes_.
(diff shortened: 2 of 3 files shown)


giuspugl:
As already I mentioned above,
the gain template has an extra method 
https://github.com/hpc4cmb/toast/blob/8dd5b4ba15ee75ce2a7d6fc50da91c7278b046f0/src/toast/templates/gaintemplate.py#L219 
which aims exactly at recalibrating the detector data once the gain fluctuation amplitudes are estimated and it has to be performed **before** the destriper offsets    other templates are subtracted  (e.g. offset baselines, 2d polynomials, etc.. ).

In TOAST2, we hard-coded this calibration step in 
 https://github.com/hpc4cmb/toast/blob/5c61f413a2bc6350923dbdc14cd373becdbab632/src/toast/pipeline_tools/mapmaker.py#L371
  https://github.com/hpc4cmb/toast/blob/5c61f413a2bc6350923dbdc14cd373becdbab632/src/toast/todmap/mapmaker.py#L1848
and 
https://github.com/hpc4cmb/toast/blob/5c61f413a2bc6350923dbdc14cd373becdbab632/src/toast/todmap/mapmaker.py#L1970


Essentially we intialize a mapmaker operator encoding as templates _only_ the gain amplitudes, it then calibrate (by dividing the signal by  the estimated gain) only if a template for the signal estimate is provided, (raising an error otherwise ). Once the signal is calibrated the mapmaker proceed by fitting and subtracting the provided templates as usual. 

Currently the Gain template operator has this method implemented, but not sure where is the right place to start implementing the calibration step in the mapmaker .

tskisner:
Ok @giuspugl , I rebased this against `toast3` to get the `Calibrate` operator.  See if you think the unit tests are ok.

giuspugl:
This is really great! thanks a lot @tskisner ! The test to me seems fine, i can include an extra assertion which checks that in absence of gain mismatches the PCG for the Calibrator should converge within 1iterations ( checked the logs and it did!) What do you think ?

tskisner:
This PR has been left hanging for a while.  We can always add more / better unit tests in a new small PR, but lets merge this before it gets out of date.

PR #399, opened 2021-05-14
Function timing improvements

tskisner:
The timers used in toast are designed to provide a built-in, "mid-level" view of the overall runtime of different pieces of the code.  They are used around functions and blocks of code which do a non-trivial amount of work, since starting and stopping these timers introduces a small amount of overhead.  The idea is to quickly give a global picture of a workflow's performance in order to know where to focus performance tuning effort using dedicated profiling tools.

Our current function_timer decorator simply gets the wrapped function's name and uses that for the global timer name.  However, this does not distinguish between the same function being called through a different call trace.  A full call trace is too verbose for the intended purpose of these timers, but it would be useful to group calls to a low-level function based on the call trace.

This work adds a "simplified" call trace to the timer name that:
- Cuts any stack frames not in the toast package
- Cuts out base-class wrapper functions (e.g. Operator.exec(), Template.add_prior(), etc)
- Includes just the function or class.method names (no source file or line numbers)

There is also a simple plotting script that can read a dumped CSV file and make a sunburst plot.

Some notes:
- Apparently enabling function timers with `Environment.enable_function_timers()` does not work.  We can figure that out at some point.  For now, they can be enabled with (for example) `TOAST_FUNCTIME=1 toast_benchmark_satellite --case tiny`.
- The plotting script uses pandas and plotly, but I don't think we want to make those package dependencies for all of toast.  We can add them to the cmbenv stack, but for people running on their local machines they can pip / conda install them if they want to use the script.

Here are some examples from running the tiny benchmark case on my laptop:
```
$>  cat toast_benchmark_satellite_out/timing.csv 
Timer,Processes,Minimum Calls,Maximum Calls,Minimum Time,Maximum Time,Mean Time,Median Time
(function) BuildPixelDistribution._exec,1,1,1,0.128526919,0.128526919,0.128526919,0.128526919
(function) BuildPixelDistribution._exec|Pipeline._exec|PointingHealpix._exec,1,2,2,0.120071915,0.120071915,0.120071915,0.120071915
(function) BuildPixelDistribution._exec|Pipeline._exec|PointingHealpix._exec|PointingDetectorSimple._exec,1,2,2,0.011679545999999999,0
.011679545999999999,0.011679545999999999,0.011679545999999999
(function) MapMaker._exec,1,1,1,16.514301034,16.514301034,16.514301034,16.514301034
(function) MapMaker._exec|BinMap._exec,1,1,1,1.285735414,1.285735414,1.285735414,1.285735414
(function) MapMaker._exec|BinMap._exec|Pipeline._exec|BuildNoiseWeighted._exec,1,2,2,0.126040968,0.126040968,0.126040968,0.126040968
(function) MapMaker._exec|BinMap._exec|Pipeline._exec|BuildNoiseWeighted._exec|PixelDistribution.global_pixel_to_submap,1,2,2,0.006473
804,0.006473804,0.006473804,0.006473804
(function) MapMaker._exec|BinMap._exec|Pipeline._exec|Pipeline._exec|TemplateMatrix._exec,1,2,2,0.001856417,0.001856417,0.001856417,0.
001856417
(function) MapMaker._exec|BinMap._exec|Pipeline._exec|Pipeline._exec|TemplateMatrix._exec|Offset._add_to_signal,1,2,2,0.00054239,0.000
54239,0.00054239,0.00054239
(function) MapMaker._exec|BinMap._exec|Pipeline._exec|PointingHealpix._exec,1,2,2,0.137313905,0.137313905,0.137313905,0.137313905
(function) MapMaker._exec|BinMap._exec|Pipeline._exec|PointingHealpix._exec|PointingDetectorSimple._exec,1,2,2,0.013350357,0.013350357
,0.013350357,0.013350357
...
```

```
$>  toast_timing_plot --file toast_benchmark_satellite_out/timing.csv 
Reading input timing from toast_benchmark_satellite_out/timing.csv
(output saved to timing.pdf, see attached plot)
```
[timing.pdf](https://github.com/hpc4cmb/toast/files/6480270/timing.pdf)
timing

### DIFF
--- a/src/toast/covariance.py
+++ b/src/toast/covariance.py
@@ -16,6 +16,22 @@ from ._libtoast import (
 from .pixels import PixelData
 
 
+# These are just wrappers to time the libtoast calls
+@function_timer
+def libtoast_cov_mult_diag(*args):
+    cov_mult_diag(*args)
+
+
+@function_timer
+def libtoast_cov_apply_diag(*args):
+    cov_apply_diag(*args)
+
+
+@function_timer
+def libtoast_cov_eigendecompose_diag(*args):
+    cov_eigendecompose_diag(*args)
+
+
 def create_local_invert(n_pix_submap, mapnnz, threshold, rcond, invert=False):
     """Generate a function for inverting locally owned submaps of a covariance.
 
@@ -30,6 +46,7 @@ def create_local_invert(n_pix_submap, mapnnz, threshold, rcond, invert=False):
 
     """
 
+    @function_timer
     def local_invert(n_submap_value, receive_locations, receive, reduce_buf):
         # Locally invert owned submaps
         for sm, locs in receive_locations.items():
@@ -43,7 +60,7 @@ def create_local_invert(n_pix_submap, mapnnz, threshold, rcond, invert=False):
             else:
                 rcond.reduce_buf[:] = 0.0
                 rdata = rcond.reduce_buf
-            cov_eigendecompose_diag(
+            libtoast_cov_eigendecompose_diag(
                 1,
                 n_pix_submap,
                 mapnnz,
@@ -110,7 +127,7 @@ def covariance_invert(npp, threshold, rcond=None, use_alltoallv=False):
             rdata = np.empty(shape=0, dtype=np.float64)
         else:
             rdata = rcond.raw
-        cov_eigendecompose_diag(
+        libtoast_cov_eigendecompose_diag(
             npp.distribution.n_local_submap,
             npp.distribution.n_pix_submap,
             mapnnz,
@@ -136,6 +153,7 @@ def create_local_multiply(n_pix_submap, mapnnz, other):
 
     """
 
+    @function_timer
     def local_multiply(n_submap_value, receive_locations, receive, reduce_buf):
         for sm, locs in receive_locations.items():
             # We have multiple copies of submap data- we will multiply just the first
@@ -144,7 +162,7 @@ def create_local_multiply(n_pix_submap, mapnnz, other):
             reduce_buf[:] = receive[locs[0] : locs[0] + n_submap_value]
             other_buf = other.reduce_buf
             other_buf[:] = other.receive[locs[0] : locs[0] + n_submap_value]
-            cov_mult_diag(1, n_pix_submap, mapnnz, reduce_buf, other_buf)
+            libtoast_cov_mult_diag(1, n_pix_submap, mapnnz, reduce_buf, other_buf)
             for lc in locs:
                 receive[lc : lc + n_submap_value] = reduce_buf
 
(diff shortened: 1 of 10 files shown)


src/toast/covariance.py
```diff
@@ -16,6 +16,22 @@
 from .pixels import PixelData
 
 
+# These are just wrappers to time the libtoast calls
```
keskitalo, line 19: Could the decorator not be combined with the pybind11 wrapper?

src/toast/timing.py
```diff
@@ -17,20 +20,55 @@
 from .utils import Environment
 
 
+stack_toast_file_pat = re.compile(r".*(toast.*)\.py")
+
+stack_toast_skip = [
```
keskitalo, line 25: This list could do with some comments. Why where these functions singled out?

keskitalo (approved):

src/toast/covariance.py
tskisner: Good point, I will investigate the feasibility.

nestordemeure (approved):

src/toast/timing.py
tskisner: I agree, it does seem arbitrary.  Basically these are cases where a base class method is just calling an internal derived class method and we don't want to clutter things.  However, I will think if there is some way to register these skipped names in the individual classes where they are defined.

tskisner: I have added a new decorator which adds a function / method to this global list of things to skip.  This is now used for the `Operator` and `Template` base class methods that are simply dispatch methods to call derived class code (so we don't have to see them in this simple trace).

src/toast/covariance.py
tskisner: After more investigation, I think doing this within the compiled extension would be much more painful than these trivial decorator-wrappers.  Another option would be to have an middle-layer "libtoast.py" which imported everything from the `_libtoast` compiled extension and selectively decorated some functions.  Then all of the current code which imports `_libtoast` objects would instead import `libtoast` to get those.

That is a tedious, but straightforward change.  I am tempted to just leave these 3 wrappers for now, since this is the only python file where those functions are used.  If we find we want to decorate more compiled code we could implement that mid-level python file in a future PR.  Thoughts?

keskitalo: Perfect fine by me.

nestordemeure: > That is a tedious, but straightforward change. I am tempted to just leave these 3 wrappers for now, since this is the only python file where those functions are used. If we find we want to decorate more compiled code we could implement that mid-level python file in a future PR. Thoughts?

Seems like a good middle ground : no  need to do a lot of modification that will not be useful.

PR #402, opened 2021-05-26
[Toast3] ground benchmark

nestordemeure:
Adding the benchmarking code for the ground workflow, factorizing some benchmarking functionalities in a dedicated file and updating ground scheduler parameters in a test file.


src/toast/scripts/benchmarking_utilities.py
```diff
@@ -0,0 +1,370 @@
+# Copyright (c) 2015-2021 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+"""
+Functions common to all benchmarking scripts.
```
tskisner, line 6: Great to have these common functions in a separate file, thanks!

tskisner (approved):
Looks good!  I only pushed a commit that ran format_source.sh will the latest versions of uncrustify and black.  Those will be in the cmbenv stack on cori soon.

src/toast/tests/ops_sim_ground.py
```diff
@@ -81,7 +81,7 @@ def test_exec(self):
                     "--site-alt",
                     "{}".format(site.earthloc.height.to_value(u.meter)),
                     "--patch",
-                    "small_patch,1,40,-40,44,-44",
+                    "tile000,1,40,0,60,-40",
```
keskitalo, line 84: The unit test should stick to the `small_patch` target to keep the running time short.

src/toast/unported/tests/ops_dipole.py
```diff
@@ -149,13 +149,28 @@ def test_sim(self):
         # construct distributed maps to store the covariance,
         # noise weighted map, and hits
 
-        invnpp = DistPixels(self.data, nnz=1, dtype=np.float64, nest=False,)
+        invnpp = DistPixels(
```
keskitalo, line 152: `black` seems to have a hard time deciding how to format these lines.

keskitalo (commented):
Looks good. I, too, appreciate putting common benchmark code into a separate file.

If you can restore the original patch definition for the unit test, that would make more sense. I think we only need a "realistic" patch definition for benchmarking.

src/toast/tests/ops_sim_ground.py
nestordemeure: Ok, i will revert it now.

src/toast/unported/tests/ops_dipole.py
nestordemeure: Yes, not a fan of what it did to some of the code...

tskisner: The challenge is that black defaults change slightly between versions, and we are all using different versions.  I have just updated the version in the latest cmbenv stack at nersc, and have installed the latest version locally.  It is slightly painful, but I'll just remind everyone the utter formatting hell that we went through in the recent past.  I feel these annoyances are much better than inconsistency everywhere...

PR #403, opened 2021-06-01
Implementing Cosmic ray operator

giuspugl:
This operator aims at injecting cosmic ray glitches to TOD. This class can be of particular interest for space and baloon observations. 
For time being the operator relies on the user to provide the data encoding the glitch distribution as well as the amplitude of noise due to the low amplitude hits in the focalplane and the correlation among detectors. 

Several things need to be addressed before merging : 
- [ ]  Random number generators, so far am using the toast random number generator for gaussian numbers, however in https://github.com/hpc4cmb/toast/blob/1ec1c2e925cbf36dbceee9a87c4576f06936ab34/src/toast/ops/sim_cosmic_rays.py#L231, https://github.com/hpc4cmb/toast/blob/1ec1c2e925cbf36dbceee9a87c4576f06936ab34/src/toast/ops/sim_cosmic_rays.py#L128  and in https://github.com/hpc4cmb/toast/blob/1ec1c2e925cbf36dbceee9a87c4576f06936ab34/src/toast/ops/sim_cosmic_rays.py#L236  using the `numpy` random numbers ,
- [ ] a traits  that can be very useful is the one encoding the conversion factor to convert cosmic ray tod data (which usually are in power units), to temperature units . 
- [ ] to properly design a test for this operator can include a routine that generate on the fly a mock cosmic ray datafile

### DIFF
--- a/src/toast/ops/sim_cosmic_rays.py
+++ b/src/toast/ops/sim_cosmic_rays.py
@@ -1,0 +1,284 @@
+# Copyright (c) 2019-2020 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+# Copyright (c) 2015-2020 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+import traitlets
+
+import numpy as np
+from scipy import interpolate, signal
+
+from ..timing import function_timer
+
+from ..traits import trait_docs, Int, Float, Unicode, Bool, Quantity, Callable
+from astropy import units as u
+
+from .operator import Operator
+
+from ..utils import Environment, Logger
+from .. import rng
+
+
+
+class InjectCosmicRays(Operator):
+    """
+    Inject the cosmic rays  signal into the TOD. So far we inject two kinds of cosmic ray noise:
+
+    Wafer noise, due to ~400 impacts per second in the wafer   undistinguishable  individually.
+    For each observation and for each detector we inject low noise component as a _white noise signal_,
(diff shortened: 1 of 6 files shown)


src/toast/ops/sim_cosmic_rays.py
```diff
@@ -0,0 +1,294 @@
...
+                counter1 = detindx
+
+                rngdata = rng.random(
+                    size,
+                    sampler="gaussian",
+                    key=(key1, key2),
+                    counter=(counter1, counter2),
+                )
+                filename = self.crfile.replace("detector", f"det{kk}")
+                data_dic = self.load_cosmic_ray_data(filename)
```
tskisner, line 173: This is fine for now, but imagine we had a job running with 1000 detectors for a year, with one observation per hour.  This job would load per-detector files 8760000 times from across all processes.  We should eventually look at how big these files are and see if we should put all the information into one file and load it once and broadcast it to all detectors.

tskisner (changes requested):
Thanks for this work @giuspugl !  Just a couple small things about incrementing the `counter2` values for a given stream after sampling from the stream.  We can wait to evaluate the file format until it becomes a bottleneck on large jobs.

tskisner:
We should rebase this against the `toast3` branch to resolve the conflicts.

src/toast/ops/sim_cosmic_rays.py
giuspugl: Totally agree with this.  the way the CRs are currently  injected is by inspecting the  detector properties stored on each data file.  however, the glitch statistical properties  as well as the wafer noise and the correlation matrix can be easily extended to all the detectors by simply reading them from a single file and broadcast it. On the other hand, we want to leave the user the possibility to inject CRs as they prefer. Can include an extra-trait taking care of  this.. what do you think?

PR #404, opened 2021-06-01
[Toast3] automatic benchmark scaling

nestordemeure:
Introduce a code that sets the size of the benchmark (number of sample) as a function of the available memory. Usefull to study weak scaling.

Examples (on a Cori KNL node) compared to preset benchmark sizes:

```
        tiny: 5000000 samples
       small: 500000000 samples
 auto 1 node: 4330268249 samples
      medium: 5000000000 samples
 auto 5 node: 21672119611 samples
       large: 50000000000 samples
auto 50 node: 217118078215 samples
      xlarge: 500000000000 samples
      heroic: 5000000000000 samples
```

PR #405, opened 2021-06-07
Toast3 simplify

tskisner:
This PR attempts to remove boilerplate from workflow code by moving some logic into the operators.  It also breaks up the `main()` function in the workflows for easier reading.

The existing workflow "style" copied the paradigm from toast2 which used a small number of very flexible workflows that could simulate any supported component and apply nearly all the possible reduction operations.  This kind of script is useful for production use with generic simulations.  For learning and interactive hacking it can be easier to just have a simple script with options directly in the script.  This allows a user to just copy the script and start changing it.  I have now added a satellite and ground workflow that follows this technique.

I will rebase this after #404 is merged, since it touches the same source file.

### DIFF
--- a/src/toast/ops/mapmaker.py
+++ b/src/toast/ops/mapmaker.py
@@ -2,6 +2,8 @@
 # All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
 
+import os
+
 import traitlets
 
 import numpy as np
@@ -230,7 +232,7 @@ class MapMaker(Operator):
 
         # Check map binning
         map_binning = self.map_binning
-        if self.map_binning is None:
+        if self.map_binning is None or not self.map_binning.enabled:
             # Use the same binning used in the solver.
             map_binning = self.binning
 
@@ -653,7 +655,7 @@ class MapMaker(Operator):
             for prod in ["map", "hits", "cov", "rcond"]:
                 dkey = "{}_{}".format(self.name, prod)
                 file = os.path.join(self.output_dir, "{}.fits".format(dkey))
-                write_healpix_fits(data[dkey], file, nest=self.pointing.nest)
+                write_healpix_fits(data[dkey], file, nest=map_binning.pointing.nest)
 
         self._log_info(comm, rank, "  finished output write in", timer=timer)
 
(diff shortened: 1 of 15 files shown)


src/toast/ops/mapmaker.py
```diff
@@ -637,6 +647,18 @@ def _exec(self, data, detectors=None, **kwargs):
 
         self._log_info(comm, rank, "  finished final binning in", timer=timer)
 
+        # Write the outputs
+        # FIXME:  This all assumes the pointing operator is an instance of the
+        # PointingHealpix class.  We need to generalize distributed pixel data
+        # formats and associate them with the pointing operator.
+        if self.output_dir is not None:
```
keskitalo, line 654: Big thumbs up for mapmaker writing out the maps. Yes, not as flexible as doing it in the pipeline but will eliminate so many lines of repeated code.

src/toast/ops/operator.py
```diff
@@ -39,7 +39,13 @@ def exec(self, data, detectors=None, **kwargs):
             None
 
         """
-        self._exec(data, detectors=detectors, **kwargs)
+        log = Logger.get()
+        if self.enabled:
```
keskitalo, line 43: Awesome! Another place to remove boilerplate from the pipeline scripts.

src/toast/ops/scan_healpix.py
```diff
@@ -85,6 +92,16 @@ def _exec(self, data, detectors=None, **kwargs):
         if self.file is None:
             raise RuntimeError("You must set the file trait before calling exec()")
 
+        # Construct the pointing distribution if it does not already exist
+
+        if self.pixel_dist not in data:
```
keskitalo, line 97: Nice!

workflows/toast_sim_ground.py
```diff
@@ -221,86 +168,135 @@ def main():
...
+    ops.mapmaker.output_dir = args.out_dir
+
+    ops.mapmaker.apply(data)
+
+    # Optionally run Madam
+    if toast.ops.madam.available():
         ops.madam.apply(data)
 
 
+def main():
```
keskitalo, line 228: This is a very clean main program!

keskitalo (approved):

PR #408, opened 2021-06-18
Observe elevation limits in SSO scheduling

keskitalo:
Many thanks to @Zhilei for reporting this issue!


keskitalo:
Self-merging this trivial fix.

Zhilei:
Thanks for looking into this! It does not seem like this has been merged into the `master` branch yet, should I use the working branch instead?

Zhilei:
Thanks for pulling the change! I just tested the new code, but it seems like we still have observations scheduled outside of the elevation range...
I apologize for being so annoying...

PR #412, opened 2021-07-08
Port atmosphere simulations

tskisner:
This extensive PR ports the atmosphere simulation code to toast3.  It also includes various other things:

- Fixes to the `format_source.sh` script that prevented formatting of some files
- Fix to parsing the ground schedule header information
- Adding atmosphere sims to the ground example workflows and benchmark
- Extend the concept of the "`info_rank0`" logger method to support all log levels and an optional timer


tskisner:
The MPI unit tests seemed to time out, but they run normally locally on 2 processes.  Will re-trigger them and see if they die in the same way.

src/toast/schedule.py
```diff
@@ -371,27 +371,31 @@ def _parse_line(line):
                 isplit, nsplit = file_split
             scan_counters = dict()
 
+            read_header = False
```
keskitalo, line 374: These changes alter the logic. Earlier, comment lines were optional but now the header only gets read _if_ there is at least one comment line before the header.

src/toast/utils.py
```diff
@@ -45,16 +45,66 @@
...
+            elif level == "DEBUG":
+                log.debug(msg)
+            elif level == "INFO":
+                log.info(msg)
+            elif level == "WARNING":
+                log.warning(msg)
+            elif level == "ERROR":
+                log.error(msg)
+            elif level == "CRITICAL":
+                log.critical(msg)
```
keskitalo, line 91: Should there be one final `else` to catch erroneous log levels?

src/toast/weather.py
```diff
@@ -227,10 +232,10 @@ def __init__(self, time=None, name=None, file=None, site_uid=0, realization=0):
         self._sim_pwv = self._draw("TQV")
         self._sim_humidity = self._draw("QV10M")
         self._sim_surface_pressure = self._draw("PS")
-        self._sim_surface_temperature = self._draw("TS")
-        self._sim_air_temperature = self._draw("T10M")
-        self._sim_west_wind = self._draw("U10M")
-        self._sim_south_wind = self._draw("V10M")
+        self._sim_surface_temperature = self._draw("TS") * u.Kelvin
```
keskitalo, line 235: Great!

src/toast/ops/sim_tod_atm.py
```diff
@@ -0,0 +1,796 @@
...
+                ob[atm_sim_key].append(sim_list)
+
+                # if self.debug_tod:
+                #     self._save_tod(ob, times, istart, nind, ind, comm)
+                tmin = tmax
+
+            # Create the wind intervals
+            ob.intervals.create_col(wind_intervals, wind_times, ob.shared[self.times])
+
+            # Observation pipeline.  We do not want to store persistant detector
```
keskitalo, line 401: *persistent

keskitalo (changes requested):

src/libtoast/src/toast_atm.cpp
```diff
@@ -220,17 +219,17 @@ void toast::atm_sim_compress_flag_hits_rank(
     double sinel0
     ) {
     double t_fake = -1.0;
-    for (int64_t ix = 0; ix < nx - 1; ++ix) {
+    for (int64_t ix = 0; ix < nx; ++ix) {
```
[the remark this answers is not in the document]
tskisner: I tried to figure out the original logic, but I believe even if that final loop iteration is not needed, it "does no harm" and should be relatively cheap.

src/toast/schedule.py
tskisner: Good point, will fix.  At some point I would like to switch formats to HDF5 like the satellite schedule, but that would break use of existing schedule files.  We can figure out the least disruptive time for that.

keskitalo (approved):

tskisner:
I have reproduced the MPI test failure locally inside the docker container used in travis tests.  Working to sort out what is going on.

tskisner:
The test failures only occur under the following conditions:
- Running inside the docker container used by github actions
- Running with group size == 1 process
- Running with at least these unit tests enabled:
 ``` suite.addTest(loader.loadTestsFromModule(test_ops_mapmaker))

        # commenting out passes
        suite.addTest(loader.loadTestsFromModule(test_ops_mapmaker_solve))

        # commenting out passes
        suite.addTest(loader.loadTestsFromModule(test_ops_mapmaker_binning))

        # commenting out passes
        suite.addTest(loader.loadTestsFromModule(test_ops_mapmaker_utils))

        # commenting out passes
        suite.addTest(loader.loadTestsFromModule(test_ops_sim_tod_atm))

        # commenting out passes
        suite.addTest(loader.loadTestsFromModule(test_ops_sim_ground))

        # commenting out passes
        suite.addTest(loader.loadTestsFromModule(test_ops_sim_satellite))
```
and other tests commented out.  If I comment out *any one* of those 6 unit tests, then they all pass.  With these 6 tests enabled, one of the processes hits an error when allocating MPI shared memory.  I am attempting to disable use of `MPI.COMM_SELF` to see if the problem is somehow related to that.  

If this does not work I suggest that we merge regardless in order to not hold up work that needs the ported atmosphere sims.  Separately I'll expand the MPI shared memory unit tests to see if I can catch the corner case we are hitting.

tskisner:
Ok, the problem was indeed in the upstream `pshmem` package.  I tagged a new release of that and it is now on pypi and conda-forge.  Now working on an updated `cmbenv` docker container with that fix.  Then I can remove the workaround from the last commit on this branch and it will use that new dependency container for the tests.

PR #413, opened 2021-07-19
Port polyfilter

keskitalo:
Port the 1 and 2D polyfilters. Upgrade the 2D filter to run on subsets of the focalplane.

### DIFF
--- a/src/toast/ops/polyfilter.py
+++ b/src/toast/ops/polyfilter.py
@@ -40,7 +40,7 @@ class PolyFilter2D(Operator):
         f".*",
         allow_none=True,
         help="Regex pattern to match against detector names. Only detectors that "
-        "match the pattern are scrambled.",
+        "match the pattern are filtered.",
     )
 
     order = Int(1, allow_none=False, help="Polynomial order")
@@ -295,7 +295,7 @@ class PolyFilter(Operator):
         f".*",
         allow_none=True,
         help="Regex pattern to match against detector names. Only detectors that "
-        "match the pattern are scrambled.",
+        "match the pattern are filtered.",
     )
 
     order = Int(1, allow_none=False, help="Polynomial order")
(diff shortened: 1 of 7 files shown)


src/toast/ops/polyfilter.py
```diff
@@ -0,0 +1,607 @@
...
+            t_clean = 0
+
+            # communicator for processes with the same sample range
+            comm = obs.comm_col
+
+            # Detectors to process
+
+            detectors = []
+            for det in obs.all_detectors:
+                if pat.match(det) is None:
```
tskisner, line 115: The `Pipeline` operator can be used to execute multiple sub-operators one "detector set" at a time (not just one detector at a time).  However, we have explicitly banned that mode of operation above (we don't allow calling `exec()` with a subset of detectors).  So if we want to filter det sets (as defined in each observation), we would need to use some other method of selecting each detector set (either a focalplane key or a regex that matches their names).  I guess this is fine for now, but at some point we might want to add extra logic to check that if a subset of detectors is given, *AND* if it is a detector set, then allow it.

src/toast/ops/polyfilter.py
```diff
@@ -0,0 +1,607 @@
...
+
+    def _finalize(self, data, **kwargs):
+        return
+
+    def _requires(self):
+        req = {
+            "meta": list(),
+            "shared": [self.shared_flags],
+            "detdata": [self.det_data, self.det_flags],
+            "intervals": [self.view],
```
tskisner, line 329: We should only append the flag and interval names if they are not `None`.

src/toast/ops/polyfilter.py
```diff
@@ -0,0 +1,607 @@
...
+        }
```
tskisner, line 456: Same note here about only appending flags / view if they are not None.

tskisner (approved):
Just some minor comments, but thanks for porting all this!

src/toast/ops/polyfilter.py
```diff
@@ -0,0 +1,607 @@
...
+    @function_timer
+    def _exec(self, data, detectors=None, **kwargs):
+        """Apply the 2D polynomial filter to the signal.
+
+        Args:
+            data (toast.Data): The distributed data.
+
+        """
+        if detectors is not None:
+            raise RuntimeError("PolyFilter2D cannot be run in batch mode")
```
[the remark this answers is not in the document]
keskitalo: Sure.

src/toast/ops/polyfilter.py
keskitalo: Done.

src/toast/ops/polyfilter.py
keskitalo: Done.

keskitalo:
Addressed the feedback. Tests pass. Merging.

PR #415, opened 2021-07-23
speed up format_source.sh

ickc:
@keskitalo, @tskisner, this
improve how format_source.sh is run. It should be functionally identical to before, except removing libtoast_mpi which doesn't exist.
The major speed up comes from changing \; to + when using find, such that spawning processes are avoided, and running the 2 find asynchronously.
Also, '{}' is unnecessary to find, it handles white spaces correctly when passing to exec.

Edit: this speeds up from 23.76s to 0.44s on my local computer.

### DIFF
--- a/src/format_source.sh
+++ b/src/format_source.sh
@@ -38,6 +38,6 @@ blktest="--check"
 
 cd "$base"
 # Directories to process
-find libtoast toast -name "*.hpp" -not -path '*Random123*' -not -path '*pybind11/*' -not -path '*gtest/*' -exec ${unexe} ${unrun} {} + &
+find libtoast toast \( -name "*.hpp" -or -name "*.cpp" \) -and -not \( -path '*Random123*' -or -path '*pybind11/*' -or -path '*gtest/*' \) -exec ${unexe} ${unrun} {} + &
 find toast ../pipelines -name "*.py" -not -path '*pybind11/*' -exec ${blkexe} ${blkrun} {} + &
 wait
--- a/src/format_source.sh
+++ b/src/format_source.sh
@@ -6,9 +6,7 @@
 #
 
 # Get the directory containing this script
-pushd $(dirname $0) > /dev/null
-base=$(pwd -P)
-popd > /dev/null
+base="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 # Check executables
 unexe=$(which uncrustify)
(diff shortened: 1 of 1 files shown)


src/format_source.sh
```diff
@@ -6,9 +6,7 @@
 #
 
 # Get the directory containing this script
-pushd $(dirname $0) > /dev/null
-base=$(pwd -P)
-popd > /dev/null
+base="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
tskisner, line 9: This is just a stylistic choice.  I think the separate use of `pushd` and `popd` are more clear to read.  However the `format_source.sh` script has already changed in toast3, so I'll just keep it as the original in that branch.

src/format_source.sh
```diff
@@ -38,16 +36,8 @@ blkrun="-l 88"
...
-for cppd in ${cppdirs}; do
-    find "${base}/${cppd}" -name "*.hpp" -not -path '*Random123*' -not -path '*pybind11/*' -not -path '*gtest/*' -exec ${unexe} ${unrun} '{}' \;
-    find "${base}/${cppd}" -name "*.cpp" -not -path '*Random123*' -not -path '*pybind11/*' -not -path '*gtest/*' -exec ${unexe} ${unrun} '{}' \;
-done
-
-for pyd in ${pydirs}; do
-    find "${base}/${pyd}" -name "*.py" -not -path '*pybind11/*' -exec ${blkexe} ${blkrun} '{}' \;
-done
+find libtoast toast \( -name "*.hpp" -or -name "*.cpp" \) -and -not \( -path '*Random123*' -or -path '*pybind11/*' -or -path '*gtest/*' \) -exec ${unexe} ${unrun} {} + &
+find toast ../pipelines -name "*.py" -not -path '*pybind11/*' -exec ${blkexe} ${blkrun} {} + &
```
tskisner, line 42: Thanks, I was not aware of this syntax until now.  Note that the `+` argument to `find ... -exec` simply passes the result of the find (the long list of files) to the command.  So it only works with commands that accept multiple files as an argument (which both uncrustify and black do).  I've added a comment to the script in the toast3 branch about this.

tskisner (approved):
Thanks for speeding this up.  I'll push the equivalent changes to toast3 in a moment.

src/format_source.sh
ickc: Admittedly this is something I picked up from https://stackoverflow.com/a/246128 in the past, which has evolved a bit.

I didn't think much when I changed that line, but thinking on why I changed it, probably it is because those statements requires more mental parsing as it is more stateful. Since those sub-shell is necessary anyway, getting an output from a subshell directly is less stateful (it is not changing anything in the current shell.)

Agree in the end it is just a stylistic choice as it works none the less!

src/format_source.sh
ickc: I checked them before patching it, but in practice, of all command line program accepting path as positional argument, I don't recall any of them not accepting multiple paths, except in those cases that requires exactly 2 paths which has different meaning (input path, output path, such as `git mv`, etc.) Also, program like this often support "recursive", so providing directories instead is enough if not because of those filters used in find (they may have cli args that can do similar filtering but probably using find as a single syntax used to provide filter to 2 completely unrelated program is smarter.)

But one thing I did hesitate in the patch is POSIX compliance! `+` (and the other useful `-print0`) is not required by POSIX, and is a more "modern" addition from GNU find.

If I remember correctly, BSD find didn't support them years ago. But when I create this patch, BSD find on macOS 10.15+ is already supporting it. As the performance implication is big, I think probably it is ok. But I should have mentioned this in the original message first. (The "need for speed" comes from wrestling with the formatter to try to manually edit it in some way that the formatter operation is an identity operation... E.g. black try to split `[:]` in 3 lines and I didn't managed to manually write something else that it would preserve.)

(Writing bash script that's compatible with any system out there is hard.)

PR #420, opened 2021-08-02
Pole raster scans

keskitalo:
This branch should not be merged to `toast3` before #419


keskitalo:
@tskisner This PR is now ready for review.

tskisner (approved):
The code looks fine, but I have not tried to run any raster scans.  And there are no specific unit tests targeting the ground schedule simulations, so no problems will be caught by our C.I. workflow.  At some point in the future we could think about how to test the scheduling code in more detail (it is not obvious to me the best way to do that).

PR #421, opened 2021-08-04
Porting Crosstalk to toast3   

giuspugl:
With this PR we aim at implementing the injection of crosstalk.
With respect to https://github.com/hpc4cmb/toast/pull/380  there are substantial changes: 
1. the CrossTalk matrix here is a dictionary of dictionaries and it is loaded by all the processing elements  
2.  the most efficient way to Crosstalk the signal between detectors is to  redistribute the data by time, (usually the data are distributed by detector groups ). We therefore operate a redistribution before injecting the cross-talk and after having updated the detector data redistribute back to the previous distribution.  
3. An implementation without data redistribution is also present but it is not as debugged as the other one.  
4. Implemented also a CrossTalk mitigator, assuming that it could be possible  to estimate the crosstalk  coefficients  either by scanning a planet or by means of a stimulator.  
5. It is also possible to read a Cross-talk matrix from file (so far assumed `*.npz` format).https://github.com/hpc4cmb/toast/blob/a2a49c3ff9b0f507631a723cde70effd923b0e63/src/toast/ops/sim_crosstalk.py#L22

- [ ] Documentation is missing in most of the functions 
- [ ] Several functions are called within the two operators  not sure whether to put them  though: https://github.com/hpc4cmb/toast/blob/a2a49c3ff9b0f507631a723cde70effd923b0e63/src/toast/ops/sim_crosstalk.py#L22 
https://github.com/hpc4cmb/toast/blob/a2a49c3ff9b0f507631a723cde70effd923b0e63/src/toast/ops/sim_crosstalk.py#L47
https://github.com/hpc4cmb/toast/blob/a2a49c3ff9b0f507631a723cde70effd923b0e63/src/toast/ops/sim_crosstalk.py#L77
https://github.com/hpc4cmb/toast/blob/a2a49c3ff9b0f507631a723cde70effd923b0e63/src/toast/ops/sim_crosstalk.py#L104


src/toast/ops/sim_crosstalk.py
```diff
@@ -0,0 +1,480 @@
...
+@function_timer
+def read_xtalk_matrix(filename, data):
+    """
+    Read Xtalk matrix from disc.
+    the  file to be in numpy binary format,
+    i.e. `*.npz`.
+
+    """
+
+    matrix = np.load(filename)["matrix"]
```
tskisner, line 30: In the future, we might consider using a different format where we have access to which row / column corresponds to a given detector, since we might have the same crosstalk matrix but different detector order in the observations (if some detectors are not active).

tskisner (approved):
Thanks for this work!  There are a couple places that will get simplified by upcoming changes that are in progress in other branches, but I think it is good to merge this now and then refine it when those other features are available.

tskisner:
@keskitalo approved this in offline discussion.

PR #422, opened 2021-08-04
Stats

keskitalo:
Add an operator that measures the data mean, variance, skewness and kurtosis and saves the result in per-observation HDF5 files.

### DIFF
--- a/workflows/toast_sim_ground.py
+++ b/workflows/toast_sim_ground.py
@@ -239,10 +239,10 @@ def reduce_data(job, args, data):
     timer = toast.timing.Timer()
     timer.start()
 
-    # Collect signal statistics
+    # Collect signal statistics before filterting
 
-    ops.statistics.output_dir = args.out_dir
-    ops.statistics.apply(data)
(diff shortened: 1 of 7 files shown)


workflows/toast_sim_ground.py
```diff
@@ -250,6 +255,11 @@ def reduce_data(job, args, data):
     ops.common_mode_filter.apply(data)
     log.info_rank("Finished common-mode-filtering in", comm=world_comm, timer=timer)
 
+    # Collect signal statistics after filterting
+
+    ops.filtered_statistics.output_dir = args.out_dir
```
tskisner, line 260: This is the same output directory used by the previous dump- will it overwrite those files?  Maybe dump to sub-directories or support a "prefix" trait in the `Statistics` operator to make the file names unique?

tskisner (approved):
This is a fine starting point for our data QA, thanks!

workflows/toast_sim_ground.py
keskitalo: I'm using the unique operator name as the prefix. No overwriting should occur.

PR #423, opened 2021-08-04
Support writing out inverse noise matrices and the noise-weighted map…

keskitalo:
…; Allow user to choose which data products to write


tskisner (approved):
Thanks, I agree that it is good to have fine-grained control over which mapmaker products are written.

PR #424, opened 2021-08-05
Fix polarization angle calculation

keskitalo:
I recently had to work out the proper math for translating detectors quaternions to polarization angles in s4sim. This is the same calculation.

### DIFF
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -372,7 +372,7 @@ class Focalplane(object):
                 quat = row["quat"]
                 a = quat[3]
                 d = quat[2]
-                pol_angle2 = np.arctan2(2 * a * d, a ** 2 - d ** 2) % np.pi
+                pol_angle = np.arctan2(2 * a * d, a ** 2 - d ** 2) % np.pi
                 row["pol_angle"] = pol_angle * u.radian
 
     def _get_pol_efficiency(self):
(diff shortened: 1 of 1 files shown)


mhasself:
Hi @keskitalo  -- sorry for lurking ... I've spent a lot of time on quaternions over the past few years so forgive me for jumping on this.

I think your edits are equivalent to:
```
a, d = quat[..., 3], quat[..., 2]
pol_angle = np.arctan2(2*a*d, a**2 - d**2) % np.pi
```

I think this form is also more numerically stable.  When you use qarray.to_angles, you're decomposing the rotation into ZYZ euler angles, and relying on the returned longitude being 0 (which is a convention one might adopt when a rotation appears to be a simple rotation about z^).  If the quaternion is even slightly different from a pure Z rotation, then the longitude could get a basically random value, and that would affect the value of position angle (~equal and opposite).  That being said ... you can get the stability back by instead computing (and I might be off by a sign here):
```
lat, lon, pos = qarray.to_angles(angrot)
pol_angle = (lon + pos) % np.pi
```

keskitalo:
Thanks @mhasself ! I had a vague recollection that there was a shortcut for extracting the polarization angle but could not find it fast enough. Your solution is obviously more robust and also more elegant.

PR #425, opened 2021-08-05
Per detector elevation noise model

keskitalo:
If operator is not configured with an elevation noise model, look for one in the focaplane database.

@tskisner what do you think of the field names "A" and "C" in the focaplane database? Using more descriptive names like "elevation_noise_a" and "elevation_noise_c" would make the ASCII view of the database cluttered but would make things easier to understand.

### DIFF
--- a/src/toast/ops/elevation_noise.py
+++ b/src/toast/ops/elevation_noise.py
@@ -182,9 +182,9 @@ class ElevationNoise(Operator):
                 if self.noise_a is not None:
                     noise_a = self.noise_a
                     noise_c = self.noise_c
-                elif "A" in focalplane[det]:
-                    noise_a = focalplane[det]["A"]
-                    noise_c = focalplane[det]["C"]
+                elif "elevation_noise_a" in focalplane[det]:
+                    noise_a = focalplane[det]["elevation_noise_a"]
+                    noise_c = focalplane[det]["elevation_noise_c"]
                 else:
                     continue
 
--- a/src/toast/ops/elevation_noise.py
+++ b/src/toast/ops/elevation_noise.py
@@ -71,13 +71,13 @@ class ElevationNoise(Operator):
     noise_a = Float(
         None,
         allow_none=True,
-        help="Parameter 'a' in (a / sin(el) + c)",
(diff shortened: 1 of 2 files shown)


src/toast/ops/elevation_noise.py
```diff
@@ -71,13 +71,13 @@ class ElevationNoise(Operator):
     noise_a = Float(
         None,
         allow_none=True,
-        help="Parameter 'a' in (a / sin(el) + c)",
+        help="Parameter 'a' in (a / sin(el) + c).  If not set, look for one in the Focalplane.",
```
tskisner, line 74: This change is fine, but it does require that the `detector_data` table in the focalplane has names that conform to this.  If not, then the calling code will have to use the current technique of passing in a common value of A / C for all detectors.  Could you also add the A / C column names to the optional columns in the Focalplane docstring here:

https://github.com/hpc4cmb/toast/blob/790f077cdb1d4278902450150e94858c70ad1baf/src/toast/instrument.py#L272

src/toast/ops/elevation_noise.py
```diff
@@ -181,6 +178,16 @@ def _exec(self, data, detectors=None, **kwargs):
                 view_flags.append(flags)
 
             for det in dets:
+                # If both the A and C values are unset, the noise model is not modified.
+                if self.noise_a is not None:
+                    noise_a = self.noise_a
+                    noise_c = self.noise_c
+                elif "A" in focalplane[det]:
+                    noise_a = focalplane[det]["A"]
```
tskisner, line 186: I agree with your comment- let's change these column names to something meaningful.

tskisner (changes requested):
Looks good!  Just a minor change to the Focalplane docstring and using better column names for A / C.

src/toast/ops/elevation_noise.py
keskitalo: Good point! Augmented the `Focalplane` docstring.

src/toast/ops/elevation_noise.py
keskitalo: Replaced "A" -> "elevation_noise_a" and "C" -> "elevation_noise_c".

keskitalo:
Merging with the requested changes.

PR #426, opened 2021-08-06
General SPT3G interface tools

tskisner:
This is still a work in progress, but will eventually support general translation to / from spt3g frames on one or multiple processes.


tskisner:
This is ready for review, but I think we should merge the other open PRs first and then rebase this and resolve any conflicts.  The main high-level operators added by this PR are the `LoadSpt3g`, `SaveSpt3g`, and `RunSpt3g` classes.  These take instances of helper classes that do the actual conversion between observations and frames.  There are default helper classes (`toast.spt3g.export_obs` and `toast.spt3g.import_obs`) that do this conversion in a general way and take additional helper classes that convert just the observation data and metadata.  Although there are default versions of those lowest-level helper classes (`export_obs_meta`, `export_obs_data`, `import_obs_meta`, `import_obs_data`), these low-level classes are the things that need to be overridden for specific experiments, since they are tied to the frame schema.  The default implementation just ensures that all common toast information can be packed into frames and converted back.

src/toast/tests/spt3g.py
```diff
@@ -0,0 +1,406 @@
...
+            lines = [f"Proc {data.comm.world_rank}: {x}" for x in lines]
+            print("".join(lines), flush=True)
+
+        # Verify
+        for ob, orig in zip(check_data.obs, original):
+            if ob != orig:
+                print(f"-------- Proc {data.comm.world_rank} ---------\n{orig}\n{ob}")
+            self.assertTrue(ob == orig)
+
+    def test_save_load_nocomp(self):
```
keskitalo, line 260: Even after reading the test source twice, I cannot figure out what "nocomp" means. "No computation" or "no comparison"? Could we add a comment?

src/toast/tests/spt3g.py
```diff
@@ -0,0 +1,406 @@
...
+        # Import the data
+        check_data = Data(comm=data.comm)
+
+        try:
+            for obframes in g3data:
+                check_data.obs.append(importer(obframes))
+
+            for ob in check_data.obs:
+                ob.redistribute(ob.comm_size, times="times")
+        except Exception as e:
```
keskitalo, line 245: Should the try-except remain here after debugging?

src/toast/spt3g/spt3g_export.py
```diff
@@ -0,0 +1,534 @@
...
+    obs, name, view_name=None, view_index=0, g3t=None, times=None, compress=None
+):
+    """Convert a single detdata object to a G3Object.
+
+    If the G3 data type is not specified, a guess will be made at the closest
+    appropriate type.
+
+    Args:
+        obs (Observation):  The parent observation.
+        name (str):  The name of the shared object.
```
keskitalo, line 81: Not "shared" but "detdata" object.

src/toast/spt3g/spt3g_export.py
```diff
@@ -0,0 +1,534 @@
...
+        view_name (str):  If specified, use this view of the shared object.
```
keskitalo, line 82: See above.

src/toast/spt3g/spt3g_import.py
```diff
@@ -0,0 +1,707 @@
...
+        name (str):  The name of the shared object.
+        offset (int):  The local data starting offset.
+        g3obj (G3Object):  The G3Object to copy.
+
+    Returns:
+        None
+
+    """
+    if name not in obs.shared:
+        raise KeyError(f"Shared object '{name}' does not exist in observation")
```
keskitalo, line 73: Why can't we just create the field?

src/toast/spt3g/spt3g_import.py
```diff
@@ -0,0 +1,707 @@
...
+        name (str):  The name of the detdata object.
+        offset (int):  The local data starting offset.
+        g3obj (G3Object):  The G3Object to copy.
+
+    Returns:
+        None.
+
+    """
+    log = Logger.get()
+    if name not in obs.detdata:
```
keskitalo, line 106: Same as above. Why can't we just create the field if it is missing?

keskitalo (approved):
Looks great. I'm looking forward to seeing some benchmarks on performance. Please look through my comments and merge away when you feel like it.

src/toast/ops/load_spt3g.py
```diff
@@ -0,0 +1,154 @@
...
+        im_rank = self.obs_import.import_rank
+        if im_rank is None:
+            raise RuntimeError(
+                "The obs_import class must be configured to import frames on one process"
+            )
+
+        # FIXME:  we should add a check that self.obs_import.comm is the same as
+        # the data.comm.comm_group communicator.
+
+        # One global process computes the list of observations and their approximate
```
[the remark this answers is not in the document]
tskisner: Sorry for the confusion, the root process is not opening any files.  It is just querying their sizes from the file system and using the sum of the g3 file sizes as a proxy for the relative observation size.  Then the observations are are distributed to the groups based on these sizes.  Then one process in each group actually loads the data.

And of course experiments might have their own version of this operator which queries a database for the observation sizes (for example).

src/toast/ops/run_spt3g.py
```diff
@@ -0,0 +1,141 @@
...
+from ..spt3g import available
+
+if available:
+    from spt3g import core as c3g
+    from ..spt3g import frame_emitter
+    from ..spt3g import frame_collector
+
+
+@trait_docs
+class RunSpt3g(Operator):
```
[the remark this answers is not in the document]
tskisner: When converting the observation data to frames, the export function already accepts a set of intervals (i.e. a view) to use for the frame boundaries:  https://github.com/hpc4cmb/toast/blob/24f6b743d19c38df727e47fec037bd3770716208/src/toast/spt3g/spt3g_export.py#L339
If that is specified, then only those intervals will be emitted as frames to be processed by the G3 pipeline run by this operator.  The default frame boundaries are set by the sample set boundaries in the observation distribution.

However, I just realized that using a set of intervals to emit frames with a subset of data would have an interesting consequence when re-converting these frames back into an observation.  In that situation, the output observation would only contain data from frames that passed through the G3 pipeline.  So all of the data not in the original intervals would be cut out.  

Not sure if there is a way around this- the G3 model for cutting data is to just drop those frames.  If we wanted to support this, we would have to pass a pre-existing observation to the `import_obs` class and have that class take each frame and try to locate it with the observation.  Then it would have to copy all the data in the frame into the corresponding observation shared / detdata fields at the correct sample offset.  And it would have to create any new data objects that were placed in the frames by the G3 pipeline.  It is a lot of extra logic...

For the short term, we could simply not use that option if we intend to return data back to toast observations.

src/toast/tests/spt3g.py
tskisner: "no compression".  Will change it.

src/toast/spt3g/spt3g_import.py
tskisner: This is a helper function that is copying a single frame's worth of one G3 object (i.e. a timestream, vector, etc) into a larger (pre-created) shared data object.  When importing, the schema of the first frame is used to determine all the shared / detdata objects that need to be created.  After these are pre-created in the observation, we go through each frame and each object and copy that small slice of data into the larger buffer.

src/toast/tests/spt3g.py
tskisner: Removed.

src/toast/spt3g/spt3g_export.py
tskisner: Fixed, thanks.

src/toast/scripts/toast_benchmark_ground
```diff
@@ -251,6 +265,70 @@ def make_schedule(args, world_comm, site):
...
+            (job_ops.elevation_model.out_model, job_ops.elevation_model.out_model),
+        ]
+    )
+    # Note that we export detector flags below to a float64 G3TimestreamMap
+    # in order to use FLAC compression.
+    data_exporter = t3g.export_obs_data(
+        shared_names=[
+            (
+                job_ops.sim_ground.boresight_azel,
+                job_ops.sim_ground.boresight_azel,
```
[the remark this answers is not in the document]
tskisner: This is pretty well documented in the `export_obs_data` class documentation:
https://github.com/hpc4cmb/toast/blob/2b48421ed61f3d1b1334548d28e66201f6b4b72e/src/toast/spt3g/spt3g_export.py#L314-L321
Those tuples define the mapping of names from observation object names to G3 frame object names, and optionally the data type to use on the G3 side.  Since we are just using an arbitrary schema, I am just using the same string for the frame key as the observation name.

I added a comment telling the user to read the class documentation for a description of the constructor arguments.

PR #427, opened 2021-08-06
Add support for atmospheric polarization

keskitalo:
This PR adds `polarization_fraction` as a trait to the atmospheric simulation. It is a constant fraction that generates Q-polarization from the pure intensity simulation.

### DIFF
--- a/workflows/toast_sim_ground.py
+++ b/workflows/toast_sim_ground.py
@@ -177,6 +177,9 @@ def simulate_data(job, toast_comm, telescope, schedule):
     ops.det_pointing_azel.boresight = ops.sim_ground.boresight_azel
     ops.det_pointing_radec.boresight = ops.sim_ground.boresight_radec
 
+    ops.det_weights_azel.detector_pointing = ops.det_pointing_azel
+    ops.det_weights_azel.hwp_angle = ops.sim_ground.hwp_angle
+
     # Create the Elevation modulated noise model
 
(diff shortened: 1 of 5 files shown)


workflows/toast_sim_ground.py
```diff
@@ -309,6 +314,9 @@ def main():
             out_model="el_noise_model",
         ),
         toast.ops.PointingDetectorSimple(name="det_pointing_azel", quats="quats_azel"),
+        # In the future, `det_weights_azel` may be a dedicated operator that does not
+        # expand pixel numbers but just Stokes weights
+        toast.ops.PointingHealpix(name="det_weights_azel", mode="IQU"),
```
tskisner, line 319: Note that unless the user specifies `--det_weights_azel.weights some_other_string`, the default name of the detector weights computed by this operator (`weights`) is the same as that used by the `pointing` operator below.  Perhaps we could set `weights="weights_azel"` or similar in the constructor?

tskisner (approved):
Just one small comment.  Also, should we add this feature to the ground benchmark?

workflows/toast_sim_ground.py
keskitalo: Done.

keskitalo:
I would hold off on adding polarized atmosphere to the benchmarks. It is likely to evolve quite a bit.

PR #428, opened 2021-08-07
Fix unit test error affecting certain Python versions

keskitalo:


keskitalo:
Self-merging this PR to fix the broken unit tests.

PR #429, opened 2021-08-09
Common mode filter

keskitalo:
Add a simple common mode filter.


tskisner (approved):
Bonus points for backporting :-)

PR #430, opened 2021-08-09
Update ground benchmark and some related changes.

tskisner:
- Add the filter stack from the ground example into the benchmark.

- Install the benchmarking_utils into the toast package rather than
  the bin directory.

- Move the boilerplate for checking unhandled exceptions into a
  context manager and use that in all workflows and scripts.

@keskitalo , do we want to increase the default ground benchmark healpix resolution from 1024 to 4096?  What about the patch size used for the benchmark (currently `schedule_patch = "tile000,1,40,0,60,-40"`)?


src/toast/mpi.py
```diff
@@ -268,22 +269,28 @@ def exception_guard(comm=None):
...
     log = Logger.get()
     failed = 0
+    rank = 0
+    if comm is not None:
+        rank = comm.rank
     try:
         yield
-    except:
-        msg = "Exception on process {}:\n".format(comm.rank)
+    except Exception:
```
keskitalo, line 277: Bare `except` clauses should be avoided if possible. This is one of the few cases where it is warranted but perhaps it would pay to add a comment?

workflows/toast_sim_ground.py
```diff
@@ -239,7 +239,7 @@ def reduce_data(job, args, data):
     timer = toast.timing.Timer()
     timer.start()
 
-    # Collect signal statistics before filterting
+    # Collect signal statistics before filtering
```
keskitalo, line 242: Thanks!

keskitalo (approved):

tskisner:
@keskitalo and @nestordemeure , some more changes I just added.  Basically the mapmaker now deletes by default the intermediate solver map domain products and also deletes each final map domain product after writing it out.  The current healpix writing introduces additional memory overhead, so now the code writes smaller products first and deletes in order to free up memory for writing the larger products.  The ground benchmark now uses a wide scanning strategy similar to a future Atacama LAT.  I think this is ready to go once tests pass.

PR #431, opened 2021-08-11
Atmospheric loading and bandpass

keskitalo:
- Scale the atmospheric fluctuations according to bandpass
- Add atmospheric background loading, also according to bandpass

Both are applied in `ObserveAtmosphere` but we could also separate the background loading into a separate operator.

This PR adds to named members to the `Focalplane` class:
- `noise`
- `bandpass`
If the detector data allows, they are created upon initialization. This renders the `DefaultNoiseModel` almost trivial: it only links the `focalplane.noise` to the appropriate observation key. It may still make sense to retain this operator so that noise simulation operators always look for an observation key instead of a focalplane member.

### DIFF
--- a/src/toast/ops/noise_model.py
+++ b/src/toast/ops/noise_model.py
@@ -46,36 +46,9 @@ class DefaultNoiseModel(Operator):
         comm = data.comm
 
         for ob in data.obs:
-            # Get the detectors we are using for this observation
-            dets = ob.select_local_detectors(detectors)
-            if len(dets) == 0:
-                # Nothing to do for this observation
-                continue
-
-            # The focalplane for this observation
-            focalplane = ob.telescope.focalplane
-
-            # Every process has a copy of the focalplane, and every process may want
-            # the noise model for all detectors (not just our local detectors).
-            # So we simply have every process generate the same noise model locally.
-
-            fmin = {}
-            fknee = {}
-            alpha = {}
-            NET = {}
-            rates = {}
-            for d in dets:
-                rates[d] = focalplane.sample_rate
-                fmin[d] = focalplane[d]["psd_fmin"]
-                fknee[d] = focalplane[d]["psd_fknee"]
-                alpha[d] = focalplane[d]["psd_alpha"]
-                NET[d] = focalplane[d]["psd_net"]
-
-            noise = AnalyticNoise(
-                rate=rates, fmin=fmin, detectors=dets, fknee=fknee, alpha=alpha, NET=NET
-            )
-
-            ob[self.noise_model] = noise
+            if ob.telescope.focalplane.noise is None:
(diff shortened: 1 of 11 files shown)


keskitalo:
@tskisner the functional parts are all there but I will expand the unit tests. Please see the notes at the top of this thread.

keskitalo:
@tskisner This PR is now ready for review, complete with unit tests.

src/toast/ops/noise_model.py
```diff
@@ -46,36 +46,9 @@ def _exec(self, data, detectors=None, **kwargs):
...
-                NET[d] = focalplane[d]["psd_net"]
-
-            noise = AnalyticNoise(
-                rate=rates, fmin=fmin, detectors=dets, fknee=fknee, alpha=alpha, NET=NET
-            )
-
-            ob[self.noise_model] = noise
+            if ob.telescope.focalplane.noise is None:
+                raise RuntimeError("Focalplane does not have a noise model")
+            ob[self.noise_model] = ob.telescope.focalplane.noise
```
tskisner, line 51: Although this operator is now even more trivial, I like having this to store the noise model in the observation instance.  This allows us to have multiple noise models (e.g. the nominal one and an estimated one) and select which one to use at the workflow level.

src/toast/ops/sim_ground.py
```diff
@@ -371,7 +377,7 @@ def _exec(self, data, detectors=None, **kwargs):
                     scan_min_el,
                     scan_max_el,
                 ) = simulate_elnod(
-                    scan.start,
+                    scan.start.timestamp(),
```
tskisner, line 380: Good catch- I fixed that in some other places, but missed this one.

src/toast/ops/sim_tod_atm.py
```diff
@@ -539,14 +504,18 @@ def _get_scan_range(self, obs, comm, prefix):
...
         # to compute the range of angles needed for simulating the slab.
 
-        # FIXME:  This is quite arbitrary and assumes that the boresight
-        # pointing has been created by the SimGround operator.  We should
-        # define the observation keys to use in some standard place and
-        # compute these limits from the data if they do not exist.
-        min_az_bore = obs["scan_min_az"].to_value(u.radian)
-        max_az_bore = obs["scan_max_az"].to_value(u.radian)
-        min_el_bore = obs["scan_min_el"].to_value(u.radian)
-        max_el_bore = obs["scan_max_el"].to_value(u.radian)
```
tskisner, line 549: Thanks for fixing this.  Now I can remove a workaround in another downstream package.

tskisner (approved):
Just a couple docstring cleanups.  Thanks for porting this functionality and cleaning up many small things along the way.

PR #432, opened 2021-08-16
Improve 2d filter

keskitalo:
This PR fixes a memory issue in 2D filtering. It also adds robustness towards poorly-conditioned filtering problems and improves the unit tests.

### DIFF
--- a/src/libtoast/src/toast_math_rng.cpp
+++ b/src/libtoast/src/toast_math_rng.cpp
@@ -29,13 +29,13 @@ void toast::rng_dist_uint64(size_t n, uint64_t key1, uint64_t key2,
         #pragma omp simd
         for (size_t i = 0; i < n; ++i) {
             data[i] = rng(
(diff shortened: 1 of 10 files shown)


src/libtoast/src/toast_math_rng.cpp
```diff
@@ -29,13 +29,13 @@ void toast::rng_dist_uint64(size_t n, uint64_t key1, uint64_t key2,
         #pragma omp simd
         for (size_t i = 0; i < n; ++i) {
             data[i] = rng(
-                RNG::ctr_type({{counter1, counter2 + i} }),
+                RNG::ctr_type({{counter1, counter2 + i}}),
```
tskisner, line 32: Lots of uncrustify noise, but you mentioned that `format_source.sh` gave an error on MacOS.  I'll look into that.  For now we can live with some noise.

tskisner (approved):
Looks great, thanks for this cleanup and additional tests!

PR #435, opened 2021-08-17
Make rotating HWP optional

keskitalo:
sim_ground and sim_satellite now handle 'None' hwp_angle trait.  Workflows use hwp_angle trait consistently.

### DIFF
--- a/src/toast/ops/sim_ground.py
+++ b/src/toast/ops/sim_ground.py
@@ -147,7 +147,9 @@ class SimGround(Operator):
         None, allow_none=True, help="Observation detdata key for flags to initialize"
     )
 
-    hwp_angle = Unicode("hwp_angle", help="Observation shared key for HWP angle")
+    hwp_angle = Unicode(
+        None, allow_none=True, help="Observation shared key for HWP angle"
+    )
 
     azimuth = Unicode("azimuth", help="Observation shared key for Azimuth")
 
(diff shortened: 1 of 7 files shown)


src/toast/ops/sim_ground.py
```diff
@@ -237,10 +239,56 @@ def _check_schedule(self, proposal):
                 )
         return sch
 
+    # Cross-check HWP parameters
```
tskisner, line 242: Just a note:  when validating a set of traits that are inter-dependent like this, we can get into a situation where none of the traits can be changed, or where an error can occur if the user does not set traits in a particular order.  For example, if all 3 traits start as `None`, then we can't set `hwp_angle` (since the other traits are None) and we can't set `hwp_rpm`, because `hwp_angle` is None.  The work arounds are:
1.  Don't use the traitlets validate for these kinds of inter-connected traits.  Instead, just check consistency at the beginning of the `_exec()` method when they are about to be used.
2.  Set all the traits inside a `hold_trait_notifications` context: 
 https://traitlets.readthedocs.io/en/stable/using_traitlets.html#basic-example-validating-the-parity-of-a-trait

So for example:
```
sim_ground = ops.SimGround()
with sim_ground.hold_trait_notifications():
    sim_ground.hwp_rpm = 5.0
    sim_ground.hwp_angle = "hwp_angle"
```
Either way is fine, we should just make sure to note this when overhauling the documentation and discussing Operator traits.

tskisner (approved):
Looks good, thanks for this change!

PR #436, opened 2021-08-17
More sensible defaults for the ground workflow

keskitalo:


workflows/toast_sim_ground.py
```diff
@@ -312,16 +312,16 @@ def main():
...
-        toast.ops.Statistics(name="raw_statistics"),
-        toast.ops.Statistics(name="filtered_statistics"),
-        toast.ops.GroundFilter(name="groundfilter"),
+        toast.ops.Statistics(name="raw_statistics", enabled=False),
+        toast.ops.Statistics(name="filtered_statistics", enabled=False),
+        toast.ops.GroundFilter(name="groundfilter", enabled=False),
         toast.ops.PolyFilter(name="polyfilter1D"),
-        toast.ops.PolyFilter2D(name="polyfilter2D"),
-        toast.ops.CommonModeFilter(name="common_mode_filter"),
+        toast.ops.PolyFilter2D(name="polyfilter2D", enabled=False),
```
tskisner, line 323: All of these same operators also exist in `scripts/toast_benchmark_ground`.  Should we disable them there as well, or are the defaults we want to benchmark different than the defaults for everyday use?

tskisner (approved):
Fine to merge as is, but we should consider any change we want in the corresponding benchmark scripts.

workflows/toast_sim_ground.py
```diff
@@ -312,16 +312,16 @@ def main():
         toast.ops.PointingDetectorSimple(
             name="det_pointing_radec", quats="quats_radec"
         ),
-        toast.ops.ScanHealpix(name="scan_map"),
+        toast.ops.ScanHealpix(name="scan_map", enabled=False),
```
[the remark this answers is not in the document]
keskitalo: Yes. Enabling this operator by default does not do anything since it requires an input map to operate.

workflows/toast_sim_ground.py
keskitalo: I would consider the common mode filter and the 2D filters to be in the "expert" interface. Unless the user specifically wants them, they should be disabled. Also, they should not be used at the same time.

For benchmarking purposes, we want the entire pipeline to be lean. That includes all operators that we anticipate needing.

PR #437, opened 2021-08-19
Make cooler scheduling more general

keskitalo:
Fix logging failure and support naming the cooler cycles.


keskitalo:
Self-merging these trivial changes.

PR #440, opened 2021-08-23
Fix 2D polynomial filter

tskisner:
The 2D polynomial filter was using `MPI.Comm.allreduce()` (lower case) which returns the result (it is not in place).  The return value was being ignored.  This converts the code to using the "upper case" version and doing an in-place allreduce.  Other small tweaks to the unit tests include not zero-ing out the sky signal when accumulating the input test polynomial, and plotting the unflagged samples for reference.

This issue is only seen if the group size is larger than one process (otherwise the allreduce is not called), which only occurs in the unit tests if running on 4 or more processes (since the unit tests create 2 groups).

The original plot of the input / output is:
2Dfiltered_tod

After this fix, the output is more reasonable:
2Dfiltered_tod_new

### DIFF
--- a/src/toast/ops/polyfilter.py
+++ b/src/toast/ops/polyfilter.py
@@ -218,7 +218,7 @@ class PolyFilter2D(Operator):
 
                 # templates = np.zeros([ndet, nmode, nsample])
                 templates = np.zeros([ndet, nmode])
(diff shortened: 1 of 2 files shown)


keskitalo (approved):
Well that is embarrassing... Thanks a lot for fixing this, Ted!

PR #441, opened 2021-08-23
Compute atmosphere absorption and loading once per observation

tskisner:
The original code was ported from toast2 where we always processed all detectors at once.  This splits the functionality into the general absorption / loading calculation which is done once per observation, and then the bandpass integration which is done only for the detectors currently being used in the `ObserveAtmosphere` operator.


tskisner:
Just a note that I successfully tested this on cori KNL and the removal of duplicate work obviously helps substantially.

keskitalo (approved):
A very neat solution!

PR #446, opened 2021-08-27
Add operator for flagging planets and solar system objects

keskitalo:


keskitalo:
Presently this operator assumes that the per-detector Az/El quaternions are available in `detdata`. I'll revise the PR so that we apply the detector pointing operator, making sure that is the case.

keskitalo:
This PR is ready for review.

src/toast/ops/flag_sso.py
```diff
@@ -0,0 +1,272 @@
...
+def to_JD(t):
+    # Unix time stamp to Julian date
+    # (days since -4712-01-01 12:00:00 UTC)
+    return t / 86400.0 + 2440587.5
+
+
+def to_MJD(t):
+    # Convert Unix time stamp to modified Julian date
+    # (days since 1858-11-17 00:00:00 UTC)
+    return to_JD(t) - 2400000.5
```
tskisner, line 45: These are now defined in multiple operator source files.  Maybe they should go to some place that can be imported by all the operators using them?

src/toast/ops/flag_sso.py
```diff
@@ -0,0 +1,272 @@
...
+    det_flag_mask = Int(1, help="Bit mask to raise flags with")
+
+    sso_names = List(
+        default_value=["Sun", "Moon"],
+        trait=Unicode,
+        allow_none=True,
+        help="Names of the SSOs, must be recognized by pyEphem",
+    )
+
+    # FIXME: Once TOAST supports Lists of Quantities,
```
tskisner, line 90: Duly noted and added to the 3.0 project board.

src/toast/ops/flag_sso.py
```diff
@@ -0,0 +1,272 @@
...
+            self.ssos.append(getattr(ephem, sso_name)())
+        self.nsso = len(self.ssos)
+
+        for obs in data.obs:
+            dets = obs.select_local_detectors(detectors)
+            if len(dets) == 0:
+                # Nothing to do for this observation
+                continue
+            site = obs.telescope.site
+            observer = ephem.Observer()
```
tskisner, line 150: I know this work is porting existing code, but am looking forward to someday when we can convert to using astropy for coordinate transforms.  Note that a start on that work exists for the transform from Az/El to RA/DEC in `toast.coordinates`, but not yet the inverse transform.  And more testing needs to be done before switching completely from pyephem.  At some point pyephem will simply stop working with the latest python versions since it is no longer maintained.

tskisner (approved):
Thanks for the ongoing porting work!

PR #450, opened 2021-09-01
Improve MemoryCounter

tskisner:
Small changes to the memreport utility function.  Use this in the MemoryCounter operator to display total memory use in addition to just use by timestream data.  Add memory counter to the ground benchmark as a test.

I debated adding this functionality to the operator base class, in order to track memory use before / after every call to `exec()`.  However, we have potentially a deep hierarchy of operators running pipelines of other operators.  And the memreport function is collective and uses a gather.  So I think it is better to explicitly use this at the highest level.

One can disable this in the ground benchmark from the commandline with:
```
--mem_count.disable
```

### DIFF
--- a/src/toast/ops/madam_utils.py
+++ b/src/toast/ops/madam_utils.py
@@ -8,7 +8,7 @@ import numpy as np
 
 import healpy as hp
 
-from ..utils import Logger, Timer, GlobalTimers, dtype_to_aligned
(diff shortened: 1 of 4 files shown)


keskitalo (approved):
Looks good.

nestordemeure (approved):
Looks good, will test and run as soon as it is merged.

PR #452, opened 2021-09-02
Add cadence map operator

keskitalo:


src/toast/ops/cadence_map.py
```diff
@@ -0,0 +1,217 @@
...
+def to_JD(t):
+    # Unix time stamp to Julian date
+    # (days since -4712-01-01 12:00:00 UTC)
+    return t / 86400.0 + 2440587.5
+
+
+def to_MJD(t):
+    # Convert Unix time stamp to modified Julian date
+    # (days since 1858-11-17 00:00:00 UTC)
+    return to_JD(t) - 2400000.5
```
tskisner, line 34: Are these functions useful for multiple operators?  Should they go in `toast.utils` or `toast.coordinates`?  We can leave them here for now, but if we find ourselves defining them in multiple places then we should move them someplace they can be imported by multiple bits of code.

tskisner (approved):
Looks good.  Thanks for using HDF5 :-)

src/toast/ops/cadence_map.py
keskitalo: Excellent point. I moved them to `toast.coordinates` and will use them in the other PRs under review.

PR #454, opened 2021-09-04
Fix an issue with pole scheduling preventing DC1 SPLAT scheduling

keskitalo:


tskisner (approved):
Seems fine.  There are currently no unit tests for the ground and satellite schedulers, so for now, your work that discovered the issue is the canonical test that it functions correctly.

PR #455, opened 2021-09-07
Improve interfaces for working with subsets of data

tskisner:
* Add method to the Data class for constructing shallow views of
  subsets of observations based on observation index, name, uid,
  and key / values.

* Improve the existing split() method to allow splitting on other
  observation properties.

* Add method to Data for getting superset of detectors across
  all observations.

* Remove "data hack" from existing atmosphere simulation operator.

* Clean up Pipeline class to use the various new functions above and remove the data slicing code (which is now handled directly by the Data class).

* Begin stub of developer docs on creation of Operators and use
  of sub-operators.  Still work to do on this.

I'm going to add a few more notes in the draft documentation, including a discussion of the future benefits of using the `Pipeline` class.  However, the code changes are in place.  We can discuss / iterate on these changes, and once happy with them we can modify the code in #446 and #451 to use the new `Data.select()` method.

### DIFF
--- a/src/toast/data.py
+++ b/src/toast/data.py
@@ -300,8 +300,12 @@ class Data(MutableMapping):
                 a subset of observations.
 
         """
+        log = Logger.get()
         if obs_val is not None and obs_key is None:
             raise RuntimeError("If you specify obs_val, you must also specify obs_key")
+
+        group_rank = self.comm.group_rank
+        group_comm = self.comm.comm_group
 
         new_data = Data(comm=self._comm, view=True)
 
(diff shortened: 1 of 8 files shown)


src/toast/data.py
```diff
@@ -141,86 +166,167 @@ def _get_optional(k, dt):
                 wcomm.barrier()
         return
 
-    def select_observations(self, key, require=False):
-        """Given an observation key, return lists of observations with unique values.
+    def split(
```
keskitalo, line 169: Any chance that the `split()` and `select()` methods could return data objects with broken communicators?

keskitalo (approved):
This is a great improvement to the API!

src/toast/data.py
tskisner: As discussed offline, the select method now also has a check for consistent dictionary values across the group.

PR #460, opened 2021-09-16
Use batched FFTs for time constant convolution / deconvolution.

tskisner:
This replaces Numpy real 1D FFTs with our own internal TOAST FFTs, which are batched and threaded for many backend FFT libraries.  In addition to still passing the unit tests, I manually plotted the differences between fourier domain results and final roundtrip convolve / deconvolve using both this code and the original (they are identical to high precision).

As a concrete test, I ran the ground benchmark on 16 KNL processes, using 2054 detectors and 6 observations.  Using identical job parameters, the old code took this long:
```
<snip>
TOAST INFO: Convolved time constant in 565.31 s
TOAST INFO: Calculated raw statistics in 82.94 s
TOAST INFO: Deconvolved time constant in 562.15 s
<snip>
```
And the new code took this long:
```
<snip>
TOAST INFO: Convolved time constant in 41.26 s
TOAST INFO: Calculated raw statistics in 81.60 s
TOAST INFO: Deconvolved time constant in 37.94 s
<snip>
```
So about 14 times faster.  For reference, this job had 16 threads per process, so we are seeing the improvement due to the threading of the FFTs in each batch.  Although it is slightly more inconvenient to have to multiply the complex fourier domain values manually (due to the half-complex packing), it is clearly beneficial in this case.

### DIFF
--- a/src/toast/ops/time_constant.py
+++ b/src/toast/ops/time_constant.py
@@ -180,12 +180,10 @@ class TimeConstant(Operator):
                     padded_start : padded_start + nsample
                 ]
 
-        # Clear the FFT plans after all observations have been processed.  If we
-        # run out of memory due to having many observations of different sizes, then
-        # we could put this line inside the loop over observations.  However,
-        # in that case we are using the plan store as just a convenience and not
-        # benefitting from the re-use of the plan creation.
-        store.clear()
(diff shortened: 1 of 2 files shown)


keskitalo (approved):
It does _look_ a whole lot more complicated but with that speed-up it is well justified!

PR #389, opened 2021-04-26
updates from design_tool_tweaks branch to master

keskitalo:
Two small tweaks to atmospheric sims and compiled kernels used in building the observation matrix in the `FilterBin` operator that has not been ported yet.

PR #482, opened 2021-09-23
Toast3 filterbin

keskitalo:
- Port the filterbin operator
- Improve handling of shared flags in the mapmaker operators
- includes #389
- make nested ordering the default.


src/toast/_libtoast/tod_pointing.cpp
```diff
@@ -131,5 +131,129 @@ void init_tod_pointing(py::module & m) {
 
     )");
 
+    m.def("healpix_pixels",
```
tskisner, line 134: It looks like you have duplicated these bindings.  Was that intentional?

src/toast/unported/todmap/atm.py
```diff
@@ -1053,18 +1053,29 @@ def load_realization(self):
...
+                    # Copy the metadata value into a dictionary, casting to correct type
+                    mdata[k] = tp(meta[k])
+                log.debug("Loaded metadata for {}".format(rname))
+            except Exception as e:
+                print(
+                    f"ERROR: failed to load cached atmosphere realization"
+                    f" from {cachefile}. Will simulate again.",
+                    flush=True,
+                )
+                mdata = None
```
tskisner, line 1070: Obviously this is in a temporary directory of toast2 code which will be deleted shortly.  If it is important to backport this to toast2, then we should open a PR against master.

src/toast/ops/filterbin.py
```diff
@@ -0,0 +1,1098 @@
...
+
+    all_data = np.hstack(all_data)
+    all_indices = np.hstack(all_indices)
+    all_indptr = np.hstack(all_indptr)
+    obs_matrix = scipy.sparse.csr_matrix((all_data, all_indices, all_indptr), shape)
+
+    log.info_rank(f"Constructed in", timer=timer, comm=None)
+
+    log.info(f"Writing {rootname}.npz ...")
+    scipy.sparse.save_npz(rootname, obs_matrix)
```
tskisner, line 108: Do we expect these files to have a long lifetime?  If so, then eventually we will need to define a format that is language-independent.  Obviously not an immediate concern!

src/toast/ops/filterbin.py
```diff
@@ -0,0 +1,1098 @@
...
+@trait_docs
+class FilterBin(Operator):
+    """OpFilterBin buids a template matrix and projects out
+    compromised modes.  It then bins the signal and optionally
+    writes out the sparse observation matrix that matches the
+    filtering operations.
+    OpFilterBin supports deprojection templates.
+
+    THIS OPERATOR ASSUMES OBSERVATIONS ARE DISTRIBUTED BY DETECTOR
+    WITHIN A GROUP
```
tskisner, line 125: Eventually we could forcibly redistribute and print a warning, but there are multiple operators that have requirements like this, so we should come up with a consistent policy on that.

src/toast/ops/filterbin.py
```diff
@@ -0,0 +1,1098 @@
...
+                    and self._deproject_pattern.match(det) is not None
+                )
+
+                if deproject or self.write_obs_matrix:
+                    # We'll need pixel numbers
+                    obs_data = data.select(obs_uid=obs.uid)
+                    self.binning.pixel_pointing.apply(obs_data, detectors=[det])
+                    pixels = obs.detdata[self.binning.pixel_pointing.pixels][det]
+                    # and weights
+                    self.binning.stokes_weights.apply(obs_data, detectors=[det])
```
tskisner, line 389: So in this scenario, we just assume that we job has enough memory to expand pixels and weights and leave them in memory in each observation?  Or do we check whether they exist and if not, create them here and delete them after each observation?

tskisner (approved):
Fine to merge after cleaning up duplicate code.  Also not sure whether you intended to expand full detector pointing in every observation and leave it there.

src/toast/_libtoast/tod_pointing.cpp
keskitalo: Nope. Fixed. Thanks!

src/toast/unported/todmap/atm.py
keskitalo: Not important enough to backport. I did make sure the feature is now included in TOAST3.

src/toast/ops/filterbin.py
keskitalo: Agreed.

src/toast/ops/filterbin.py
keskitalo: The intermediate files only exist until serially compiled into a single scipy sparse matrix (which still uses numpy save sets under the hood).

src/toast/ops/filterbin.py
tskisner: I see now that you are calling `apply()` on only a single detector and a single observation.  This is fine in the case of operators like the pixels and weights operators that do not have any operations in their `_finalize()` method, and which currently only have a CPU implementation.  In the general case, one would use a `Pipeline` that used a list of operators composed of pixels, weights, and the bulk of the code below (implemented as a another operator).  The Pipeline calls exec for each operator in turn and runs finalize at the end.  This change can wait until we have GPU kernels for those 3 things.

PR #483, opened 2021-09-23
Improve 2D polynomial filter

tskisner:
This work moves the solve step of the 2D polynomial filter into compiled code.  Still testing on cori.  Will convert from a draft PR after those tests complete.

### DIFF
--- a/src/libtoast/src/toast_tod_filter.cpp
+++ b/src/libtoast/src/toast_tod_filter.cpp
@@ -367,3 +367,114 @@ void toast::add_templates(double * signal, double * templates, double * coeff,
 
     return;
 }
+
+void toast::filter_poly2D_solve(
+    int64_t nsample, int32_t ndet, int32_t ngroup, int32_t nmode,
+    int32_t const * det_group, double const * templates, uint8_t const * masks,
+    double const * signals, double * coeff
+) {
+    // For each sample, solve for the regression coefficients.
+    // The templates are flat packed across (detectors, modes).
+    // The mask is flat packed across (samples, detectors).
+    // The signals are flat packed across (samples, detectors).
+    // The coefficients are flat packed across (samples, groups, modes).
+
+    toast::AlignedVector <double> rhs(nmode);
+    toast::AlignedVector <double> A(nmode * nmode);
+    toast::AlignedVector <double> singular_values(nmode);
+
+    int inmode = (int)nmode;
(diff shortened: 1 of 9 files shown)


tskisner:
I tested this on cori-KNL and verified that some cases which previously ran for hours until exceeding wall clock limits now actually finish.

src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -367,3 +367,120 @@ void toast::add_templates(double * signal, double * templates, double * coeff,
...
+                    rhs.data(), &inmode, singular_values.data(), &rcond_limit,
+                    &rank, work.data(), &lwork, &info
+                );
+                int64_t offset = isamp * (ngroup * nmode) + igroup * nmode;
+                if (info == 0) {
+                    // Solve was successful
+                    for (int64_t m = 0; m < nmode; ++m) {
+                        coeff[offset + m] = rhs[m];
+                    }
+                } else {
```
keskitalo, line 475: I wonder under what circumstances the minimum norm solution could not be found. I'm uncomfortable with failing silently and writing out zeros. Seems like corresponding detector samples should be flagged. I realize that the Python implementation did not address this either since the least squares solution is pretty much guaranteed to be found.

src/toast/tests/runner.py
```diff
@@ -254,4 +256,9 @@ def test(name=None, verbosity=2):
     if ret > 0:
         sys.exit(ret)
 
+    alltimers = timing.gather_timers(comm=comm)
```
keskitalo, line 259: We may be abusing the unit test system if we try to measure performance with them. Likewise for debugging. I learned just yesterday that it is utterly frustrating to debug parallel code with the unit tests when processes fail silently, only producing `error` on the terminal.

keskitalo (approved):
Thanks a lot for this critical performance improvement!

src/libtoast/src/toast_tod_filter.cpp
tskisner: But the detector samples are already flagged if the coefficients are zero:

src/toast/tests/runner.py
tskisner: I found it useful to compare timing in the unit tests for debugging.  But I will comment out the dumping for production use and just re-enable as needed.

PR #486, opened 2021-09-25
Use batched FFTs for noise simulation

tskisner:
Many years ago (toast-1.x) we used our internal batched (threaded) FFTs for noise simulation.  In toast-2.x we somehow moved away from that.  For some job configurations (few processes and many threads per process), this is quite wasteful.  This work restores use of our batched FFTs for noise simulation.  The old serial implementation remains and can be enabled by a trait which is useful for testing.  The unit tests check agreement between the old and new implementations.

Running the ground benchmark (auto case) on 50 cori-KNL nodes with 4 processes per node and 16 threads per process previously took this long to simulate the noise:
```
TOAST INFO: Simulated detector noise in 2124.09 s
```
After this change, the same step runs in:
```
TOAST INFO: Simulated detector noise in 308.51 s
```
Definitely not perfect scaling with threads, but much improved.

### DIFF
--- a/src/libtoast/include/toast/tod_filter.hpp
+++ b/src/libtoast/include/toast/tod_filter.hpp
@@ -22,8 +22,7 @@ void filter_poly2D_solve(
     int64_t nsample, int32_t ndet, int32_t ngroup, int32_t nmode,
     int32_t const * det_group, double const * templates, uint8_t const * masks,
     double const * signals, double * coeff
-);
-
+    );
 }
 
 #endif // ifndef TOAST_TOD_FILTER_HPP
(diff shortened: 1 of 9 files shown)


keskitalo (approved):
Looks good. No real comments to be made. Really appreciate the improved performance. Some surprising auto formatter changes.

tskisner:
I think the format_source.sh script has a broken version-checking logic on macos, so it may be that an older uncrustify version was run recently.  I ran this with the latest version (0.73):
```
$ uncrustify --version
Uncrustify_d-0.73.0_f
```

PR #489, opened 2021-09-29
Fix memory handling in serial noise simulation

tskisner:
This makes serial noise handling the default, since there is a large temporary memory overhead associated with the batched (threaded) case.  This also explicitly frees the memory after each timestream simulated.


keskitalo (approved):
Looks good. Feel free to merge once the tests pass.

PR #493, opened 2021-10-04
Consolodate creation of MPI communicators

tskisner:
* All split communicators are stored within the toast.Comm class,
  including the world, group, rank, node, world node-rank, group
  node-rank, and a store of row / column communicators that are
  created as needed for different values of process_rows.

* These communicators are re-used within each observation, including
  passing the node and group node-rank communicators to the
  pshmem.MPIShared constructor.

* Overall cleanup across the code of the use of these communicators.

### DIFF
--- a/src/toast/observation_data.py
+++ b/src/toast/observation_data.py
@@ -431,8 +431,8 @@ class DetDataManager(MutableMapping):
     """
 
     def __init__(self, dist):
-        self.samples = dist.samps[dist.comm_rank].n_elem
-        self.detectors = dist.dets[dist.comm_rank]
+        self.samples = dist.samps[dist.comm.group_rank].n_elem
+        self.detectors = dist.dets[dist.comm.group_rank]
         self._internal = dict()
 
     def _data_shape(self, sample_shape):
@@ -803,8 +803,8 @@ class SharedDataManager(MutableMapping):
     """
 
     def __init__(self, dist):
-        self.n_detectors = len(dist.dets[dist.comm_rank])
-        self.n_samples = dist.samps[dist.comm_rank].n_elem
+        self.n_detectors = len(dist.dets[dist.comm.group_rank])
+        self.n_samples = dist.samps[dist.comm.group_rank].n_elem
         self.comm = dist.comm
         self.comm_row = dist.comm_row
         self.comm_col = dist.comm_col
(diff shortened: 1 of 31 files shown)


src/toast/observation_data.py
```diff
@@ -431,8 +431,8 @@ class DetDataManager(MutableMapping):
     """
 
     def __init__(self, dist):
-        self.samples = dist.samps[dist.comm_rank].n_elem
-        self.detectors = dist.dets[dist.comm_rank]
+        self.samples = dist.samps[dist.comm.group_rank].n_elem
```
keskitalo, line 434: I like this new name for `comm_rank`. It is more informative.

src/toast/tests/ops_polyfilter.py
```diff
@@ -189,22 +189,22 @@ def test_polyfilter2D(self):
...
-            for idet, det in enumerate(ob.local_detectors):
-                flags = np.array(ob.shared["flags"])
-                flags |= ob.detdata["flags"][det]
-                good = flags == 0
-                signal = ob.detdata["signal"][det]
-                x = np.arange(signal.size)
-                ax.plot(x, signal, "-", label=f"{det} unfiltered")
-                ax.plot(x, good, "-", label=f"{det} input good samples")
-            ax.legend(loc="best")
+        # if data.comm.world_rank == 0:
```
keskitalo, line 192: I don't think the issue is these matplotlib commands. I was able to hang the 2D polyfilter test on my laptop without the plotting commands. The precise location of the hang also seems to depend on the order of the polyfilter unit tests. Could be some sort of memory addressing error.

src/toast/ops/sim_tod_atm.py
```diff
@@ -558,10 +562,9 @@ def _compute_absorption_and_loading(self, obs, absorption_key, loading_key, comm
         freq_min, freq_max = bandpass.get_range()
         n_freq = self.n_bandpass_freqs
         freqs = np.linspace(freq_min, freq_max, n_freq)
-        if comm is None:
-            ntask = 1
-            my_rank = 0
-        else:
+        ntask = 1
```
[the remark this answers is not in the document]
tskisner: Indeed, the original is valid python code, and I'll change it back.  Unscoped python variables is one of the language features that I hate, and so I tend to use this form as a "hint" about the scope.  But obviously that is personal preference.

PR #499, opened 2021-10-10
Benchmark helper script

tskisner:
The ground benchmark has several "setup" steps that ideally are done serially or from a different job configuration prior to running the benchmark.  This PR adds a setup script and modifies the ground benchmark to require those input files.


src/toast/scripts/toast_benchmark_ground
```diff
@@ -43,6 +52,13 @@ def parse_arguments():
     parser = argparse.ArgumentParser(
         description="Run a TOAST ground workflow scaled appropriately to the MPI communicator size and available memory."
     )
+    parser.add_argument(
```
nestordemeure, line 55: We might want to make that argument mandatory rather than optional. If a user runs the script blindly they might be thrown off by the default folder name not existing and be unclear on what to do.

Also the help might mention the script used to generate the inputs.

PR #503, opened 2021-10-15
Dc1 tweaks002

keskitalo:
* Fix issue in atmospheric simulation range calculation. Instead of measuring the azimuthal extent at the boresight elevation, use a synthetic focalplane with detectors arranged in a ring around it.
* Add default flag bits to Observation. By default, the first bit is used for flagging invalid data that gets rejected by all reduction operations.
* Fix issue in SimGround that left one sample gaps between stable scans and turnarounds
* Introduce new intervals called "throws" that encompass the entirely sweeping motion: acceleration+scan+deceleration
* Fixes to FilterBin to handle (flag) extremely short intervals
* FilterBin now supports HDF5 input/output
* Improve FilterBin unit tests
* Change observation name to include the scan and subscan index. Very useful for debugging because it identifies the exact line in the schedule file.
* Add FilterBin to the example ground workflow
* Add command line tool to co-add several maps using the noise matrices. Invert automatically when needed.


src/toast/ops/sim_ground.py
```diff
@@ -669,7 +705,7 @@ def _exec(self, data, detectors=None, **kwargs):
                 site=site,
             )
 
-            name = f"{scan.name}_{int(scan.start.timestamp())}"
+            name = f"{scan.name}-{scan.scan_indx}-{scan.subscan_indx}"
```
tskisner, line 708: As we discussed offline, this will invalidate the cached atmosphere sims for the ground benchmark.  I'll work on regenerating after this is merged.

src/toast/templates/template.py
```diff
@@ -45,14 +47,20 @@ class Template(TraitConfig):
...
-        None, allow_none=True, help="Observation detdata key for solver flags to use"
+    det_flags = Unicode(
+        defaults.det_flags,
+        allow_none=True,
+        help="Observation detdata key for solver flags to use",
     )
 
-    flag_mask = Int(0, help="Bit mask value for solver flags")
+    det_flag_mask = Int(
+        defaults.det_mask_invalid, help="Bit mask value for solver flags"
```
tskisner, line 62: Really nice to have det flags always enabled and a global definition of invalid data!

workflows/toast_sim_ground.py
```diff
@@ -416,13 +407,18 @@ def reduce_data(job, args, data):
...
+    ops.filterbin.output_dir = args.out_dir
+
     log.info_rank("Making maps", comm=world_comm)
     if args.obsmaps:
         # Map each observation separately
         timer_obs = toast.timing.Timer()
         timer_obs.start()
         group = data.comm.group
         orig_name = ops.mapmaker.name
+        orig_name_filterbin = ops.filterbin.name
```
tskisner, line 421: Thanks for this fix!

src/toast/ops/filterbin.py
```diff
@@ -309,23 +354,23 @@ def _exec(self, data, detectors=None, **kwargs):
         # Filter data
 
         self._initialize_obs_matrix()
-        log.info_rank(
+        log.debug_rank(
             "OpFilterBin: Initialized observation_matrix in",
```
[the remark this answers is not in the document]
keskitalo: Done.

PR #510, opened 2021-10-29
Update spt3g tools.

tskisner:
* Update API to match latest toast3

* Re-enable unit tests

* Fix silly performance issue where frame intervals
  were being recreated many times.

Just a note that I see an unrelated unit test failure locally when running on 4 processes (our github CI is limited to 2 processes).  This should be sorted out before merging.

### DIFF
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -439,7 +439,7 @@ class Focalplane(object):
         self.sample_rate = sample_rate
 
         if file is not None:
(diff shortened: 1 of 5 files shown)


tskisner:
As a follow up, the unit test deadlock I see occurs on the toast3 branch as well, so is unrelated to this PR.  I'll open a separate issue about that.

keskitalo (approved):
No comments. Looks good!

PR #511, opened 2021-10-29
Improve the atmospheric interpolation

keskitalo:


src/toast/ops/sim_tod_atm.py
```diff
@@ -152,7 +152,12 @@ class SimAtmosphere(Operator):
         help="Maximum wind drift before discarding the volume and creating a new one",
     )
 
-    freq = Quantity(None, allow_none=True, help="Observing frequency")
```
tskisner, line 155: Thanks for removing that stale trait since we use the detector bandpass already

src/toast/ops/sim_tod_atm.py
```diff
@@ -152,7 +152,12 @@ class SimAtmosphere(Operator):
...
+    sample_rate = Quantity(
+        None,
+        allow_none=True,
+        help="Rate at which to sample atmospheric TOD before interpolation.  "
+        "Default is no interpolation.",
+    )
```
tskisner, line 160: I like specifying the interpolation in this way (as a sampling rate).  It feels more intuitive.

tskisner (approved):
Looks good, thanks!

PR #513, opened 2021-11-02
HDF5 Observation I/O

tskisner:
This PR adds support for round-trip save and load to HDF5, using MPI-IO if available.  There may be some small tweaks to this before merging and after more tests at scale on systems at NERSC.

Still for future work is porting additional features from the tidas package.

### DIFF
--- a/src/toast/io/observation_hdf_load.py
+++ b/src/toast/io/observation_hdf_load.py
@@ -188,12 +188,21 @@ def load_hdf5(
     # enable reading detector and shared data in parallel.
     hf = None
     hfgroup = None
+    file_version = None
     if parallel:
         hf = h5py.File(path, "r", driver="mpio", comm=comm.comm_group)
         hgroup = hf
+
     else:
         hf = h5py.File(path, "r")
         hgroup = hf
+
+    # Data format version check
+    file_version = hgroup.attrs["toast_format_version"]
+    if file_version != 0:
+        msg = f"HDF5 file '{path}' using unsupported data format {file_version}"
+        log.error(msg)
+        raise RuntimeError(msg)
 
     # Observation properties
     obs_name = hgroup.attrs["observation_name"]
--- a/src/toast/io/observation_hdf_load.py
+++ b/src/toast/io/observation_hdf_load.py
@@ -5,6 +5,8 @@
 import os
 import numpy as np
 
+import datetime
+
 from astropy import units as u
 
 import h5py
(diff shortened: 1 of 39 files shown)


tskisner:
Another note- this branch includes a couple of unrelated fixes:
- Create the default detector data and flags in `SimSatellite` (analogous to what is done in `SimGround`)
- Change the MPI test runner so that if one process hits an error, the job is aborted rather than hanging at the barrier.

tskisner:
Ok @keskitalo , this is ready for review.  There may be a few small tweaks as I get more results from performance testing, but nothing drastic.  I also added a few small fixes for a timer that was being used for logging but never started and other assorted things.

src/toast/io/observation_hdf_load.py
```diff
@@ -0,0 +1,359 @@
...
+    parallel = have_hdf5_parallel()
+
+    # Open the file and get the root group.  In both serial and parallel HDF5,
+    # multiple readers are supported.  We open the file with all processes to
+    # enable reading detector and shared data in parallel.
+    hf = None
+    hfgroup = None
+    file_version = None
+    if parallel:
+        hf = h5py.File(path, "r", driver="mpio", comm=comm.comm_group)
```
keskitalo, line 193: Is there any advantage to opening the file in the parallel access mode when only reading?

src/toast/io/observation_hdf_load.py
```diff
@@ -0,0 +1,359 @@
...
+    obs_sample_sets = None
+    if hgroup.attrs["observation_sample_sets"] != "NONE":
+        obs_sample_sets = [
+            [int(x) for x in y]
+            for y in json.loads(hgroup.attrs["observation_sample_sets"])
+        ]
+
+    # Instrument properties
+
+    # FIXME:  We should add save / load methods to these classes to
```
keskitalo, line 224: I agree, this looks clunky but will do for now.

src/toast/io/observation_hdf_save.py
```diff
@@ -0,0 +1,753 @@
...
+        hgroup.attrs["observation_sample_sets"] = obs_all_samp_sets
+
+        log.verbose_rank(
+            f"{log_prefix}  Wrote observation attributes (parallel={parallel}) in",
+            comm=save_comm,
+            timer=vtimer,
+        )
+
+        # Instrument properties
+        inst_group = hgroup.create_group("instrument")
```
keskitalo, line 578: Why don't the telescope and site information get dumped with the rest of the metadata?

src/toast/ops/load_hdf5.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Class traits
+
+    API = Int(0, help="Internal interface version for this operator")
+
+    volume = Unicode(
+        None, allow_none=True, help="Top-level directory containing the data volume"
+    )
+
+    # FIXME:  We should add a filtering mechanism here to dump a subset of
```
keskitalo, line 45: "dump" -> "load"

workflows/toast_sim_ground.py
```diff
@@ -591,6 +597,7 @@ def main():
         toast.ops.TimeConstant(
             name="convolve_time_constant", deconvolve=False, enabled=False
         ),
+        toast.ops.SaveHDF5(name="save_hdf5", enabled=False),
```
keskitalo, line 600: We should also add the option to load data from file but that can wait for now.

keskitalo (approved):
Awesome! Having TOAST facilities for loading and saving TOD and associated metadata will come extremely handy.

src/toast/io/observation_hdf_load.py
tskisner: My "hope" was that using the parallel driver for reading would somehow result in fewer filesystem accesses when getting the HDF5 file information (hierarchy of groups, etc).  But I don't know what HDF5 is doing internally, so I am not sure if it is loading that on one process and broadcasting, etc.

src/toast/io/observation_hdf_load.py
tskisner: Yes, there are definitely some refinements needed here.  Just trying to get something out the door for immediate use.

src/toast/io/observation_hdf_save.py
tskisner: The idea in separating these is that the instrument information is currently stored separately in the `Observation` class (it is not part of the internal metadata dictionary).  And this instrument information is needed to construct the observation.  Keeping them separate allows us to instantiate the observation when loading the data prior to importing the metadata.

workflows/toast_sim_ground.py
tskisner: The `toast_sim_ground.py` workflow is set up primarily to do simulations from scratch starting with a scanning schedule and focalplane information.  In the case of using the `LoadHDF5` operator, we would be using other techniques to select a set of files to load.  All of the instrument information, boresight pointing, etc would already be provided, including detector data with some kinds of signal (real data or all simulated components or a subset of components, etc).  In that case I think it would make sense to have a different workflow with mostly data reduction operations.

PR #514, opened 2021-11-05
Toast3 noise estim

keskitalo:
Port the noise estimation operator.


src/toast/ops/noise_estimation.py
```diff
@@ -0,0 +1,1068 @@
...
+                shared=dup_shared,
+                detdata=dup_detdata,
+                intervals=dup_intervals,
+            )
+            log.debug_rank(
+                f"{data.comm.group:4} : Duplicated observation in",
+                comm=temp_obs.comm.comm_group,
+                timer=timer,
+            )
+            # Redistribute this temporary observation to be distributed by sample sets
```
tskisner, line 243: I think you mean distributed by arbitrary sample ranges.

tskisner (approved):
Many thanks for this huge amount of work porting and overhauling the noise estimation tools.  My only thought is that if some of the utility functions are more generally useful outside of this operator, then we could move them to a separate file in the root of the package.  However we don't need to do that until we have a use case.

PR #515, opened 2021-11-09
More I/O cleanups

tskisner:
This includes a variety of I/O related things:
 
- Add scripts to setup.py that were missing
    
- Add more timing and logging to HDF5 loading
    
- When using parallel HDF5, warn the user if the per-process buffer size exceeds the 2^31 byte limit of MPI
    
- Use dataset read_direct() and write_direct() rather than fancy indexing in case it saves a copy
    
- Add explicit typecast when loading attributes for builtin observation metadata
    
- When writing spt3g data, add an option to purge the observation data copy


keskitalo (approved):
Looks good. I'm glad to have the facility to translate HDF5 to spt3g.

PR #516, opened 2021-11-09
Dc1 tweaks004

keskitalo:
Add yield cut operator.

### DIFF
--- a/src/toast/scripts/toast_fake_focalplane
+++ b/src/toast/scripts/toast_fake_focalplane
@@ -192,7 +192,7 @@ def main():
     # Guard against being called with multiple processes
     mpiworld, procs, rank = get_world()
     if rank == 0:
-        fp.write(args.out)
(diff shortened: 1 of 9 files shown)


src/toast/scripts/toast_fake_focalplane
```diff
@@ -192,7 +192,7 @@ def main():
     # Guard against being called with multiple processes
     mpiworld, procs, rank = get_world()
     if rank == 0:
-        fp.write(args.out)
+        fp.save_hdf5(args.out)
```
tskisner, line 195: Thanks for catching this change that was missed in the previous hdf5 updates!

tskisner (approved):
Thanks for this useful feature.  In the future we could consider adding a single per-detector "observation flag" value that applies to the detector for the whole observation, and use that to cut or categorize detectors.  But this definitely works for now.

PR #519, opened 2021-11-30
Dc1 tweaks006

keskitalo:
This PR introduces a number of small tweaks:

- Issue a warning when atmospheric simulation does not have enough work for all tasks
- add a shortcut when distributing observations to a group of maximum size
- add PWV modulation support to elevation noise model
- add `thinfp` parameter to loading focalplane data
- raise the actual exception instead of `exit()` when running serially
- speed up elevation noise significantly by only expanding the pointing in the middle of the observation
- add option to apply an extra scaling to NET in the elevation noise operator
- add `realization` trait to sim_ground so that simulated weather conforms to the right realization
- fix a branching bug in interpolating azimuth for sampled atmosphere
- add helpful error messages to loading HDF5 files
- fix a bug in scheduler related to horizontal patches and observing breaks
- allow user to control the output precision in `healpix_coadd` script

### DIFF
--- a/src/toast/dist.py
+++ b/src/toast/dist.py
@@ -74,47 +74,48 @@ def distribute_discrete(sizes, groups, pow=1.0, breaks=None):
             f"{groups} groups."
         )
         raise RuntimeError(msg)
+    elif len(sizes) == groups:
+        # One chunk per group, trivial solution
+        dist = [DistRange(off, 1) for off in range(len(sizes))]
+    else:
+        chunks = np.array(sizes, dtype=np.int64)
+        weights = np.power(chunks.astype(np.float64), pow)
+        max_per_proc = float(distribute_partition(weights.astype(np.int64), groups))
+        target = np.sum(weights) / groups
 
-    chunks = np.array(sizes, dtype=np.int64)
-    weights = np.power(chunks.astype(np.float64), pow)
-    max_per_proc = float(distribute_partition(weights.astype(np.int64), groups))
+        dist = []
 
-    target = np.sum(weights) / groups
-
-    dist = []
-
-    off = 0
-    curweight = 0.0
-
-    all_breaks = None
-    if breaks is not None:
-        # Check that the problem makes sense
-        all_breaks = np.unique(breaks)
-        all_breaks = all_breaks[all_breaks > 0]
-        all_breaks = all_breaks[all_breaks < chunks.size]
-        all_breaks = np.sort(all_breaks)
-        if all_breaks.size + 1 > groups:
-            raise RuntimeError(
-                "Cannot divide {} chunks to {} groups with {} breaks.".format(
-                    chunks.size, groups, all_breaks.size
+        all_breaks = None
+        if breaks is not None:
+            # Check that the problem makes sense
+            all_breaks = np.unique(breaks)
+            all_breaks = all_breaks[all_breaks > 0]
+            all_breaks = all_breaks[all_breaks < chunks.size]
+            all_breaks = np.sort(all_breaks)
+            if all_breaks.size + 1 > groups:
+                raise RuntimeError(
+                    "Cannot divide {} chunks to {} groups with {} breaks.".format(
+                        chunks.size, groups, all_breaks.size
+                    )
                 )
-            )
 
-    at_break = False
-    for cur in range(0, weights.shape[0]):
-        if curweight + weights[cur] > max_per_proc or at_break:
-            dist.append(DistRange(off, cur - off))
-            over = curweight - target
-            curweight = weights[cur] + over
-            off = cur
-        else:
-            curweight += weights[cur]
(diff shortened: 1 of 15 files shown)


src/toast/dist.py
```diff
@@ -74,47 +74,48 @@ def distribute_discrete(sizes, groups, pow=1.0, breaks=None):
...
-        # Check that the problem makes sense
-        all_breaks = np.unique(breaks)
-        all_breaks = all_breaks[all_breaks > 0]
-        all_breaks = all_breaks[all_breaks < chunks.size]
-        all_breaks = np.sort(all_breaks)
-        if all_breaks.size + 1 > groups:
-            raise RuntimeError(
-                "Cannot divide {} chunks to {} groups with {} breaks.".format(
-                    chunks.size, groups, all_breaks.size
+    elif len(sizes) == groups:
```
tskisner, line 77: That seems like a reasonable short-circuit case.

src/toast/instrument.py
```diff
@@ -423,6 +425,7 @@ class Focalplane(object):
         file (str):  Load the focalplane from this file.
         comm (MPI.Comm):  If loading from a file, optional communicator to broadcast
             across.
+        thinfp (int):  Only sample the detectors in the file.
```
tskisner, line 428: I agree that this seems like a generally useful feature.  The resulting "thinned" focalplane will depend on how an experiment has indexed their detectors, but if they want some other way of sampling a subset of detectors, then an experiment would provide those tools and not use this feature.

src/toast/ops/elevation_noise.py
```diff
@@ -115,55 +155,69 @@ def _exec(self, data, detectors=None, **kwargs):
             log.error(msg)
             raise RuntimeError(msg)
 
-        for ob in data.obs:
-            focalplane = ob.telescope.focalplane
+        for obs in data.obs:
```
tskisner, line 158: I have some other changes to the `ElevationNoise` operator (plus adding unit tests) in a branch where I am working through some HDF5 unit test failures.  So I will have to manually merge these changes.  Is there a reason you change the iteration variable `ob` --> `obs`?  Just preference?  I guess it doesn't matter since I'll have to merge this line-by-line with the other changes, but will try to make sure that the name is used consistently when doing that.

src/toast/ops/sim_tod_atm_utils.py
```diff
@@ -260,7 +260,12 @@ def _exec(self, data, detectors=None, **kwargs):
                             (tmax_det - tmin_det) * self.sample_rate.to_value(u.Hz)
                         )
                         t_interp = np.linspace(tmin_det, tmax_det, n_interp)
-                        az_interp = np.interp(t_interp, times[good], az)
+                        # Az is discontinuous if we scan across az=0.  To interpolate,
+                        # we must unwrap it first ...
+                        az_interp = np.interp(t_interp, times[good], np.unwrap(az))
+                        # ... however, the checks later assume 0 < az < 2pi
```
tskisner, line 266: Thanks for fixing this!

src/toast/ops/sim_ground.py
```diff
@@ -92,6 +92,8 @@ class SimGround(Operator):
         help="Name of built-in weather site (e.g. 'atacama', 'south_pole') or path to HDF5 file",
     )
 
+    realization = Int(0, help="The realization index")
```
tskisner, line 95: Should we call this "weather_realization" or something to indicate what it is being used for?

tskisner (approved):
Fine to merge as-is, comments are mostly seeking clarification.

src/toast/ops/elevation_noise.py
keskitalo: I cannot recall what the logic behind changing the name was... However, we should pick either `ob` or `obs` as the canonical name and use it throughout. TOAST-2 is pretty consistent about using `obs` but sizable fraction of TOAST-3 switched to `ob`.

src/toast/ops/sim_ground.py
keskitalo: All operators producing randomized output should have a realization index trait. What that index refers to should be apparent from the context. Do you agree?

src/toast/ops/elevation_noise.py
tskisner: Correct, much of the code was switched to `ob` to avoid confusion with the `Data.obs` member.

src/toast/ops/sim_ground.py
tskisner: I guess my point was that the the `SimGround` operator is simulating a lot of stuff:  telescope pointing, weather, etc.  Is the `realization` for the weather?  The telescope pointing?  The HWP angle?  I guess we could rename that if we are ever simulating different realizations of different types of data.  Or we could derive the realization indices for different things from one input realization number.

PR #520, opened 2021-11-30
Update pybind11 to 2.8.1

tskisner:


tskisner:
Self-merging this, since it only updates the bundled pybind11 package.

PR #521, opened 2021-12-01
Prototyping OpenACC interfaces

tskisner:
This is just a draft PR for code discussion.


tskisner:
This work adds python bindings for several OpenACC memory movement functions, which allows us to control memory movement from python.  The pre-staged memory can then be passed to compiled functions that use OpenACC with the `data present()` clause.  The build and runtime environments is controlled by a variety of switches

**CMake**
```
-DUSE_CUDA=TRUE :  If set, attempt to find CUDA toolkit or NVHPC
-DUSE_OPENACC=TRUE : If set, test for compiler flags needed to build OpenACC code
-DUSE_OPENACC_MEMPOOL=TRUE : If both CUDA and OpenACC are FOUND, then transparently 
    use a CUDA mempool instead of the acc_copyin / acc_copyout functions.
```
**Environment**
```
CUDA_MEMPOOL_FRACTION=0.5  # For example
```
This defines the amount of per-process GPU memory to claim for the memory pool.  For example, if there are 8 MPI processes running on a node with 4 GPUs, then two processes will be assigned to each GPU.  In that case, each process will be allowed to allocate up to half of the GPU memory.  So in this scenario if `CUDA_MEMPOOL_FRACTION` is set to 0.5, then each process will allocate 0.25 of the GPU memory.

nestordemeure (approved):
The miniapp runs, using cuda, on perlmutter. Everything seems good to me.

src/libtoast/src/toast_math_linearalgebra.cpp
```diff
@@ -32,18 +32,20 @@ void toast::LinearAlgebra::gemm(char TRANSA, char TRANSB, int M, int N,
...
     // prepare inputs
     cublasOperation_t transA_gpu = (TRANSA == 'T') ? CUBLAS_OP_T : CUBLAS_OP_N;
     cublasOperation_t transB_gpu = (TRANSB == 'T') ? CUBLAS_OP_T : CUBLAS_OP_N;
 
     // send data to GPU
-    double * A_gpu = GPU_memory_pool.toDevice(A, LDA * ((TRANSA == 'N') ? K : M));
-    double * B_gpu = GPU_memory_pool.toDevice(B, LDB * ((TRANSB == 'N') ? N : K));
-    double * C_gpu = (BETA == 0.) ? GPU_memory_pool.alloc <double> (LDC * N)
-                                 : GPU_memory_pool.toDevice(C, LDC * N);
+    double * A_gpu = pool.toDevice(A, LDA * ((TRANSA == 'N') ? K : M));
```
keskitalo, line 42: I'm not a fan of this syntax. It is concise but usually makes the code less readable. Obviously a matter of preference.

src/toast/_libtoast/accelerator.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+            std::ostringstream o;
+            o << "delete device mem for host pointer " << p_host << " (" << n_bytes << " bytes)";
+            log.verbose(o.str().c_str());
+
+            # ifdef USE_OPENACC_MEMPOOL
+
+            // FIXME:  If there was a pool.Free(cpu pointer), we would call that here.
+            auto & pool = GPU_memory_pool::get();
+            pool.fromDevice(reinterpret_cast <char *> (p_host));
```
keskitalo, line 314: Semantically, `pool.fromDevice()` and `acc_delete()` do not seem equivalent. One implies retrieving the data from the card while other implies freeing the device memory. If `acc_delete()` also updates the host pointer, perhaps that should be part of the routine name: `acc_sync_and_delete()`.

src/toast/_libtoast/accelerator.cpp
```diff
@@ -0,0 +1,401 @@
...
+    register_stub(m, "acc_is_present");
+    register_stub(m, "acc_copyin");
+    register_stub(m, "acc_copyout");
+    register_stub(m, "acc_update_device");
+    register_stub(m, "acc_update_self");
+    register_stub(m, "acc_delete");
+
+    #endif // HAVE_OPENACC
+
+    // Small test code used by the unit tests.
```
keskitalo, line 342: This is a curious place to add test code. Is this the permanent location?

src/toast/atm.py
```diff
@@ -17,13 +17,19 @@
...
 
-from ._libtoast import (
-    atm_sim_compute_slice,
-    atm_sim_observe,
-    atm_sim_compress_flag_hits_rank,
-    atm_sim_compress_flag_extend_rank,
-    atm_sim_kolmogorov_init_rank,
-)
+available_atm = None
+if available_atm is None:
```
keskitalo, line 21: This test is unnecessary

src/toast/ops/operator.py
```diff
@@ -134,23 +150,48 @@ def provides(self):
...
         Returns:
-            (list):  List of pre-defined accelerator names supported by this
-                operator (and by TOAST).
+            (bool):  True if the operator can use OpenACC, else False.
 
         """
-        return self._accelerators()
+        return self._supports_acc()
+
+    def _use_acc(self, data):
```
keskitalo, line 172: Are we positive it should not be an error condition to ask for GPU acceleration when not all data is staged to the card?

src/toast/ops/pipeline.py
```diff
@@ -133,22 +149,30 @@ def _exec(self, data, detectors=None, **kwargs):
                         pstr, op.name
                     )
                     log.verbose(msg)
-                    op.exec(data, detectors=selected_set)
+                    if use_acc:
```
keskitalo, line 152: This pattern gets repeated. Would it not make more sense to have `op.exec(data, detectors=selected_set, use_acc=use_acc)`?

keskitalo (approved):
Can't wait to see the hackathon work build on this foundation!

src/libtoast/src/toast_math_linearalgebra.cpp
tskisner: Can you clarify a bit?  Are you talking about the original function signature (which is unmodified in this PR) or the use of a singleton class rather than a global variable?

src/toast/_libtoast/accelerator.cpp
tskisner: The pool currently does not provide some methods (update device and update host), and I agree that to match the syntax, the pool should eventually support a method that calls `Free` with a host pointer lookup.  Note that the names in this file are designed to match the function names in the OpenACC standard.

src/toast/_libtoast/accelerator.cpp
tskisner: It is compiled code that is only used in the unit tests for this source file.  I suppose we should eventually put it in a separate file like `_libtoast/test_accelerator.cpp`.

src/libtoast/src/toast_math_linearalgebra.cpp
keskitalo: The original signature with the in-line conditional.

src/toast/atm.py
tskisner: What is your recommended way to ensure that the following import try/except is executed exactly once on first import of the file, and that the `available_atm` package variable is set to True or False for all subsequent uses?

src/toast/ops/operator.py
tskisner: Each operator can support OpenACC, but the decision of whether to use it or not is controlled by the argument to `_exec()`.  This allows us to disable the OpenACC codepath if we know from the calling code that it does not make sense.  Similarly, the Pipeline operator can check if all its operators support accelerators, and can ensure that the inputs are all staged.

src/toast/_libtoast/accelerator.cpp
nestordemeure: What are the methods that would be needed? (I could add them this morning)
- free(cpu_ptr)
- update(cpu_ptr)

src/toast/ops/pipeline.py
tskisner: True- the `_exec()` methods of the operators do support kwargs, so even it they are not looking for a `use_acc=False` kword, it should do no harm.

src/toast/atm.py
keskitalo: Hah! I didn't realize that the declaration of `available_atm` would not be executed more than once. Withdrawn.

keskitalo: And back on the table. Seems like none of the code in a previously imported file will execute again

src/toast/_libtoast/accelerator.cpp
nestordemeure: Done!

src/toast/atm.py
tskisner: I changed this and another place in the code with a similar pattern.

src/toast/ops/pipeline.py
tskisner: Fixed.

PR #522, opened 2021-12-01
Toast3 gpu fft

nestordemeure:
includes:
- GPU recycling allocator
- cublas
- cuFFT


nestordemeure:
I would like to do two things before the merge:
- simplify the fromdevice function to make it closer to the openAcc one
- check wether I can get finegrained failure information from the batched cublas functions

nestordemeure:
I just improved the documentation of the gpu preallocator and finished the simplification.

You can push/pull memory to GPU (using preallocated memory) using the following functions:

```cpp
// get gpu memory, copies given number of elements onto gpu memory
double* gpu_ptr = toDevice(cpu_ptr, nb_elements);

// get data back in cpu memory
fromDevice(cpu_ptr);
```

nestordemeure:
The cublas code is now updated, all the things are ready for the merge on my side.

PR #523, opened 2021-12-07
Toast3 gpupool

nestordemeure:
I cleaned and streamlined the memory pool code (I was unhappy with some code duplication and potential linear complexity).

### DIFF
--- a/src/libtoast/include/toast/gpu_helpers.hpp
+++ b/src/libtoast/include/toast/gpu_helpers.hpp
@@ -10,6 +10,7 @@
 #include <string>
 #include <vector>
 #include <mutex>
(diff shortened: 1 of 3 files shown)


nestordemeure:
I want to run the miniapp on that branch once perlmutter is back before merging but everything should be done.

tskisner:
These changes look good.  I can't think of a case where cuda malloc / free would be called from a threaded region, but the locking here does no harm.

tskisner (approved):

nestordemeure:
> I can't think of a case where cuda malloc / free would be called from a threaded region, but the locking here does no harm.

It happens once where calling some linear algebra I believe.

PR #530, opened 2022-01-20
General improvements ported from the hackathon efforts:

tskisner:
* Add "global" operator dependency type for objects in the data
  dictionary.

* Clean up incorrect operator dependencies.

* Add a boolean return to the detdata ensure() method, so that
  operators can tell if the data was already existing and short
  circuit calculations.

### DIFF
--- a/src/toast/ops/sim_tod_atm.py
+++ b/src/toast/ops/sim_tod_atm.py
@@ -303,13 +303,23 @@ class SimAtmosphere(Operator):
         loading_key = f"{self.name}_loading"
 
         # Set up the observing operator
+        if self.shared_flags is None:
+            # Cannot observe samples that have no pointing
+            shared_flags = self.detector_pointing.shared_flags
+            shared_flag_mask = self.detector_pointing.shared_flag_mask
+        else:
+            # Trust that the user has provided a flag that excludes smaples
+            # without pointing
+            shared_flags = self.shared_flags
+            shared_flag_mask = self.shared_flag_mask
+
         observe_atm = ObserveAtmosphere(
             times=self.times,
             det_data=self.det_data,
             quats=self.detector_pointing.quats,
             view=temporary_view,
-            shared_flags=self.shared_flags,
-            shared_flag_mask=self.shared_flag_mask,
+            shared_flags=shared_flags,
+            shared_flag_mask=shared_flag_mask,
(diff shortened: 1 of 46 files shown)


nestordemeure (approved):
Looking good (liking the fact that it adds code but also comments). Should be okay to merge as soon as it passes tests.

src/toast/ops/sim_tod_atm.py
```diff
@@ -303,13 +303,23 @@ def _exec(self, data, detectors=None, **kwargs):
         loading_key = f"{self.name}_loading"
 
         # Set up the observing operator
+        if self.shared_flags is None:
+            # Cannot observe samples that have no pointing
+            shared_flags = self.detector_pointing.shared_flags
+            shared_flag_mask = self.detector_pointing.shared_flag_mask
+        else:
+            # Trust that the user has provided a flag that excludes smaples
```
keskitalo, line 311: smaples -> samples

src/toast/ops/stokes_weights.py
```diff
@@ -168,6 +168,23 @@ def _exec(self, data, detectors=None, **kwargs):
                 # Nothing to do for this observation
                 continue
 
+            # Create (or re-use) output data for the weights
+
+            if self.single_precision:
+                existed = ob.detdata.ensure(
```
keskitalo, line 174: This PR uses `exists` and `existed` for the return value. Are we happy with that or should we pick one and use it consistently?

src/toast/scripts/toast_mini
```diff
@@ -117,6 +119,10 @@ def parse_arguments():
     args.psd_net = 50.0e-6
     args.psd_fmin = 1.0e-5
 
+    # DEBUGGING
```
keskitalo, line 122: Are these left here intentionally?

keskitalo (approved):
Go ahead and merge as soon as the tests pass, unless you want to act on the small comments.

src/toast/ops/sim_tod_atm.py
tskisner: fixed

src/toast/ops/stokes_weights.py
tskisner: Good point, changed to "exists" everywhere.

PR #532, opened 2022-01-31
Work in progress cleaning up HDF5 noise model I/O.

tskisner:
Opening a draft PR for testing and debugging

### DIFF
--- a/src/toast/atm.py
+++ b/src/toast/atm.py
@@ -1133,8 +1133,8 @@ class AtmSim(object):
         self._nr = 1000
 
         self._rmin_kolmo = 0.0
-        diag = np.sqrt(self._delta_x ** 2 + self._delta_y ** 2)
-        self._rmax_kolmo = np.sqrt(diag ** 2 + self._delta_z ** 2) * 1.01
+        diag = np.sqrt(self._delta_x**2 + self._delta_y**2)
+        self._rmax_kolmo = np.sqrt(diag**2 + self._delta_z**2) * 1.01
 
         self._rstep = (self._rmax_kolmo - self._rmin_kolmo) / (self._nr - 1)
 
--- a/src/toast/tests/instrument.py
+++ b/src/toast/tests/instrument.py
@@ -18,6 +18,8 @@ from ..instrument import Focalplane
 
 from ..instrument_sim import fake_hexagon_focalplane
 
+from ..io import hdf5_open
+
 from ._helpers import create_outdir
 
 
@@ -35,17 +37,24 @@ class InstrumentTest(MPITestCase):
         fp_file = os.path.join(self.outdir, "focalplane.h5")
         check_file = os.path.join(self.outdir, "check.h5")
 
-        if self.comm is None or self.comm.rank == 0:
-            fp.save_hdf5(fp_file)
+        hf = hdf5_open(fp_file, "w", comm=self.comm)
+        fp.save_hdf5(hf, comm=self.comm)
+        if hf is not None:
+            hf.close()
+        del hf
+
         if self.comm is not None:
             self.comm.barrier()
 
-        newfp = Focalplane(file=fp_file, comm=self.comm)
+        newfp = Focalplane()
 
-        if self.comm is None or self.comm.rank == 0:
-            newfp.save_hdf5(check_file)
-        if self.comm is not None:
-            self.comm.barrier()
+        hf = hdf5_open(fp_file, "r", comm=self.comm)
+        newfp.load_hdf5(hf, comm=self.comm)
+        if hf is not None:
+            hf.close()
+        del hf
+
+        self.assertTrue(newfp == fp)
 
     def test_focalplane_full(self):
         names = ["det_01a", "det_01b", "det_02a", "det_02b"]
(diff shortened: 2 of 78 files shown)


tskisner:
Ok, this is ready for review.  This work includes some structural improvements to the HDF5 I/O:
    
* Checking for whether we have MPI-enabled HDF5 and are using it,
  as well as opening files in either mode are moved into helper
  functions.
    
* Helper functions for parallel Table write and the save_hdf5 /
   load_hdf5 methods of classes are changed to work only with
   already-open h5py.Group objects.  Opening of the files are
   moved to higher-level code.
    
* Bug fix when creating intervals from a global list, in the
  case where that list is empty.
    
* When saving noise models and grouping PSDs by their common
  frequency arrays, build the hash of that array on one process,
  to avoid LSB fluctuations causing different hash values on
  different processes.
    
* Remove duplicate code in the observation HDF5 save / load
  functions.
    
* Add a new trait to the SaveHDF5 operator that casts detector
  data in float64 to float32.

src/toast/atm.py
```diff
@@ -1133,8 +1133,8 @@ def _initialize_kolmogorov(self):
         self._nr = 1000
 
         self._rmin_kolmo = 0.0
-        diag = np.sqrt(self._delta_x ** 2 + self._delta_y ** 2)
-        self._rmax_kolmo = np.sqrt(diag ** 2 + self._delta_z ** 2) * 1.01
+        diag = np.sqrt(self._delta_x**2 + self._delta_y**2)
```
keskitalo, line 1136: Looks like `black` version 22.1 changes the default on space around `**`. Given that `pip` still installs `black` version 21.12. we should expect some noise from this in the commit history. Or should we require 22.1 to run `format_source.sh`?

src/toast/tests/instrument.py
```diff
@@ -87,12 +96,22 @@ def test_focalplane_full(self):
...
+        fp.save_hdf5(hf, comm=self.comm)
+        if hf is not None:
+            hf.close()
+        del hf
+
         if self.comm is not None:
             self.comm.barrier()
 
-        newfp = Focalplane(file=fp_file, comm=self.comm)
+        newfp = Focalplane()
```
keskitalo, line 108: This new interface inflates the amount of code needed to load a focalplane in the workflow from 1 line to 6 lines. Sometimes that is unavoidable but could that complexity be hidden in the `Focalplane` class?

src/toast/tests/noise.py
```diff
@@ -39,15 +43,27 @@ def test_noise_hdf5(self):
...
             {x: u.Quantity(freqs, u.Hz) for x in streams},
-            {x: u.Quantity(np.ones(len(freqs)), u.K ** 2 * u.second) for x in streams},
+            {x: u.Quantity(np.ones(len(freqs)), u.K**2 * u.second) for x in streams},
             mixmatrix=mix,
         )
 
         nse_file = os.path.join(self.outdir, "noise.h5")
 
-        nse.save_hdf5(nse_file, comm=self.comm)
+        hf = hdf5_open(nse_file, "w", comm=self.comm)
```
keskitalo, line 52: This is another place where we could move boiler plate into the `Noise` class.

workflows/toast_sim_ground.py
```diff
@@ -154,13 +154,22 @@ def load_instrument_and_schedule(args, comm):
...
+            sample_rate=sample_rate, thinfp=args.thinfp
         )
+        hf = toast.io.hdf5_open(args.focalplane, "r", comm=comm, force_serial=True)
+        focalplane.load_hdf5(hf, comm=comm)
+        if hf is not None:
+            hf.close()
+        del hf
         log.info_rank(f"Saving focalplane to {fname_pickle}", comm=comm)
-        pickle.dump(focalplane, open(fname_pickle, "wb"))
+        if comm is None or comm.rank == 0:
```
keskitalo, line 171: Wow! I can't believe it even worked without this...

keskitalo (approved):
Only small comments. Merge away when ready.

src/toast/atm.py
tskisner: Resolved offline, latest PyPI version is 22.1.  I think our intention is to track the defaults provided by latest `black`, because it relieves us of having to think about it.  However, I agree it does make some commit noise when those defaults change.

src/toast/tests/noise.py
tskisner: As discussed offline, having all of the lower level classes only work with `h5py.Group` objects relieves those low-level functions from having to handle all the different types of possible cases.  The downside is that it adds more lines to the higher level code to open the HDF5 file and close it if the local process has a valid handle.  I'll add a small context manager around hdf5_open() to clean up some of this.

src/toast/tests/instrument.py
tskisner: The latest commit introduces an `H5File` wrapper class that acts as a context manager for opening HDF5 files in parallel or serial or on one process of a communicator.  This still expands 1 line into 3 lines, but hopefully makes it easier for high-level code to open a file and load simple objects (like a focalplane or noise model) without having to add duplicated file handling features to the focalplane and noise classes (and soon the Telescope, Site, Weather, etc classes).

PR #533, opened 2022-01-31
Fix the ElevationNoise operator

tskisner:
Ensure that the elevation noise operator always creates a new noise model, rather than changing it in place.  The input noise model may be a purely analytic one, and so it should not be modified in place.


keskitalo (approved):
Awesome!

PR #505, opened 2021-10-21
implementation of TEB convolution

giuspugl:
This implementation follows the approach implemented by Davide Maino. We perform  3 convolutions,
    obtained by spliting both the sky into their {T,E,B} components and the beam alm
    into the {T, P}   ones, with P being defined as `P=b_{lm}^E + b_{lm}^B`
    The convolution  is performed by  coupling each IQU component of the signal propertly as:
    `skyT_lm * beamT_lm, skyE_lm * Re{beamP_lm}, skyB_lm * Im{beamP_lm}`. In this way, we are able to account for the polarized components of the simulated GRASP beams that were neglected with the Operator in https://github.com/hpc4cmb/toast/blob/38a5b2c5c79500d75811d23807a7d27f557d5818/src/toast/ops/conviqt.py#L571 

For extra details please refer to [this note ](https://giuspugl.github.io/reports/Notes_TEB_convolution.html)

PR #534, opened 2022-02-01
TEB convolution  in TOAST3  

giuspugl:
As suggested by @keskitalo , open a new PR from the developments in  https://github.com/hpc4cmb/toast/pull/505 

This implementation follows the approach implemented by Davide Maino. We perform  3 convolutions,     obtained by splitting both the sky into their {T,E,B} components and the beam alm     into the {T, P}   ones, with P being defined as `P=b_{lm}^E + b_{lm}^B`     The convolution  is performed by  coupling each IQU component of the signal propertly as:
    `skyT_lm * beamT_lm, skyE_lm * Re{beamP_lm}, skyB_lm * Im{beamP_lm}`. In this way, we are able to account for the polarized components of the simulated GRASP beams that were neglected with the Operator in https://github.com/hpc4cmb/toast/blob/38a5b2c5c79500d75811d23807a7d27f557d5818/src/toast/ops/conviqt.py#L571 

For extra details please refer to [this note ](https://giuspugl.github.io/reports/Notes_TEB_convolution.html)

### DIFF
--- a/src/toast/ops/conviqt.py
+++ b/src/toast/ops/conviqt.py
@@ -237,6 +237,9 @@ class SimConviqt(Operator):
         if self.detector_pointing is None:
             raise RuntimeError("detector_pointing cannot be None.")
 
+        if self.hwp_angle is not None:
+            raise RuntimeError("Standard conviqt operator cannot handle HWP angle")
+
         log = Logger.get()
 
         timer = Timer()
@@ -265,9 +268,8 @@ class SimConviqt(Operator):
 
             detector = self.get_detector(det)
 
-            theta, phi, psi, psi_pol = self.get_pointing(data, det, verbose)
-            np.savez(f'{self.det_data}_{det}.npz', psi=psi, psi_pol=psi_pol,
-                    theta=theta, phi=phi) 
+            theta, phi, psi, psi_pol, hwp_angle = self.get_pointing(data, det, verbose)
+
             pnt = self.get_buffer(theta, phi, psi, det, verbose)
             del theta, phi, psi
 
@@ -311,7 +313,7 @@ class SimConviqt(Operator):
             raise RuntimeError(f"focalplane does not include {det}")
         props = focalplane[det]
         if "psi_pol" in props.colnames:
-            psi_pol = props["pol_angle"].to_value(u.radian)
+            psi_pol = props["psi_pol"].to_value(u.radian)
         elif "pol_angle" in props.colnames:
             warnings.warn(
                 "Use psi_pol and psi_uv rather than pol_angle", DeprecationWarning
@@ -348,12 +350,14 @@ class SimConviqt(Operator):
             epsilon = 0
         return epsilon
 
-    def get_sky(self, skyfile, det, verbose):
+    def get_sky(self, skyfile, det, verbose, pol=None):
         timer = Timer()
         timer.start()
+        if pol is None:
+            pol = self.pol
         sky = conviqt.Sky(
             self.lmax,
-            self.pol,
+            pol,
             skyfile,
             self.fwhm.to_value(u.arcmin),
             self.comm,
@@ -366,10 +370,12 @@ class SimConviqt(Operator):
             timer.report_clear(f"initialize sky for detector {det}")
         return sky
 
-    def get_beam(self, beamfile, det, verbose):
+    def get_beam(self, beamfile, det, verbose, pol=None):
         timer = Timer()
         timer.start()
-        beam = conviqt.Beam(self.lmax, self.beammmax, self.pol, beamfile, self.comm)
+        if pol is None:
+            pol = self.pol
+        beam = conviqt.Beam(self.lmax, self.beammmax, pol, beamfile, self.comm)
         if self.normalize_beam:
             beam.normalize()
         if verbose:
@@ -393,7 +399,7 @@ class SimConviqt(Operator):
         nullquat = np.array([0, 0, 0, 1], dtype=np.float64)
         timer = Timer()
         timer.start()
-        all_theta, all_phi, all_psi, all_psi_pol = [], [], [], []
+        all_theta, all_phi, all_psi, all_psi_pol, all_hwp_angle = [], [], [], [], []
         for obs in data.obs:
             if det not in obs.local_detectors:
                 continue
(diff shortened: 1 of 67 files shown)


src/toast/ops/conviqt.py
```diff
@@ -658,3 +658,97 @@ def _exec(self, data, detectors=None, **kwargs):
...
+
+            theta, phi, psi, psi_pol = self.get_pointing(data, det, verbose)
+
+            # I-beam convolution
+            pnt = self.get_buffer(theta, phi, psi, det, verbose)
+
+            convolved_data = self.convolve(sky, beam_T, detector, pnt, det, verbose)
+
+            del pnt,
+            # Q-beam convolution
```
keskitalo, line 727: I presume the comment should say `E-beam convolution`?

src/toast/tests/ops_sim_tod_conviqt.py
```diff
@@ -282,6 +282,89 @@ def test_sim(self):
...
+            pixel_pointing=pixels,
+            stokes_weights=weights,
+            noise_model=default_model.noise_model,
+            sync_type="alltoallv",
+        )
+        binner.apply(data)
+
+        # Study the map on the root process
+
+        toast_bin_path = os.path.join(self.outdir, "toast_bin.fits")
```
keskitalo, line 352: I would feel more confident if this version of the operator was compared to another version in a case where we expect identical results.

src/toast/ops/conviqt.py
```diff
@@ -658,3 +658,97 @@ def _exec(self, data, detectors=None, **kwargs):
...
+            pnt = self.get_buffer(theta, phi, psi, det, verbose)
+            convolved_data += np.cos(2 * psi_pol) * self.convolve(
```
keskitalo, line 729: I am confused about this line here. The note linked in the description says that
``` Q + iU = Sum[(b^E + ib^B)(a^E + ia^B)] * D ```
which implies that
```
Q = Sum[b^E * a^E - b^B * a^B]
U = Sum[b^E * a^B + b^B * a^E]
```
but on this line we seem to be evaluating `Sum[b^E * (a^E + ia^B)]` -- a complex-valued quantity. The issue may be completely semantical but I propose you implement a unit test that compares the standard conviqt operator to this new one. In the absence of a HWP, they should agree.

**Review by keskitalo — CHANGES_REQUESTED** [review]
Thanks a lot for wrapping this work up! I got so confused looking at the code that I had to ask for a more extensive unit testing. The current test only shows that the code runs.

File: src/toast/tests/ops_sim_tod_conviqt.py (hunk 2 above)
Comment (giuspugl) [reply src/toast/tests/ops_sim_tod_conviqt.py:352]: sounds good ! will implment a test

File: src/toast/ops/conviqt.py (hunk 1 above)
Comment (giuspugl) [reply src/toast/ops/conviqt.py:727]: It is the convolution with the E-beam, resulting in the Q signal

File: src/toast/ops/conviqt.py (hunk 3 above)
Comment (giuspugl) [reply src/toast/ops/conviqt.py:729]: This is a good point! IT might also linked to https://github.com/hpc4cmb/toast/pull/534#pullrequestreview-869602524 
The way, I implemented this convolution indeed is  as you suggest, meaning that the beams i use in the convolution of Q and U signals are respectively 
```
beam_E = [0, b^E , -b^B ]
beam_B = [0, b^B,b^E]
```

giuspugl: see https://github.com/hpc4cmb/toast/blob/73c758831f3f694ef1d483e53f02de2f928bb929/src/toast/tests/ops_sim_tod_conviqt.py#L113
and https://github.com/hpc4cmb/toast/blob/73c758831f3f694ef1d483e53f02de2f928bb929/src/toast/tests/ops_sim_tod_conviqt.py#L126

src/toast/ops/conviqt.py
mreineck: Sorry for interrupting with stupid questions: but don't you need both E and B components to compute the Q signal? There is no one-to-one correspondence between E and Q (and B and U), only between E/B and Q/U together, correct?
If you have a pointer where I can read more about this, I'd be really grateful.

giuspugl: Correct the sky E and B alms are convolved with the _polarized _ beam  and both participate to the Q and U  signals. 
have a look at  [this note](https://giuspugl.github.io/reports/Notes_TEB_convolution.html) for some reference

mreineck: Thanks!
So if I understand correctly this still makes some assumptions about the beam, and if we ever wnt to tackle the fully general problem, we have to look at the expressions in https://arxiv.org/abs/2012.10437v1 . I'm still working on that, but I honestly have no idea when that will be ready...
How far will the current approximations take us in terms of LiteBIRD simulations? Will they be good enough for all practical purposes?

giuspugl: The improvements apply mainly to polarization data , now we can include the polarized component of the grasp beam and do not assume it to be the same as the unpolarized one.  As a result this approach yields to lower side lobe pick up than the latter approximation .

src/toast/tests/ops_sim_tod_conviqt.py
giuspugl: The tests have been reviewed there are major changes involving mainly the scanning strategy (to solve for QU pixels we need more >3 hits), the way the input beam is simulated. 
See  (https://github.com/hpc4cmb/toast/blob/aa03ca92dad37a57a5c5f6afabd65fb61bdc9787/src/toast/tests/ops_sim_tod_conviqt.py#L35)

giuspugl: The operators `SimWeightedConviqt` and `SimTEBConviqt` are tested and compared with `SimConviqt` as they are expect to produce exactly the same convolution outputs in the absence of  spinning HWP

mreineck:
Do you want me to also port this extension over to the `totalconvolve` operator?

giuspugl:
@mreineck this is a good idea! i would firstly merge this PR and then port this to the `totalconvolve` operator !

giuspugl:
In #550 
-we corrected a bug in  the implementation of the TEB convolution
- implemented a new test for convolution w/ spinning HWP 

the PR #550 got merged  into this branch and  I think we're ready to final review.

keskitalo:
The functional parts of this PR were merged into `toast3` in #552. Thanks a lot for this work, @giuspugl !

PR #536, opened 2022-02-07
Fix unit test hang when running with >= 4 processes

### LINKED ISSUE #512: Strange unit test deadlock
Running the unit tests on 4 MPI processes (toast3 branch) I get a reproducible deadlock.  This occurs in the `ops_groundfilter` tests, and during the generation of the ground schedule on one process.  While the other 3 processes are waiting, the root process hangs while adding arguments to the argparser (checked with simple print statements around each call to `add_argument()`).  This is a strange and arbitrary location for a problem, and running the unit tests with just this test suite succeeds.  Because of this, it seems like some kind of memory stomping issue.

I am working on investigating with valgrind.

tskisner:
These small changes fix #512.

### DIFF
--- a/src/toast/ops/polyfilter.py
+++ b/src/toast/ops/polyfilter.py
@@ -401,6 +401,10 @@ class PolyFilter2D(Operator):
             obs.detdata[self.det_data][:] = temp_ob.detdata[self.det_data][:]
             gt.stop("Poly2D:  Copy output")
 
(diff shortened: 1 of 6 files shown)


tskisner:
Since this has no impact on the the CI unit tests (which run on only 2 processes), I will self-merge these trivial changes.

PR #538, opened 2022-02-12
Fix performance issue with detector weights

tskisner:
For extremely large focalplanes, the computation of the detector weights in a noise model was very slow.  This work uses numpy `searchsorted()` rather than boolean logic to speed things up.


tskisner:
Tested (and approved offline) by @keskitalo .  Merging now.

PR #539, opened 2022-02-15
Small changes needed when dumping to spt3g format

tskisner:


tskisner:
This fixes a couple of logic errors discovered when running HDF5 to SPT3G conversions

src/toast/spt3g/spt3g_utils.py
```diff
@@ -227,7 +223,9 @@ def compress_timestream(ts, params, rmstarget=2**10, rmsmode="white"):
         offset (float) :  The removed offset
 
     """
-    if not params:
+    if params is None:
+        return ts, 1, 0
+    if isinstance(params, bool) and not params:
```
keskitalo, line 228: seems like this test could have been `if params is None or params is False` but just as fine like this.

keskitalo (approved):

src/toast/spt3g/spt3g_utils.py
tskisner: Fixed

PR #541, opened 2022-03-04
Split mapmaker for better modularity

tskisner:
This is a work in progress, and unit tests need to be updated still.  Define new operators for template amplitude solver and the application of amplitudes to timestream.


giuspugl:
this is really great! thanks a lot @tskisner ! 
we need to make sure to perhaps firstly port   the implementation to fit the gain amplitudes from #398 .

tskisner:
This is ready for a review, but there is a unit test failure that I am still tracking down, and which does not seem to trigger an overall failure in the test runner.  Some notes:

- The "MapMaker" operator is now a wrapper around `SolveAmplitudes`, `ApplyAmplitudes(subtract)`, and `BinMap`.
- There is a separate operator, `Calibrate`, which calls `SolveAmplitudes` and `ApplyAmplitudes(divide)`
- Workflows could obviously call these low-level operators directly to do more complicated things, but having the high-level wrappers is convenient

After this is merged, we can revisit #398

giuspugl (commented):
Everything looks good to me . 
Please update the  [__init__ file](https://github.com/hpc4cmb/toast/blob/391a04c45bce870073196910938382a43cf473f5/src/toast/ops/__init__.py#L60) to import also `Calibrate`.

src/toast/ops/filterbin.py
```diff
@@ -114,7 +113,7 @@ def normalize(self, good=None):
         """Normalize templates"""
         for start, stop, template in zip(self.starts, self.stops, self.templates):
             if good is None:
-                norm = np.sum(template**2) ** 0.5
+                norm = np.sum(template ** 2) ** 0.5
```
keskitalo, line 116: This is old `black` isn't it? Or did they change the unary operator spacing back?

tskisner: Sorry, ran with a stale black, just pushed the fix.

keskitalo (approved):
Thanks a lot for this work. I know refactoring large pieces of code can be tedious but sometimes it cannot be avoided!

src/toast/noise.py
```diff
@@ -56,8 +56,18 @@ def __init__(
         if mixmatrix is None:
             # Default diagonal mixing matrix
             self._keys = self._dets
-            self._keys_for_dets = {x: [x,] for x in self._dets}
-            self._dets_for_keys = {x: [x,] for x in self._dets}
+            self._keys_for_dets = {
```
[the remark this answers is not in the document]
tskisner: This was my fault- my editor used a stale copy of `black`.  Will restore in next push.

PR #542, opened 2022-03-09
Support using the Weather class as a container

### LINKED ISSUE #472: Port override of simulated weather parameters
In #379, there was a mechanism to override the weather parameters that were randomly drawn for an observation.  We should consider how we want to handle this now.  For context, our current `SimGround` operator accepts the name or file path for a given weather model (distributions of parameters for every day / hour in the year).  Then an instance of that weather model is created for each observation and the time in the model is set to the time of the observation.  The random parameters drawn are determined by the site ID, the realization, and the time of the observation, and are reproducible.

Here are some plausible scenarios:

1.  We want to simulate data from real observations.  In that case an experiment would be loading actual pointing data and would create a new weather class (inheriting from the `Weather` base class) which just returned the real conditions during that observation.
2. We want to use the `SimGround` class to simulate the telescope scanning, but we want to set the weather ourselves.  In this case, we could have a weather class that was just a container that returned constant values for the parameters.  A workflow or experiment specific operator could go through every observation and set the weather to be some fixed values.
3. We want to use the existing `SimWeather` class, which uses MERRA-2 data to draw random conditions from the historical distributions, but we want to use our own distributions.  One can load an existing HDF5 weather file, manipulate it in python, and dump out a new file with any modifications.  Then this file can be passed to the `SimGround` operator.

It seems like we could support (1) and (2) by adding optional fixed parameters to the Weather base class, so that a weather model with constant parameters is handled by just using the base class.  @keskitalo , is there a use case where the user wants to have parameters drawn from an existing MERRA-2 data file but then wants to manually tweak a subset of those parameters?

tskisner:
For real weather measurements or simulations of specific conditions, we want to be able to fully specify the weather properties.  This minor changes makes the base class just a container to hold values.  This can then be used in workflows where the site weather properties are loaded from disk, for example.  Fixes #472

### DIFF
--- a/src/toast/tests/weather.py
+++ b/src/toast/tests/weather.py
@@ -6,6 +6,8 @@ from datetime import datetime
 
 import numpy as np
 
+import astropy.units as u
+
 from .mpi import MPITestCase
 
 import numpy.testing as nt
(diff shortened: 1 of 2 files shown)


tskisner:
This is a very minor change and all tests pass.  Self merging.

PR #544, opened 2022-03-22
Add scheduler options for rotating boresight in a restricted range

keskitalo:
This PR extends the scheduler with a new option needed in the CMB-S4 simulations.


tskisner (approved):
Looks good!

PR #550, opened 2022-04-11
Tentative fix to TEB convolution

keskitalo:
Re-interpret the TEB convolution:
* T-component requires simple intensity-only convolution
* EB-convolution mixes the input Q/U sky components in harmonic domain by mapping 

`a^E_lm -> a^E_lm cos(2psi) + a^B_lm sin(2psi)`

and

`a^B_lm -> a^B_lm cos(2psi) - a^E_lm sin(2psi)`.

The `cos(2psi)` and `sin(2psi)` terms are evaluated separately and co-added. In the absence of a HWP, the `sin(2psi)` term is zero and the second polarized convolution can be omitted.

### DIFF
--- a/src/toast/ops/conviqt.py
+++ b/src/toast/ops/conviqt.py
@@ -237,6 +237,9 @@ class SimConviqt(Operator):
         if self.detector_pointing is None:
             raise RuntimeError("detector_pointing cannot be None.")
 
+        if self.hwp_angle is not None:
+            raise RuntimeError("Standard conviqt operator cannot handle HWP angle")
+
         log = Logger.get()
 
         timer = Timer()
@@ -265,9 +268,9 @@ class SimConviqt(Operator):
 
             detector = self.get_detector(det)
 
-            theta, phi, psi, psi_pol = self.get_pointing(data, det, verbose)
-            np.savez(f'{self.det_data}_{det}.npz', psi=psi, psi_pol=psi_pol,
-                    theta=theta, phi=phi) 
+            theta, phi, psi, psi_pol, hwp_angle = self.get_pointing(data, det, verbose)
+            #np.savez(f'{self.det_data}_{det}.npz', psi=psi, psi_pol=psi_pol,
+            #        theta=theta, phi=phi)
             pnt = self.get_buffer(theta, phi, psi, det, verbose)
             del theta, phi, psi
 
@@ -397,7 +400,7 @@ class SimConviqt(Operator):
         nullquat = np.array([0, 0, 0, 1], dtype=np.float64)
         timer = Timer()
         timer.start()
-        all_theta, all_phi, all_psi, all_psi_pol = [], [], [], []
+        all_theta, all_phi, all_psi, all_psi_pol, all_hwp_angle = [], [], [], [], []
         for obs in data.obs:
             if det not in obs.local_detectors:
                 continue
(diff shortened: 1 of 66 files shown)


src/toast/tests/ops_sim_tod_conviqt.py
```diff
@@ -55,9 +55,16 @@ def setUp(self):
             self.slm = create_fake_sky_alm(self.lmax, self.fwhm_sky)
 
             hp.write_alm(self.fname_sky, self.slm, lmax=self.lmax, overwrite=True)
+
+            # Inputs for the TEB convolution
+            # FIXME : the TEB conviqt operator should do this match rather than leave it for the user
```
giuspugl, line 60: Shall we address this `FIXME` before merging it ?

src/toast/ops/conviqt.py
```diff
@@ -732,26 +747,28 @@ def _exec(self, data, detectors=None, **kwargs):
...
+            convolved_data += np.cos(4 * hwp_angle) * self.convolve(
+                skyEB, beam_P, detector, pnt, det, verbose, pol=True
             )
             del (pnt,)
-            # B-convolution
+            # BE-convolution
             pnt = self.get_buffer(theta, phi, psi, det, verbose)
-            convolved_data += np.sin(2 * psi_pol) * self.convolve(
-                skyB, beam_P, detector, pnt, det, verbose
+            convolved_data += np.sin(4 * hwp_angle) * self.convolve(
```
giuspugl, line 768: It is not clear  to me why now the convolved data are weighted by the sin(or cosine) of the hwp angle and not by the psi_pol angle..

keskitalo: The psi_pol angle is already included in the beam and pointing matrix

src/toast/tests/ops_sim_tod_conviqt.py
```diff
@@ -574,154 +574,199 @@ def test_sim_TEB_conviqt(self):
...
+            self.comm, obs_time=120 * u.min, pixel_per_process=2, hwp_rpm=None,
         )
-        self.make_beam_file_dict(data)
-        
-        
+        data_w_hwp = create_satellite_data(
+            self.comm, obs_time=120 * u.min, pixel_per_process=2, hwp_rpm=100,
+        )
+        self.make_beam_file_dict(data_wo_hwp)
+        self.make_beam_file_dict(data_w_hwp)
```
giuspugl, line 591: Thanks a lot @keskitalo ! i think we're  ready to merge this PR. 
Just a minor comment. 
I don't think this extra-call  `make_beam_file_dict` is needed. If look at its   definition in https://github.com/hpc4cmb/toast/blob/daa4d7e06c3a1f4e949dc6fab71ae43241ebea8e/src/toast/tests/ops_sim_tod_conviqt.py#L215 
essentially what it does is to override the dictionary `beam_file_dict`. Perhaps i can work out a fix to this with something like: 

```
def make_beam_file_dict(self, data): 

        fname2 = self.fname_beam.replace(".fits", "_bottom.fits") 
        beam_file_dict = {}
        for det in data.obs[0].local_detectors:
            if det[-1] == "A":
                beam_file_dict[det] = self.fname_beam
            else:
                beam_file_dict[det] = fname2

        return beam_file_dict 

``` 
please let me know if that makes the test routine cleaner .

keskitalo:
I did a full source format. that should make the merging easier.

keskitalo:
Feel free to merge into your branch. It is probably time to request a review and merge to toast3

PR #551, opened 2022-04-22
Upgdate script to use the new API

keskitalo:


tskisner (approved):
Thanks for fixing this!

PR #552, opened 2022-04-22
Add support to HWP in beam convolution

keskitalo:


keskitalo:
Fix the divergent history of #534

cmake.sh
```diff
@@ -0,0 +1 @@
+cmake .. -DCMAKE_INSTALL_PREFIX=$PREFIX
```
tskisner, line 1: Was this script accidentally committed?  If you have a specific set of cmake options needed for a system, feel free to add a script to the `platforms` directory.  We could consider adding a `default.sh` script or similar.

src/toast/ops/conviqt.py
```diff
@@ -266,6 +266,8 @@ def _exec(self, data, detectors=None, **kwargs):
             detector = self.get_detector(det)
 
             theta, phi, psi, psi_pol = self.get_pointing(data, det, verbose)
+            np.savez(f'{self.det_data}_{det}.npz', psi=psi, psi_pol=psi_pol,
+                    theta=theta, phi=phi)
```
tskisner, line 270: Could we guard this debug dumping by a line like:
```
# (assuming that Environment was imported at the top of the file)
env = Environment.get()
if env.log_level == "VERBOSE":
    np.savez(...)
```
or use some other internal `verbose` settting in the class.

src/toast/tests/ops_sim_tod_conviqt.py
```diff
@@ -96,19 +128,111 @@ def setUp(self):
...
+                self.beam_file_dict[det] = fname2
+
+        return
+
+    def test_sim_conviqt(self):
         if not ops.conviqt.available():
             print("libconviqt not available, skipping tests")
             return
 
         # Create a fake scan strategy that hits every pixel once.
```
tskisner, line 230: Remove stale comment

src/toast/tests/ops_sim_tod_conviqt.py
```diff
@@ -202,99 +320,263 @@ def test_sim(self):
...
-            norm = np.mean(cl_out[compare] / ref)
-            if not np.allclose(
-                norm * ref,
-                cl_out[compare],
-                rtol=1e-5,
-                atol=1e-5,
-            ):
-                fail = True
+        # Create a fake scan strategy that hits every pixel once.
+        #        data = create_healpix_ring_satellite(self.comm, nside=self.nside)
```
tskisner, line 344: You can remove this, and the comment

tskisner (changes requested):
Just a couple of small tweaks, but otherwise looks good

cmake.sh
keskitalo: Oops! Removed the unnecessary file.

src/toast/ops/conviqt.py
keskitalo: That is a leftover debugging statement. Now removed.

src/toast/tests/ops_sim_tod_conviqt.py
keskitalo: Fixed.

keskitalo:
Addressed review comments and will merge once the unit tests finish.

PR #553, opened 2022-04-26
Ground sim tweaks

keskitalo:
Two small tweaks:
- timestamps from `SimGround` now end before the commanded end time.  This way two consecutive observations avoid having one sample overlaps
- user can instruct the operator to begin the constant elevation scan at a randomized phase


tskisner (approved):
Looks good, this is a nice feature.

PR #556, opened 2022-04-28
Add the missing TEB convolution commits

keskitalo:
Adding some missing commits that were left out from the original "merge" branch.


tskisner (approved):
Thanks to @giuspugl for catching the missing changes in the previous PR!

PR #560, opened 2022-05-02
Flag and import cleanup

tskisner:
This changes a few operators to consistently use the default names for flags and the default bit mask.  Most operators were already changed to use these.  This PR also includes running the `isort` tool on the python source files, which organizes the import statements in a way that agrees with PEP8.


keskitalo:
Looks neat! Merge away at will.

PR #563, opened 2022-05-04
Debug wheel builds

tskisner:
Work in progress debugging previous failures.


tskisner:
Self-merging this, since it only impacts the wheel building scripts and not any actual code.

PR #564, opened 2022-05-04
Introduce a small helper class for tracking observing sessions.

tskisner:
This small addition provides a way to associate observations from individual wafers / frequencies on the same telescope for the same span of time.


tskisner:
The design of this `Session` class was discussed with @keskitalo a couple days ago.  Its use is backwards compatible with current workflows, and simply provides a future mechanism for grouping observations that are from the same telescope and time period.

PR #567, opened 2022-05-17
Add basic support for WCS pixelization.

tskisner:
* Add a new PixelsWCS operator

* Add I/O routines

* Add unit tests for a couple of projections

Currently this work adds a dependency on pixell, which is used
only to build the WCS parameters.  This is something we could
implement internally if needed.  For example, if we want to
support more architectures than are currently supported by
pixell wheels.


tskisner:
The unit test failures on MacOS are reproducible (a segfault inside the lapack routines of the Apple accelerate framework), but I have not tracked down what is triggering them.  In addition to the standard WCS projections added to this PR, the pixels class now supports specifying a shared data object containing the variable Lon/Lat center of the projection.  This allows us to make maps in source-centered coordinates.  As an example, here are some plots from the unit tests showing the same data with a moving point source.  One map was made in fixed RA/DEC coordinates and the other specified the fake source ephemeris as the center offset:
source_CAR_notrack_map fits_I
source_CAR_map fits_I

src/toast/ops/mapmaker.py
```diff
@@ -423,9 +424,14 @@ def _exec(self, data, detectors=None, **kwargs):
                 ]
             ).apply(data)
 
-        # FIXME:  This all assumes the pointing operator is an instance of the
-        # PointingHealpix class.  We need to generalize distributed pixel data
-        # formats and associate them with the pointing operator.
+        # FIXME:  This I/O technique assumes "known" types of pixel representations.
```
keskitalo, line 427: Did you create an issue for this?

src/toast/ops/noise_estimation.py
```diff
@@ -128,7 +128,8 @@ class NoiseEstim(Operator):
     )
 
     output_dir = Unicode(
-        ".",
+        None,
```
keskitalo, line 131: Noise estimation code in the WCS PR?

src/toast/ops/pixels_wcs.py
```diff
@@ -0,0 +1,577 @@
...
+            self._done_auto = True
+
+    @traitlets.observe("center_offset")
+    def _reset_auto_bounds(self, change):
+        # Track whether we need to recompute the bounds.
+        if change["new"] is not None:
+            if self.auto_bounds:
+                self._done_auto = False
+
+    @traitlets.observe("projection", "center", "bounds", "dimensions", "resolution")
```
keskitalo, line 174: These are neat!

src/toast/ops/pixels_wcs.py
```diff
@@ -0,0 +1,577 @@
...
+                        # print(f"det {det} has {np.sum(good)} good pixels")
+                        # print((ob.detdata[self.pixels][det][vslice])[good])
+                        self._local_submaps[
+                            (ob.detdata[self.pixels][det][vslice])[good]
+                            // self._n_pix_submap
+                        ] = 1
+
+    @function_timer
+    def _get_scan_range(self, obs):
+        # FIXME: mostly copied from the atmosphere simulation code- we should
```
keskitalo, line 474: I agree. Perhaps given an observation, the `Focalplane` could tell it's range?

src/toast/ops/scan_wcs.py
```diff
@@ -0,0 +1,344 @@
...
+from ..traits import Bool, Instance, Int, Unicode, trait_docs
+from ..utils import Logger
+from .operator import Operator
+from .pipeline import Pipeline
+from .pointing import BuildPixelDistribution
+from .scan_map import ScanMap, ScanMask
+
+
+@trait_docs
+class ScanWCSMap(Operator):
```
keskitalo, line 21: Is there no way to write a `ScanMap` operator that works on any `Pixels` class?

keskitalo (commented):
Excellent. This makes a great addition to TOAST portfolio of capabilities.

keskitalo (approved):

src/toast/ops/noise_estimation.py
tskisner: These are residual small things that were missed in #571 .  Apologies to slip them in here.

src/toast/ops/mapmaker.py
tskisner: Done.

src/toast/ops/pixels_wcs.py
tskisner: Now tracked in issue #573

src/toast/ops/scan_wcs.py
tskisner: I agree we should generalize this.  Captured now in #574.

PR #569, opened 2022-06-03
Several small fixes to the atmosphere simulation:

tskisner:
* Allow turnaround intervals to be None, in the case of telescopes
  with scan patterns that do not have well defined turnarounds.

* Dumping of debug TODs of atmosphere was broken.  Move this
  functionality into the observing operator and dump TOD snapshots
  at intermediate points in the process of calibration, loading,
  etc.  This helps identify problems in specific steps.


tskisner:
This code does not change results for our existing tests with simulated ground pointing that has turnarounds.  The changes to debug dumping work locally (previously they were completely broken since the toast3 port).  Merging this.

PR #571, opened 2022-06-06
Improvements to noise estimation

tskisner:
* Noise estimation operator now optionally creates a toast.Noise
  model directly, in addition to writing the existing FITS dump

* Add a new operator which can fit an existing noise model to
  1/f spectrum parameters.  This allows creation of "smooth"
  Fourier domain kernels for filtering from noisy estimates.

### DIFF
--- a/src/toast/ops/noise_estimation.py
+++ b/src/toast/ops/noise_estimation.py
@@ -18,6 +18,7 @@ from .. import qarray as qa
 from .._libtoast import filter_poly2D, filter_polynomial, subtract_mean, sum_detectors
 from ..intervals import interval_dtype
 from ..mpi import MPI, Comm, MPI_Comm, use_mpi
+from ..noise import Noise
 from ..observation import default_values as defaults
 from ..timing import function_timer
 from ..traits import (
@@ -120,6 +121,10 @@ class NoiseEstim(Operator):
 
     shared_flag_mask = Int(
         defaults.shared_mask_invalid, help="Bit mask value for optional shared flagging"
+    )
+
+    out_model = Unicode(
+        None, allow_none=True, help="Create a new noise model with this name"
     )
 
     output_dir = Unicode(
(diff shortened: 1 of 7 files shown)


tskisner:
This is now ready for review.  The unit tests now simulate noisy data, estimate the PSD for each detector and each observation, and then use the new `FitNoiseModel` operator to do a nonlinear least squares fit to the usual 1/f spectrum parameters.  This fit is weighted toward higher frequency in order to force the NET to have a good fit.  Here is an example from the plots made by the unit tests:
estimate_model_small_patch-0-0_D0B

src/toast/ops/noise_estimation.py
```diff
@@ -236,14 +241,14 @@ def _redistribute(self, data, obs):
                 intervals=dup_intervals,
             )
             log.debug_rank(
-                f"{data.comm.group:4} : Duplicated observation in",
+                f"{obs.comm.group:4} : Duplicated observation in",
```
keskitalo, line 244: Thanks for catching these.

src/toast/ops/noise_estimation.py
```diff
@@ -289,7 +294,7 @@ def _exec(self, data, detectors=None, **kwargs):
 
         log = Logger.get()
 
-        if self.focalplane_key:
+        if self.focalplane_key is not None:
```
keskitalo, line 297: That is cleaner, thanks.

src/toast/ops/noise_estimation.py
```diff
@@ -654,7 +685,7 @@ def discard_outliers(self, binfreq, all_psds, all_times, all_cov):
                         del all_cov[ii]
 
             if nbad > 0:
-                log.info(f"Masked extra {nbad} psds due to outliers.")
+                log.debug(f"Masked extra {nbad} psds due to outliers.")
         return all_psds, all_times, nempty + nbad, all_cov
```
keskitalo, line 689: That is a better verbosity level for that information.

src/toast/tests/ops_noise_estim.py
```diff
@@ -110,8 +110,10 @@ def test_noise_estim(self):
         estim.apply(data)
```
keskitalo, line 110: We had a unit test, why did it not catch the wrong communicators being used?

src/toast/ops/noise_model.py
```diff
@@ -54,3 +56,154 @@ def _requires(self):
...
+        # We weight the residual so that the high-frequency values specifying
+        # the white noise plateau / NET are more important.  Also the lowest
+        # estimated bin is usually garbage.  We arbitrarily choose a weighting
+        # of:
+        #
+        #  W = f_nyquist - (f_nyquist / (1 + f^2))
+        #
+        weights = np.ones_like(current) * freqs[-1]
+        weights -= freqs[-1] / (1.0 + freqs**2)
+        resid = np.multiply(weights, current - data)
```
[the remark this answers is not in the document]
tskisner: They are equivalent in this case of 1D arrays.  I guess it was just habit from multiplying higher dimensional arrays.  I can change it if you think that is more clear...

src/toast/tests/ops_noise_estim.py
tskisner: The problems would only manifest if there were multiple groups working on multiple observations.  This may not be the case in the original data you used to develop the code.  The unit tests on github CI use 2 groups each with one process.  Running with 4 processes (2 groups of 2) made it fail more reliably.  This is a larger problem, for example the SimConviqt unit tests consistently deadlock when run on 4 processes.  We need automated testing a larger system (as you know we have been discussing that in other contexts).

src/toast/ops/noise_model.py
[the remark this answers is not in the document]
tskisner: Resolved offline.

PR #572, opened 2022-06-10
Add extra logic to avoid unnecessary communicator splits

tskisner:
This work attempts to minimize MPI communicator operations that may be fragile on some systems.


tskisner:
Although these minor changes did not help the original problem, they are a good cleanup of the communicator usage and do not impact anything else.

PR #575, opened 2022-06-13
Test wheels again.

tskisner:


tskisner:
This work did not resolve the actual wheel builds, but it does include scripts for setting up a conda development environment, which is useful and needed.  The actual wheel builds will need to follow the same techniques as numpy and scipy in terms of linking to openblas.  Merging this now to get access to the conda scripts

PR #576, opened 2022-06-16
Small improvements to noise fitting

tskisner:
* Add unit tests for extreme noise cases
* Make fitting more robust


tskisner:
Since these are just expanded tests and fixes for corner cases, self-merging.

PR #577, opened 2022-06-18
Work in progress on session-based atmosphere simulations

tskisner:
This changes the atmosphere simulations to generate a single realization per observing session, which might have multiple individual observations looking at the same sky.  Not finished yet.


keskitalo (approved):
Awesome! The refactor of atmospheric simulation operator was well overdue.

PR #579, opened 2022-06-29
Several small fixes discovered during testing of SSO sims and #577

tskisner:
* Fix HDF5 data shape parameter to work with h5py < 3

* Handle error of data split with no session name

* Add equality operation for GroundSite and Weather

* Minor fixes to use of sessions in satellite and ground
  simulations

* Small cleanups to config handling, and fix parsing of
  containers of quantities.

* Ensure that weather simulation time is loaded in UTC from hdf5


AERAdler (approved):
This is fine.

PR #580, opened 2022-07-07
Test building wheels with homebrew gcc on macos

tskisner:
As we move away from the Accelerate framework on MacOS (which have some severe bugs and is no longer updated much), attempt to switch toolchains to homebrew gcc.  This has the added benefit of support OpenMP as well.


tskisner:
This work only impacts the wheel deployment scripts (and the removal of a temporary copy of unported toast-2 sources).  Merging now in preparation for a pre-release tag.

PR #582, opened 2022-08-09
Aoa tweaks

keskitalo:
Tweaks coming from CMB-S4 AoA work.

* Fixes a bug in the scheduler that intermittently caused it to add mixed rising/setting scans when the patch was observed close to Az=180 degrees.
* Add a command line tool for combining observation matrices. It is just an interface around existing Python facilities.
* Fix a bug in caching elements of the observation matrix to speed up future calculations
* Re-enable observation matrix unit tests that were disabled during recent refactor work
* Fix the detector noise weights to be defined as 1 / sigma^2, where sigma^2 is the noise variance per detector sample. Without this, the noise matrices generated by the mapmaker cannot be translated into depth maps.

### DIFF
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -430,10 +430,6 @@ class Focalplane(object):
             etc.  Will be calculated from the detector offsets by default.
         sample_rate (Quantity):  The common (nominal) sample rate for all detectors.
         thinfp (int):  Only sample the detectors in the file.
-        fmin (float):  Override noise model fmin.
-        fknee (float):  Override noise model fknee.
-        alpha (float):  Override noise model alpha.
-        NET (float):  Override noise model NET.
 
     """
 
@@ -446,18 +442,10 @@ class Focalplane(object):
         field_of_view=None,
         sample_rate=None,
         thinfp=None,
-        fmin=None,
-        fknee=None,
-        alpha=None,
-        NET=None,
     ):
         self.detector_data = detector_data
         self.field_of_view = field_of_view
         self.sample_rate = sample_rate
-        self.fmin = fmin
-        self.fknee = fknee
-        self.alpha = alpha
-        self.NET = NET
         self.thinfp = thinfp
         if detector_data is not None and len(detector_data) > 0:
             # We have some dets
@@ -466,20 +454,6 @@ class Focalplane(object):
     @function_timer
     def _initialize(self):
         log = Logger.get()
-
-        self.detector_data = self.detector_data.copy()
-        if self.fmin is not None:
-            self.detector_data["psd_fmin"] = self.fmin
-        if self.fknee is not None:
-            self.detector_data["psd_fknee"] = self.fknee
-            # Noise simulation has a requirement fmin < fknee
-            bad = self.detector_data["psd_fknee"] <= self.detector_data["psd_fmin"]
-            self.detector_data["psd_fmin"][bad] \
-                = self.detector_data["psd_fknee"][bad] / 10
-        if self.alpha is not None:
-            self.detector_data["psd_alpha"] = self.alpha
-        if self.NET is not None:
-            self.detector_data["psd_net"] = self.NET
 
         if self.thinfp is not None:
             # Pick only every `thinfp` pixel on the focal plane
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -430,6 +430,10 @@ class Focalplane(object):
             etc.  Will be calculated from the detector offsets by default.
         sample_rate (Quantity):  The common (nominal) sample rate for all detectors.
         thinfp (int):  Only sample the detectors in the file.
+        fmin (float):  Override noise model fmin.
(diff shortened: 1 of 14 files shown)


workflows/toast_sim_ground.py
```diff
@@ -104,6 +109,21 @@ def parse_config(operators, templates, comm):
         help="Only sample the provided focalplane pixels",
     )
 
+    parser.add_argument(
+        "--fknee",
```
tskisner, line 113: Would it not be better to generate a focalplane file that has this value?  The [included script](https://github.com/hpc4cmb/toast/blob/7a0a7b29d5f0e886f7ed75a389c43a09bd230f70/src/toast/scripts/toast_fake_focalplane.py#L85) already does that.  Otherwise it seems like we risk adding more and more options here to override specific things in the inputs, rather than just changing the inputs.

For example, we already have other options (`--thinfp` and `--sample_rate`) which are also part of the focalplane and should arguably be removed (i.e. just use a different focalplane).  However, I realize that might break existing scripts.

workflows/toast_sim_ground.py
```diff
@@ -104,6 +109,21 @@ def parse_config(operators, templates, comm):
...
+        required=False,
+        type=float,
+        help="Override noise model knee frequency [Hz]",
+    )
+
+    parser.add_argument(
+        "--pwv_limit",
```
tskisner, line 120: Is it better to have this option at the workflow level, or add it as a trait to the `SimGround` operator?  For example, we could do a [continue in the code here](https://github.com/hpc4cmb/toast/blob/toast3/src/toast/ops/sim_ground.py#L717) if the pwv is too high and avoid creating the observation in the first place, rather than creating it and then deleting it.

workflows/toast_sim_ground.py
```diff
@@ -147,26 +167,33 @@ def load_instrument_and_schedule(args, comm):
 
     fname_pickle = (
         f"{os.path.basename(args.focalplane)}_"
-        f"thinfp={args.thinfp}_fsample={args.sample_rate}.pck"
+        f"thinfp={args.thinfp}_fsample={args.sample_rate}_fknee={args.fknee}.pck"
```
tskisner, line 170: Why not just pass a different file name as the `--focalplane` option?

tskisner (changes requested):
I feel strongly that we should keep simulation-specific constructor options out of the `Focalplane` base class (obviously those and any other parameters can just be passed in with the detector properties table).

For the extra commandline options to `toast_sim_ground.py`, I feel less strongly.  However, I still think they might be better handled by moving those options out of the workflow and into higher level scripting or into operator traits.  For example, it seems like the commandline options should specify the inputs (focalplane, schedule), workflow options (output directory, mapmaking "flow" like per-obs maps, other options that affect the order / flow of operations) and the other options could be implemented by passing in a different input file or by changing operator traits from parameter file / commandline.

src/toast/instrument.py
```diff
@@ -430,6 +430,10 @@ class Focalplane(object):
             etc.  Will be calculated from the detector offsets by default.
         sample_rate (Quantity):  The common (nominal) sample rate for all detectors.
         thinfp (int):  Only sample the detectors in the file.
+        fmin (float):  Override noise model fmin.
+        fknee (float):  Override noise model fknee.
+        alpha (float):  Override noise model alpha.
+        NET (float):  Override noise model NET.
```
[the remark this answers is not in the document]
keskitalo: I agree, that would have been more elegant.  I reverted these changes.

workflows/toast_sim_ground.py
keskitalo: I removed `--fknee` but I feel both `--thinfp` and `sample_rate` have their place. They allow rapid descoping of the simulation for testing purposes.

tskisner: Sounds good, thanks for the consideration.

workflows/toast_sim_ground.py
keskitalo: This was a quick and dirty solution that was needed for the present simulation task.  What we need is some way of identifying the observations to cut before they are distributed to process groups.  Otherwise we run the risk of severe load imbalance.  It also seems that generating a process group without any observations causes the workflow to fail.

The current hack has the benefit of being limited to the ground simulation workflow.  Until we have a proper solution to the problem, I would keep it there.  How about I edit the help string for `--pwv_limit` to reflect this thinking and create an issue to address the full problem at a later date?

workflows/toast_sim_ground.py
keskitalo: Agreed.

workflows/toast_sim_ground.py
tskisner: Ok, it is fine to leave it as-is.  I agree that ideally we would need to have the `SimGround` operator go and instantiate all the weather instances first, prior to distribution among groups.  That is possible, but more work than just adding a trait.

tskisner (approved):
Thanks for the fixes

PR #583, opened 2022-08-15
Pointing model updates

tskisner:
The generic simulated focalplane functions have poorly documented conventions, and this leads to confusion when building toast compatible focalplane models.  This work in progress first lays out the simulated focalplanes in the Xi / Eta / Gamma 2D conventions adopted by Simons Observatory, which is convenient for looking at focalplaned projected on the sky from the observer.  It then converts that model to individual detector quaternion rotations from the boresight, where the detector X axis is defined to be the polarization sensitive direction.  Still some changes needed here, and will add docs to a markdown file that will integrate into the updated jupyter book docs.

Examples of plots in both coordinate systems (we would typically be scanning parallel to the Y / Xi axis in this case):
fake_hex_xieta
fake_hex_xyz

### DIFF
--- a/src/toast/instrument_sim.py
+++ b/src/toast/instrument_sim.py
@@ -8,37 +8,8 @@ from astropy.table import Column, QTable
 
 from . import qarray as qa
 from .instrument import Focalplane
+from .instrument_coords import quat_to_xieta, xieta_to_quat
 from .vis import set_matplotlib_backend
-
-
-def cartesian_to_quat(offsets):
-    """Convert cartesian angle offsets and rotation into quaternions.
-
-    Focalplane geometries are often described in terms of wafer locations or
-    separations given in simple X/Y angle offsets from a center point.
-    this helper function converts such parameters into a quaternion describing
-    the rotation.
-
-    Args:
-        offsets (list of arrays):  each item of the list has 3 elements for
-            the X / Y angle offsets in degrees and the rotation in degrees
-            about the Z axis.
-
-    Returns:
-        (list): list of quaternions for each item in the input list.
-
-    """
-    centers = list()
-    zaxis = np.array([0, 0, 1], dtype=np.float64)
-    for off in offsets:
-        angrot = qa.rotation(zaxis, off[2] * np.pi / 180.0)
-        wx = off[0] * np.pi / 180.0
-        wy = off[1] * np.pi / 180.0
-        wz = np.sqrt(1.0 - (wx * wx + wy * wy))
-        wdir = np.array([wx, wy, wz])
-        posrot = qa.from_vectors(zaxis, wdir)
-        centers.append(qa.mult(posrot, angrot))
-    return centers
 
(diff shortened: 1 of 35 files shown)


tskisner:
The unit tests for the `SimGround` and `SimSatellite` operators now generate plots of the boresight and detector pointing / orientation on the sky.  This is a good visual check that the focalplane coordinate frame is consistent with expectations.
pointing_ground
pointing_satellite

tskisner:
This has been rebased and is ready for review.  However, we should not merge until we finish testing against external downstream packages, which is in progress.

src/toast/instrument_sim.py
```diff
@@ -459,22 +465,18 @@ def rhombus_layout(
...
-    # convert pol vector to radians
-    pol = polang * np.pi / 180.0
+    pixdiam = angwidth.to_value(u.radian) / dim
 
     # number of digits for pixel indexing
 
     ndigit = 0
-    test = npix
+    test = npos
     while test > 0:
```
keskitalo, line 480: This could be `ndigit = int(np.log10(np.abs(test))) + 1`

src/toast/qarray.py
```diff
@@ -412,6 +414,122 @@ def from_vectors(v1, v2):
...
+            or (object_ndim(psi) == 1)
+        ):
+            return out.array().reshape((1, 4))
+        else:
+            return out.array()
+    else:
+        return out.array().reshape((-1, 4))
+
+
+def to_iso(q):
```
keskitalo, line 483: I remember using `to_angles()` a few times. Did you search and replace all instances of the method?

keskitalo (approved):
Looks great! Thanks for the tedious work.

src/toast/qarray.py
tskisner: I believe so- although I also fixed that in the separately in the sotodlib toast3 branch

src/toast/instrument_sim.py
tskisner: Implemented for both hex and rhombus layouts, with a slight variation.

src/toast/ops/conviqt.py
```diff
@@ -437,7 +437,7 @@ def get_pointing(self, data, det, verbose):
                     if verbose:
                         timer.report_clear(f"initialize flags for detector {det}")
 
-                theta, phi, psi = qa.to_angles(quats)
+                theta, phi, psi = qa.to_iso(quats)
```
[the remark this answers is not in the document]
tskisner: This is now changed.

src/toast/qarray.py
```diff
@@ -412,6 +414,122 @@ def from_vectors(v1, v2):
         return out.array().reshape((-1, 4))
 
 
+def from_iso(theta, phi, psi):
```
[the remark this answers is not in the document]
tskisner: Implemented now.

PR #584, opened 2022-08-16
Add small sandbox test for OMP offload in a python extension.

tskisner:
This adds a small standalone test of making a compiled extension that includes openmp target offload code.  Tested with the `gcc-12-offload-nvptx` compiler package from the `ubuntu-toolchain-r` repo.  This was combined with a minimal conda env that had libomp replaced with the libgomp shipped with the above compiler.  Obviously not a long term solution.  This is just to confirm the correct syntax of the omp pragmas.  One strange thing is that this code seems to only work passing the device pointers (found with manual lookup) to the `target is_device_ptr()` clause, rather than using the internal OMP memory mapping table and passing the host pointers on entering the data region with `target data use_device_ptr()`.  Still investigating why, but this demonstrates that it is possible (at least with this compiler) to have OpenMP target offload code inside a dynamically loaded python extension.

### DIFF
--- a/etc/pyomptarget/Makefile
+++ b/etc/pyomptarget/Makefile
@@ -1,0 +1,27 @@
+
+# GCC 12.1.0 with offload support
+CXX := g++-12
+CXXFLAGS := -O0 -g -fPIC -pthread -std=c++11 -fcf-protection=none -fno-stack-protector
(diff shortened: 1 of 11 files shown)


nestordemeure (approved):
Looks good!

Issue #581, opened 2022-07-20
`pixels_healpix()` does not work with `shared_flags=None`


keskitalo:
In the `toast3` branch, Healpix pixel numbers are expanded with operator `PixelsHealpix`. It in turn calls a `pybind11` wrapper called `pixels_healpix`.

`pixels_healpix` expects a `py::buffer` argument called `shared_flags`. `PixelsHealpix` will pass `None` for that argument, unless `self.detector_pointing.shared_flags` is set. This causes the call to fail.

keskitalo:
Fixed in #586

PR #586, opened 2022-09-08
simple fix to missing flags

keskitalo:
Addresses #581


tskisner (approved):
Thanks for catching and fixing this place where the flags-are-none situation was not handled.

PR #587, opened 2022-09-12
Fix two harmless warnings

keskitalo:
Ground workflow currently treats all warnings as errors. This PR fixes harmless warnings that can prevent the workflow from completing.


keskitalo:
Self-merging after discussing with @tskisner .

PR #588, opened 2022-09-16
SPT-3G export tune up

cnweaver:
This set of changes encompasses my fixes to make data exported by TOAST readable by spt3g_software to a point such that a skymap can be produced from simulated detector data. As my focus was on the export case, some new functionality isn't mirrored for importing, although some of the more fundamental parts are. 

The main changes are:
- Exporting focal plane geometry using SPT-3G datastructures
- Setting units on the detector data and correcting how units are translated
- Accounting for a coordinate system convention difference when translating directions
- Splitting data into `G3Frame`s in a manner more typical for experimental data

There is also some optimization of `IntervalList` intersections, as this was a bottleneck for exporting data.

### DIFF
--- a/src/toast/intervals.py
+++ b/src/toast/intervals.py
@@ -262,17 +262,38 @@ class IntervalList(Sequence, AcceleratorObject):
         result = list()
         curself = 0
         curother = 0
+        lenself = len(self.data)
+        lenother = len(other)
+
+        # Find the first entry in ivals whose last is not smaller than val,
+        # searching within the range of indices [ileft,iright)
+        def lbound(ivals, val, ileft, iright):
+            # represents the width of the remaining interval
+            count = iright-ileft
+            while count>0:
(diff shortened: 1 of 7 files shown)


tskisner:
Thanks @cnweaver , will check this out soon.

cnweaver:
Hm, it looks like there might be something fishy about the change in [b1a3e6c](https://github.com/hpc4cmb/toast/pull/588/commits/b1a3e6c134a90bd05396dc3d3038eb98665ac8fe). I'll look into it as soon as I can.

tskisner:
Just a note, I am working on a PR that systematically enforces consistent units for timestreams across all simulation operators, noise estimation, etc.  Should only impact a few lines of this PR, but wanted to give you a heads up.

cnweaver:
Okay, I'll be happy to rebase/sync merge/whatever to work with that.

PR #594, opened 2022-10-06
Perturb HWP angle

keskitalo:
Add an operator to perturb an ideally rotating HWP


tskisner (approved):
Thanks for this useful operator!

PR #595, opened 2022-10-11
Simplify communication in PerturbHWP

tskisner:
This small change just computes the perturbation on the first process row and sets the shared memory.


keskitalo (approved):
Neat!

PR #597, opened 2022-10-21
Add a scheduler parameter for disabling SSO avoidance below horizon

keskitalo:
This PR adds scheduler options to disable Solar and Lunar avoidance when the avoided object is below a user-specified elevation.


tskisner (approved):
Looks good, thanks!

PR #599, opened 2022-10-25
Make sure the default view to detector pointing is restored

keskitalo:


tskisner (approved):
Thanks for spotting my error in PR #593 !

PR #602, opened 2022-10-28
Use the correct azimuth branch

keskitalo:
Recent changes to TOAST facilities changed the branch of the Azimuth angle being computed in the atmospheric simulation operator. This PR makes sure that the operator always sees the same branch.


tskisner (approved):
Looks good, thanks!

PR #604, opened 2022-11-02
Address bugs discovered in toast3_gpu branch.

tskisner:
* Fix typos

* Make AmplitudesMap inherit from AcceleratorObject

* Observations should also stage any AccleratorObjects in their metadata.


nestordemeure (approved):
seems good! merging, will test later today

PR #605, opened 2022-11-04
Fixing the unit_conversion routine

giuspugl:
This PR will fix the function 
https://github.com/hpc4cmb/toast/blob/70effe3c18b0275c819be4f0b17211b2352d2ed8/src/toast/utils.py#L754 
the proposed solution should properly work.


tskisner (approved):
Thanks for fixing that silly mistake!

PR #607, opened 2022-11-10
Add WCS pointing support to ground simulation example

tskisner:
* Add 3 possible pixelization schemes to the workflow, with a helper function to select between them.

* Cleanup a few places where the pixel distribution was created.

### DIFF
--- a/src/toast/scripts/toast_fake_focalplane.py
+++ b/src/toast/scripts/toast_fake_focalplane.py
@@ -190,7 +190,7 @@ def main():
     # Guard against being called with multiple processes
     mpiworld, procs, rank = get_world()
     if rank == 0:
-        with H5File(args.out, "w", comm=world, force_serial=True) as f:
+        with H5File(args.out, "w", comm=mpiworld, force_serial=True) as f:
             fp.save_hdf5(f.handle)
         plotfile = "{}.pdf".format(args.out)
         plot_focalplane(
--- a/workflows/toast_sim_ground.py
+++ b/workflows/toast_sim_ground.py
@@ -226,6 +226,85 @@ def use_full_pointing(job):
     return full_pointing
 
 
+def select_pointing(job, args):
+    """Select the pixelization scheme for both the solver and final binning."""
+
+    ops = job.operators
+
+    n_enabled_solve = np.sum(
+        [
+            ops.pixels_wcs_azel.enabled,
+            ops.pixels_wcs_radec.enabled,
+            ops.pixels_healpix_radec.enabled,
+        ]
+    )
+    if n_enabled_solve != 1:
+        raise RuntimeError(
+            "Only one pixelization operator should be enabled for the solver."
+        )
+
+    n_enabled_final = np.sum(
+        [
+            ops.pixels_wcs_azel_final.enabled,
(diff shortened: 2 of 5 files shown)


tskisner:
In order to test these changes, I did the simple exercise of creating a focalplane:
```
toast_fake_focalplane --min_pix 7 --fov_deg 5
```
And then created a schedule:
```
toast_ground_schedule \
    --site-name "Atacama" \
    --telescope "LAT" \
    --site-lon "-67.8" \
    --site-lat "-23.0" \
    --site-alt 5200 \
    --patch "small_patch,1,40,-40,44,-44" \
    --start "2020-01-01 00:00:00" \
    --stop "2020-01-01 06:00:00" \
    --out schedule.txt
```
And then tested the workflow with all defaults:
```
toast_sim_ground.py \
    --focalplane focalplane.h5 \
    --schedule schedule.txt
```
I then also tested using the WCS pixelization with all defaults:
```
toast_sim_ground.py \
    --focalplane focalplane.h5 \
    --schedule schedule.txt \
    --pixels_healpix_radec.disable \
    --pixels_wcs_radec.enable
```

src/toast/scripts/toast_fake_focalplane.py
```diff
@@ -190,7 +190,7 @@ def main():
     # Guard against being called with multiple processes
     mpiworld, procs, rank = get_world()
     if rank == 0:
-        with H5File(args.out, "w", comm=world, force_serial=True) as f:
+        with H5File(args.out, "w", comm=mpiworld, force_serial=True) as f:
```
keskitalo, line 193: How did this work before? I know I have run the fake focalplane script fairly recently.

workflows/toast_sim_ground.py
```diff
@@ -226,6 +226,85 @@ def use_full_pointing(job):
     return full_pointing
 
 
+def select_pointing(job, args):
```
keskitalo, line 229: This seems like a lot of code to duplicate in all workflows that support multiple pixelizations.  Fine for now but we should think of how to avoid the duplication.

src/toast/scripts/toast_fake_focalplane.py
tskisner: I'm not sure.  This change was made 7 months ago:  https://github.com/hpc4cmb/toast/pull/551/files and it is possible that I have not run that particular script since then.

workflows/toast_sim_ground.py
tskisner: If there is truly code duplication, then I agree.  I can imagine that individual projects would put these high-level, common workflow functions into their own packages (for example, `sotodlib`).  If we have more than one workflow within toast itself using identical code then we could move it into a subpackage.  However, individual projects will likely need their own sorts of helper functions like this.  For example, an instrument on the LMT is not going to use healpix pointing for small patch observations and would not have that in their workflows at all.  Similarly individual experiments working with real data will likely have their own operators for detector pointing models and stokes weights that deal with more complicated pointing reconstruction, HWP customization, etc.  So building a generic version of high-level workflow helper functions is probably specific to an experiment (at least for real data).

PR #609, opened 2022-11-15
Documentation overhaul

tskisner:
Initial draft of new mkdocs documentation:

- Organize into tutorials, cookbook, design discussion, API reference.

- Implement API reference

- Add extensive pointing model document to Design section

- Begin merging / updating old tutorials

- Copy unit test helper tools into a new toast.interactive submodule, so that those tools can be used both by the tutorials and tests.

- Add new %toast magic for ipyparallel initialization in notebooks.


tskisner:
Ok, this is an intermediate step to the docs overhaul.  I am going to try and break up the remaining tasks into smaller pieces.

src/toast/interactive/utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    """
+    if topdir is None:
+        pwd = os.path.abspath(".")
+        topdir = os.path.join(pwd, "toast_output")
+    retdir = topdir
+    if subdir is not None:
+        retdir = os.path.join(topdir, subdir)
+    if (mpicomm is None) or (mpicomm.rank == 0):
+        if not os.path.isdir(topdir):
```
keskitalo, line 35: Matter of preference, but you don't need to check for existence with `os.makedirs(topdir, exist_ok=True)`.

keskitalo (approved):
I really appreciate the new interactive tools.  We'll need a tutorial to showcase them.

Since you had to go through the documentation, it might have been helpful to tag all sections that are in progress or in need of verification. I'm not suggesting to go back now but perhaps in the next PR.

src/toast/interactive/utils.py
tskisner: Good point, will fix it in the next PR.

Issue #606, opened 2022-11-09
Failing conviqt test


gabrielecoppi:
I am running toast3 installed using `pip install --pre --upgrade toast3`. I have set-up a conda environment with all the dependencies (including _madam_ and _conviqt_) and I am failing the _conviqt_ tests. This is the message that I am getting:

```
Proc 0: Traceback (most recent call last):
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/tests/ops_sim_tod_conviqt.py", line 492, in test_sim_TEB_conviqt
    sim_conviqt.apply(data)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/timing.py", line 107, in df
    return f(*args, **kwargs)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/ops/operator.py", line 115, in apply
    self.exec(data, detectors, use_accel=use_accel, **kwargs)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/timing.py", line 107, in df
    return f(*args, **kwargs)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/ops/operator.py", line 52, in exec
    self._exec(
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/timing.py", line 81, in df
    return f(*args, **kwargs)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/ops/conviqt.py", line 264, in _exec
    all_detectors = self._get_all_detectors(data, detectors)
Proc 0:   File "/home/gabriele/miniforge3/envs/cmb/lib/python3.10/site-packages/toast/ops/conviqt.py", line 317, in _get_all_detectors
[...]

PR #610, opened 2022-11-16
Fix new unit test failures

keskitalo:
Address #606


tskisner (approved):
Thanks for fixing this!

PR #611, opened 2022-11-17
fixing bugs on CosmicRays, GainDrifts & GainScrambler 

giuspugl:
the rng operator is intialized with integers like 
https://github.com/hpc4cmb/toast/blob/44d9f84ecc28752de2028cf3f402891ba34a1049/src/toast/ops/gainscrambler.py#L75
, in particular the variable `counter2` was initialized to the detector index `detindx` but it relies on the fact that the `uid` is integer, which is not always the case. for time being i am enumerating the list of the detectors and set `detindx` from there: 
https://github.com/hpc4cmb/toast/blob/ebbfae8409f81ce767b9644950c82c02d6220bd7/src/toast/ops/gainscrambler.py#L79

Any smarter idea?

### DIFF
--- a/src/toast/ops/sim_gaindrifts.py
+++ b/src/toast/ops/sim_gaindrifts.py
@@ -93,7 +93,6 @@ class GainDrifter(Operator):
         "wafer",
         help='focalplane table column to use for grouping detectors: can be any string like "wafer", "pixel"',
     )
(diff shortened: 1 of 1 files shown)


giuspugl:
The tests aren't passing because of the callable trait `responsivity_function`  . This PR aims at addressing also that.

PR #615, opened 2022-11-28
Fix MPI tests and redistribution

tskisner:
Work in progress trying to fix MPI tests and redistribution of observations.


keskitalo (approved):
Looks good!

Issue #613, opened 2022-11-25
Wrong assertion in GainTemplate tests


giuspugl:
i think the result  [detdata key](https://github.com/hpc4cmb/toast/blob/44d9f84ecc28752de2028cf3f402891ba34a1049/src/toast/ops/mapmaker.py#L556)
should encode the TOD calibrated, i.e. corrected by the gain fluctuations ,  expressed  as  the linear combination of  Legendre polinomials weighted by the amplitudes estimated by the toast mapmaker. 
Instead the (https://github.com/hpc4cmb/toast/blob/44d9f84ecc28752de2028cf3f402891ba34a1049/src/toast/tests/template_gain.py#L83)[test suite ] asserts exactly that the `result` data to be an array of    `np.ones` local samples. 
What we should compare to be close to 1 should be instead the  ` ops.Calibrate.template_matrix.templates[0].data['calibrator_solve_amplitudes']`

giuspugl:
Also  in 
https://github.com/hpc4cmb/toast/blob/44d9f84ecc28752de2028cf3f402891ba34a1049/src/toast/ops/mapmaker_templates.py#L1009
 what we should be dividing is not the template by the uncalibrated signal but the fitted gain amplitudes. 
Similarly to what we do in https://github.com/hpc4cmb/toast/blob/ddae67de4154067d2317999f6cf6c31dd48ae047/src/toast/todmap/mapmaker.py#L411

PR #617, opened 2022-12-05
implementing a new assertion test for the Gaintemplate fitting

giuspugl:
this PR addresses the issues #613.

### DIFF
--- a/src/toast/tests/template_gain.py
+++ b/src/toast/tests/template_gain.py
@@ -12,8 +12,8 @@ from .. import ops
 from ..templates import GainTemplate
 from ..utils import rate_from_times
 from ._helpers import create_outdir, create_satellite_data, close_data
-from .mpi import MPITestCase
-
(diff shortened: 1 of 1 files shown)


giuspugl:
at the moment the test doesn't pass, mainly because : 
- an error on the size of the fitted gain_amplitudes `calibration.template_matrix.templates[0].data['Calibrate_solve_amplitudes'] ` instance of https://github.com/hpc4cmb/toast/blob/44d9f84ecc28752de2028cf3f402891ba34a1049/src/toast/ops/mapmaker_templates.py#L33  (should be `n_observations *  n_detectors * n_polyorders` instead there is an extra list of zeros ) 
-   `calibration.template_matrix.templates[0].data['temp_LHS']`  has only data from 1 detector  (am running the test serially ) 
-  why does `calibration.template_matrix.templates[0].data['temp_LHS']`  have `1/K` units  ?

PR #618, opened 2022-12-08
Processing mask

keskitalo:
Support a processing mask in the filtering operators.

It also fixes an issue with the filters setting flagged samples to zero. Flagged samples are not considered when fitting templates, but templates are subtracted from flagged samples.

### DIFF
--- a/src/toast/observation.py
+++ b/src/toast/observation.py
@@ -68,6 +68,7 @@ def set_default_values(values=None):
         "shared_mask_irregular": 4,
         "det_mask_invalid": 1,
         "det_mask_sso": 1 + 2,
+        "det_mask_processing": 2,
         #
         # ground-specific flag masks
         #
--- a/src/toast/ops/madam.py
+++ b/src/toast/ops/madam.py
@@ -899,7 +899,7 @@ class Madam(Operator):
                     for psdstart, psd, detw in psdlist:
                         psdstarts.append(psdstart)
                         psdvals.append(psd)
-                    detweights[idet] = psdlist[0][2].to_value(1 / u.K**2)
+                    detweights[idet] = psdlist[0][2]
                 npsdtot = np.sum(npsd)
                 psdstarts = np.array(psdstarts, dtype=np.float64)
                 psdvals = np.hstack(psdvals).astype(madam.PSD_TYPE)
--- a/src/toast/ops/madam.py
+++ b/src/toast/ops/madam.py
@@ -899,7 +899,7 @@ class Madam(Operator):
                     for psdstart, psd, detw in psdlist:
                         psdstarts.append(psdstart)
                         psdvals.append(psd)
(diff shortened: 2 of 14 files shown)


src/toast/observation.py
```diff
@@ -68,6 +68,7 @@ def set_default_values(values=None):
         "shared_mask_irregular": 4,
         "det_mask_invalid": 1,
         "det_mask_sso": 1 + 2,
+        "det_mask_processing": 2,
```
tskisner, line 71: This is fine, although we could consider renaming this to `det_mask_filter` if that is what the intended use is rather than the generic "processing".

src/toast/ops/madam.py
```diff
@@ -899,7 +899,7 @@ def _stage_data(
                     for psdstart, psd, detw in psdlist:
                         psdstarts.append(psdstart)
                         psdvals.append(psd)
-                    detweights[idet] = psdlist[0][2]
+                    detweights[idet] = psdlist[0][2].to_value(1 / u.K**2)
```
tskisner, line 902: Thanks, this was fixed in the other PR as well.  Better to get fixed twice than not at all :-)

src/toast/tests/ops_polyfilter.py
```diff
@@ -100,9 +100,9 @@ def test_polyfilter(self):
             order=1,
             det_data=defaults.det_data,
             det_flags=defaults.det_flags,
-            det_flag_mask=defaults.shared_mask_invalid,
+            det_flag_mask=defaults.det_mask_invalid,
```
tskisner, line 103: Good catch!

src/toast/ops/scan_healpix.py
```diff
@@ -242,7 +242,8 @@ class ScanHealpixMask(Operator):
     )
 
     det_flags_value = Int(
-        1, help="The detector flag value to set where the mask result is non-zero"
+        defaults.det_mask_processing,
+        help="The detector flag value to set where the mask result is non-zero",
```
tskisner, line 246: We should add this change to the `ScanWCSMask` operator as well.

tskisner (approved):
I think it is good to have the concept of a more restrictive flagging for filtering and processing.  If you want to update `ScanWCSMask` in this PR that is fine or I can add it when rebasing the split ground PR.

src/toast/observation.py
keskitalo: The flag bits could be used in destriping as well so I picked this general name.

src/toast/ops/madam.py
keskitalo: It was preventing me from running the unit tests

src/toast/ops/scan_healpix.py
keskitalo: Done

PR #620, opened 2022-12-19
Common mode noise

keskitalo:
Implement an operator that modifies the noise model to include common noise modes.

### DIFF
--- a/src/toast/ops/common_mode_noise.py
+++ b/src/toast/ops/common_mode_noise.py
@@ -102,13 +102,10 @@ class CommonModeNoise(Operator):
                 msg = f"You must set the '{trait}' trait before calling exec()"
                 raise RuntimeError(msg)
 
-        # FIXME : Is this necessary?
-        """
         if detectors is not None:
             msg = "You must run this operator on all detectors at once"
             log.error(msg)
             raise RuntimeError(msg)
-        """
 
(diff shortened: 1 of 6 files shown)


src/toast/ops/common_mode_noise.py
```diff
@@ -0,0 +1,241 @@
...
+                    noise_key = f"{self.name}_{self.component}"
+                else:
+                    noise_key = f"{self.name}_{self.component}_{key}"
+                mixing_matrix[noise_key] = {}
+                noise_uid = name_UID(noise_key)
+                nse_keys.append(noise_key)
+                nse_freqs[noise_key] = freqs * u.Hz
+                nse_psds[noise_key] = net**2 * (
+                    (freqs**alpha + fknee**alpha) / (freqs**alpha + fmin**alpha)
+                )
```
tskisner, line 186: Not needed for this PR, but the analytic 1/f noise model [is used in several places](https://github.com/hpc4cmb/toast/blob/017d65603a3d6e6566090e52634fcd4965713de2/src/toast/noise_sim.py#L106-L114).  Perhaps we should make a helper function for that in `toast.sim_noise.py` someday?

tskisner (approved):
Thanks for the detailed unit tests!

PR #622, opened 2023-01-09
Conviqt tweaks

keskitalo:
Make the use of the conviqt TEB operator more intuitive: the beam and sky components are now synthesized automatically rather than having the user do it.

The PR also changes the way the beam file names are synthesized. Users can use any focalplane attributes in the file name and they automatically get replaced with the corresponding values in the file path.

### DIFF
--- a/src/toast/ops/conviqt.py
+++ b/src/toast/ops/conviqt.py
@@ -3,8 +3,10 @@
 # a BSD-style license that can be found in the LICENSE file.
 
 import os
+import tempfile
 import warnings
 
+import healpy as hp
 import numpy as np
 import traitlets
 from astropy import units as u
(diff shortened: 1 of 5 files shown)


keskitalo:
@kisner @giuseppe I just merged a branch in `sotodlib` that depends on this PR. Should I expect a fairly quick turnaround on this PR or should I revert the `sotodlib` merge so that `toast_so_sim.py` workflow isn't broken?

giuspugl:
i really like this tweaks so i would  merge this into toast , they are definitely more user friendly. thanks !

src/toast/ops/noise_estimation.py
```diff
@@ -173,7 +173,7 @@ class NoiseEstim(Operator):
     )
 
     pairs = List(
-        default_value=None,
+        # default_value=None,
```
tskisner, line 176: Thanks, I will fix all of these in a systematic way during the fixes to the template calibration.

src/toast/ops/conviqt.py
```diff
@@ -750,30 +755,42 @@ def _exec(self, data, detectors=None, **kwargs):
         for det in all_detectors:
             verbose = self.comm.rank == 0 and self.verbosity > 0
 
+            # Find one process that has focalplane data for this detector
+            # and broadcast the focalplane properties
```
tskisner, line 759: In general, the same detector could have different properties from one observation to the next.  For example, a different pointing solution.  This technique is OK as long as the properties used later to select the beam file name are constant across the data set.

tskisner (approved):
Thanks for this work!

PR #625, opened 2023-01-13
clean up jax device assignement code

nestordemeure:
includes:
- small fixes to run with the latest versions of Numpy and JAX
- a clean up of the jax device asignement code


src/toast/accelerator/accel.py
```diff
@@ -48,38 +50,50 @@
...
+    Args:
+        node_procs (int): number of processes per node
+        node_rank (int): rank of the current process, within the node
+        disabled (bool): gpu computing is disabled
+
+    Returns:
+        None: the device is stored in a backend specific global variable
+    """
+    # FIXME some functions (such as poiting_detector) require the omp device to have been assigned
+    # so it should be called even when using JAX or running on CPU
```
tskisner, line 91: This behavior is intentional.  The MPI initialization assigns GPU devices in the compiled extension.  If target offload support is not being used, the assigned device is set to the host device.  This assignment is then used by the compiled code to determine whether we are offloading.

tskisner (approved):
Thanks for these fixes on the jax side!

PR #627, opened 2023-01-19
Add support for horizontal scanning strategies that cross the north-s…

keskitalo:
…outh meridian


keskitalo:
Example hit map for a Chilean telescope with a 10-degree FOV, observing three Southern horizontal patches at 50, 55 and 60 degrees, each with 60-degree azimuth throw. The period is January and Sun is avoided at 30-degree boresight distance.

image

tskisner (approved):
Looks good, thanks!

PR #619, opened 2022-12-12
Large updates to the testing and binary wheel workflows.

tskisner:

PR #628, opened 2023-01-21
Clean up use of container traits.

tskisner:
Although container traits (List, Set, Tuple, Dict) formally support None values, it is challenging to distinguish between this and empty containers when passing things through the config dictionary to/from files on disk.  Because of that, we internally enforce these trait types to not have None values.  This work cleans up all operators to conform.  Additionally, there were 3 List trait definitions that tried to enforce the type of the elements with the "trait=" syntax. That caused further issues and these were removed.  In practice we check the container elements already later in the code.

### DIFF
--- a/src/toast/ops/arithmetic.py
+++ b/src/toast/ops/arithmetic.py
@@ -6,7 +6,7 @@ import traitlets
 
 from ..mpi import MPI
 from ..timing import function_timer
-from ..traits import Int, List, Unicode, trait_docs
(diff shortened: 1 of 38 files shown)


tskisner:
The linux tests in CI are broken and being fixed in #619.  I have run the tests locally on 4 MPI processes and with MPI disabled.  These changes are just tedious conversions of default trait container values from None to empty.

PR #629, opened 2023-01-26
Use a reversed signal to avoid discontinuity

keskitalo:
Improve time constant operator to handle data with offsets and drifts


keskitalo:
The only downside I see is that the FFT length is up to two times longer.  The advantage of this approach is that it makes no assumptions about the signal content.

tskisner (approved):
Thanks for this, I verified locally that it solved the problem on SO LAT Jupiter scans.

PR #630, opened 2023-01-30
Benchmark updates.

tskisner:
* Add PixelData method to compute simple stats on distributed maps

* Simplify Ground job setup by removing the need to pre-cache the simulated atmosphere.

* Write generated stats to the benchmark output directory.  If the given benchmark script / case exists in the bundled json file, then print comparison values.

* Add small script to update the bundled json file from job outputs.


nestordemeure (approved):
Looks fine.

src/toast/pixels.py
```diff
@@ -921,6 +921,125 @@ def sync_alltoallv(self, local_func=None):
         self.reverse_alltoallv()
         return
 
+    @function_timer
+    def stats(self, comm_bytes=10000000):
+        """Compute some simple statistics of the pixel data.
+
+        The map should already be consistent across all processes with overlapping
```
keskitalo, line 928: Any way to check if the map was synchronized?

src/toast/scripts/benchmarking_utilities.py
```diff
@@ -664,7 +724,14 @@ def create_input_maps(
         new=True,
         fwhm=np.radians(3.0 / 60.0),
     )
-    healpy.write_map(input_map_path, maps, nest=True, fits_IDL=False, dtype=np.float32)
+    healpy.write_map(
```
keskitalo, line 727: If this is taking a long time, we might consider using the HDF5 format used in DC0 sims.

keskitalo (approved):
Looks great! Just two small questions.

src/toast/pixels.py
tskisner: I suppose we could have each process compute a hash of their local submap data and compare across all processes that had the same submap.  This would require some extra coding.  Since we are only using this in the benchmarks at the moment where we control what has been done to the map data, I would prefer to punt that to the future if possible.

src/toast/scripts/benchmarking_utilities.py
tskisner: The call the synfast is the slow part I believe.

PR #632, opened 2023-02-06
More bandpass methods

keskitalo:
This PR addresses some development needs that arose from simulating a catalog of point sources.


tskisner (approved):
These changes look good.  I like that derived classes would just need to implement `freqs()` and `bandpass()` methods for a given instrument (for example returning measured FTS spectra) and then the `convolve()` and flux calibration methods would use them.

PR #633, opened 2023-02-11
Small fix to satellite benchmark

tskisner:
This adds a fixed mission start date, rather than using the current time.


tskisner:
This is a trivial change, to only the benchmark script.  I have verified the results agree between a local system and perlmutter.nersc.gov.  Merging...

PR #634, opened 2023-02-11
Add general support for compression of DetectorData

tskisner:
The DetectorData class is our primary container for timestream data from a set of co-sampled detectors.  This is used for signal, flags, detector pointing, etc.  When storing this data to disk, it is useful to compress it.  This work adds support for gzip and flac compression (and a "none" type for testing) of DetectorData:

* Add optional compile time support for linking to libFLAC versions >= 1.4.0, which was the first version to support 32bit integers.

* Add compiled utility functions for compressing / decompressing 32bit integer data with FLAC, and pybind11 interfaces to these.

* Add new file `io/compression.py`, which has functions for compressing and decompressing DetectorData.  The compressed format is a stream of bytes, a set of byte ranges for each detector, and a dictionary of compression properties.

* Conditional routines to compress / decompress DetectorData with FLAC.  For floating point data, an optional resolution can be specified for each detector to represent the floating point amount of one integer bit.  Default behavior maps the full amplitude range of the input to the signed 31bit range supported in flac.

### DIFF
--- a/cmake/FindFLAC.cmake
+++ b/cmake/FindFLAC.cmake
@@ -1,0 +1,66 @@
+# - Find FLAC
+# Find the native FLAC includes and libraries
+#
+#  FLAC_INCLUDE_DIRS - where to find FLAC headers.
+#  FLAC_LIBRARIES    - List of libraries when using libFLAC.
+#  FLAC_FOUND        - True if libFLAC found.
+#  FLAC_DEFINITIONS  - FLAC compile definitons 
+
+if (FLAC_INCLUDE_DIR)
+    # Already in cache, be silent
+    set (FLAC_FIND_QUIETLY TRUE)
+endif ()
+
+find_package (Ogg QUIET)
+
+find_package (PkgConfig QUIET)
+pkg_check_modules(PC_FLAC QUIET flac)
+
+set(FLAC_VERSION ${PC_FLAC_VERSION})
+
+find_path (FLAC_INCLUDE_DIR FLAC/stream_decoder.h
+	HINTS
+		${PC_FLAC_INCLUDEDIR}
+		${PC_FLAC_INCLUDE_DIRS}
+		${FLAC_ROOT}
+	)
+
+# MSVC built libraries can name them *_static, which is good as it
+# distinguishes import libraries from static libraries with the same extension.
+find_library (FLAC_LIBRARY
+	NAMES
+		FLAC
+		libFLAC
+		libFLAC_dynamic
+		libFLAC_static
+	HINTS
+		${PC_FLAC_LIBDIR}
+		${PC_FLAC_LIBRARY_DIRS}
+		${FLAC_ROOT}
+	)
+
+# Handle the QUIETLY and REQUIRED arguments and set FLAC_FOUND to TRUE if
+# all listed variables are TRUE.
+include (FindPackageHandleStandardArgs)
+find_package_handle_standard_args (FLAC
+	REQUIRED_VARS
+		FLAC_LIBRARY
+		FLAC_INCLUDE_DIR
+	VERSION_VAR
+        FLAC_VERSION
+	)
+
+if (FLAC_FOUND)
+	set (FLAC_INCLUDE_DIRS ${FLAC_INCLUDE_DIR})
+	set (FLAC_LIBRARIES ${FLAC_LIBRARY} ${OGG_LIBRARIES})
+    if (NOT TARGET FLAC::FLAC)
+		add_library(FLAC::FLAC UNKNOWN IMPORTED)
+		set_target_properties(FLAC::FLAC PROPERTIES
+			INTERFACE_INCLUDE_DIRECTORIES "${FLAC_INCLUDE_DIR}"
+			IMPORTED_LOCATION "${FLAC_LIBRARY}"
+			INTERFACE_LINK_LIBRARIES Ogg::ogg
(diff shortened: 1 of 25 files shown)


cmake/FindFLAC.cmake
```diff
@@ -0,0 +1,66 @@
+# - Find FLAC
```
keskitalo, line 1: No copyright header

etc/flac_test/Makefile
```diff
@@ -0,0 +1,10 @@
+
```
keskitalo, line 1: no copyright

etc/flac_test/main.cpp
```diff
@@ -0,0 +1,342 @@
+
```
keskitalo, line 1: no copyright

etc/flac_test/Makefile
```diff
@@ -0,0 +1,10 @@
...
+CXX := g++
+CXXFLAGS := -O0 -g -fPIC -I/home/kisner/software/conda/envs/toast/include
```
keskitalo, line 3: Did you intend to commit this file?

src/toast/io/CMakeLists.txt
```diff
@@ -5,6 +5,9 @@ install(FILES
     __init__.py
     observation_hdf_save.py
     observation_hdf_load.py
+    observation_hdf_load_v0.py
```
keskitalo, line 8: This seems like a debugging modification.

src/toast/io/compression.py
```diff
@@ -0,0 +1,274 @@
...
+                    "Cannot specify quanta for FLAC compression of integers"
+                )
+        else:
+            quanta = None
+            comp_params["quanta"] = quanta
+
+        if detdata.dtype in ftypes:
+            (
+                comp_bytes,
+                comp_ranges,
```
keskitalo, line 131: I would be interested to see how polynomial detrending would reduce the dynamical range of the data. Also, the data we simulate probably could do with more digitization noise to improve the compression ratio.

src/toast/io/observation_hdf_load_v0.py
```diff
@@ -0,0 +1,132 @@
+# Copyright (c) 2021-2021 by the parties listed in the AUTHORS file.
```
keskitalo, line 1: Should this file be included in the PR?

src/toast/observation_data.py
```diff
@@ -439,7 +439,13 @@ def __eq__(self, other):
             return False
         if self.units != other.units:
             return False
-        if not np.allclose(self.data, other.data):
+        if self.dtype == np.dtype(np.float64):
+            if not np.allclose(self.data, other.data, rtol=1.0e-5, atol=1.0e-8):
+                return False
+        elif self.dtype == np.dtype(np.float32):
+            if not np.allclose(self.data, other.data, rtol=1.0e-3, atol=1.0e-5):
```
keskitalo, line 446: This seems like a very relaxed comparison for float32.

src/toast/tests/io_compression.py
```diff
@@ -0,0 +1,395 @@
+# Copyright (c) 2015-2021 by the parties listed in the AUTHORS file.
```
keskitalo, line 1: Outdated years

keskitalo (approved):
Looks great! Really appreciate the thorough testing.

src/toast/io/observation_hdf_load_v0.py
tskisner: Yes, the idea is to always support reading past versions of the file format, but only support writing the latest format.  As the format evolves, some of the old reading code will be moved out of the way into separate files like this.

etc/flac_test/Makefile
tskisner: Yes, the `etc/` directory contains small accessory test codes which are useful to keep around as an archive, but are not directly included in the package.

etc/flac_test/main.cpp
tskisner: Fixed

cmake/FindFLAC.cmake
tskisner: Thanks for catching this- it was copied from libsndfile and I have added the proper copyright at the top.

src/toast/tests/io_compression.py
tskisner: Fixed, but we should also just recursively update the ending copyright years across the codebase (as a separate, standalone PR).

src/toast/observation_data.py
tskisner: Good point, I was trying to be flexible for the case of verifying data written to disk with FLAC compression and then loaded back in.  But I should just do a custom comparison for that case.  Will implement it.

src/toast/io/compression.py
tskisner: The FLAC algorithm does do [modeling of the signal](https://xiph.org/flac/format.html#prediction) to reduce the residuals that need to be encoded.  However I believe this is done on spans of the blocksize number of samples (e.g. 4096 or 8192 samples).  Certainly doing a more complicated fit would be possible, in which case we should replace the gain / offset model with a set of coefficients for each detector.  Also note by default we map the full range of data to +/- 2^30, but one can set the quanta for all detectors to a constant from the commandline / config for the SaveHDF5 operator.

PR #635, opened 2023-02-13
Tweak the format of the toast ground schedule

keskitalo:
* Choose a more logical branch when outputting scans that cross the zero azimuth
* Make a new concise schedule format the default.

### DIFF
--- a/src/toast/schedule.py
+++ b/src/toast/schedule.py
@@ -295,20 +295,6 @@ class GroundSchedule(object):
             except Exception:
                 start_time = dateutil.parser.parse(start_time)
                 stop_time = dateutil.parser.parse(stop_time)
-            # DEBUG begin
-            print(f"PARSED ground scan with \n"
-                  f"PARSED name = {name} \n"
-                  f"PARSED start_time = {start_time} \n"
-                  f"PARSED stop_time = {stop_time} \n"
-                  f"PARSED boresight_angle = {boresight_angle} \n"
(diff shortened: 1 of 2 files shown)


src/toast/schedule.py
```diff
@@ -337,6 +295,20 @@ def _parse_line(line):
...
+                  f"PARSED stop_time = {stop_time} \n"
+                  f"PARSED boresight_angle = {boresight_angle} \n"
+                  f"PARSED azmin = {azmin} \n"
+                  f"PARSED azmax = {azmax} \n"
+                  f"PARSED el = {el} \n"
+                  f"PARSED scan = {scan} \n"
+                  f"PARSED subscan = {subscan} \n",
+                  flush=True,
+            )
+            # DEBUG end
```
tskisner, line 311: Did you intend to leave this here?  Maybe concatenate to a string and use `log.debug()`?

tskisner (approved):
The updates themselves are fine, but for consideration in the future:
- We could use a `QTable` for both the satellite and ground schedules and just get rid of the `SatelliteScan` and `GroundScan` objects completely.  Then we could just dump / load to ECSV and get rid of the custom ground schedule text parsing.
- We could split up `schedule_sim_ground.py` into several files since it is so long.

src/toast/schedule.py
keskitalo: I didn't mean to leave it. Thanks for noticing!

PR #638, opened 2023-02-17
Support loading list of HDF5 files

tskisner:
This just adds an additional option to the `LoadHDF5` operator which supports loading a list of filenames, in addition to loading a whole directory of files.


tskisner:
Superseded by #639

PR #642, opened 2023-02-28
fixed missing/order for provide

nestordemeure:
Added missing `requires` for operators.

This introduce occasional overlaps between `requires` and `provides` (something that did not previously occur in the codebase) which required tweaking the order of operations in the `requires`/`provides` for the `Pipeline` class (otherwise we end up with missing `requires`/`provides` for the full pipeline) as well as not deleting intermediate product (otherwise we end up with missing `requires` in the pipeline, that particular change that was coming with hybrid pipelines anyway).


tskisner (approved):
Thanks for catching these dependency errors.

PR #643, opened 2023-02-28
Jax bugfix

nestordemeure:
* fixed bug due to empty intervals
* made staging accelerator tests compatible with JAX's staging policy
* passed missing use_accel in accelerator test

### DIFF
--- a/src/toast/tests/accelerator.py
+++ b/src/toast/tests/accelerator.py
@@ -203,7 +203,7 @@ class AcceleratorTest(MPITestCase):
         # Duplicate for future comparison
         check_data = Data(comm=data.comm)
         for ob in data.obs:
-            check_data.obs.append(ob.duplicate())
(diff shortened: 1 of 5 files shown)


src/toast/tests/accelerator.py
```diff
@@ -134,36 +137,41 @@ def test_memory(self):
...
+            data[tname] = buffer
 
         # Check that it is present
         for tname, buffer in data.items():
             self.assertTrue(accel_data_present(buffer))
 
         # Change host copy
         for tname, buffer in data.items():
+            # FIXME we are changing the host copy while data is on device
+            #       is it legal?!
```
tskisner, line 151: These lines are just testing the low-level data movement functions.  There are checks in the `AcceleratorObject` methods against this kind of thing.

tskisner (approved):
Thanks for the cleanup!

PR #644, opened 2023-03-02
Hybrid pipeline

nestordemeure:
Added hybrid GPU pipelines, able to run mixes of CPU and GPU operators.

* adds a `SetDict` class to `utils.py` to easily perform set operation on lists of keys that might be on GPU
* update data movement infrastructure in `Pipeline` class
* add a `TOAST_GPU_HYBRID_PIPELINES` environment variable (defaults to `True`, can be set manually to `False`) to forbid hybrid pipelines (one can also forbid them locally by setting a class variable)
* update the `requires` / `provides` of several operators (most of them untested until now)


src/toast/ops/sim_tod_atm_observe.py
```diff
@@ -595,7 +595,7 @@ def _requires(self):
         if self.det_flags is not None:
             req["detdata"].append(self.det_flags)
         if self.weights is not None:
-            req["weights"].append(self.weights)
+            req["detdata"].append(self.weights)
```
tskisner, line 598: Thanks for catching this!

tskisner (approved):
Thanks for working on the dependency logic.  Although the `provides` and `requires` lists have changed their meaning slightly, I agree these changes are needed to consistently move data back and forth.  As a separate future PR we could improve the names to be closer to the intended meanings, but for now we have:
```
requires == "Things which must already exist and will be used read-only"
provides == "Things which might exist, are created if needed, and modified"
```

nestordemeure:
The best formulation I can come up with is this (in both cases things might already exist, or not):

```
requires == "Things which might exist and will be read",
provides == "Things which might exist and will be modified".
```

`reads` and `modifies` maybe?

tskisner:
Something for a future PR...

keskitalo (approved):
Looks great! I really appreciate all the fixes to the `requires()` and `provides()` methods.

PR #646, opened 2023-03-09
Overhaul SSO flagging, improve unit test

keskitalo:
Fix SSO flagging issues that were found during CMB-S4 simulations.


tskisner (approved):
Thanks for this work!

PR #647, opened 2023-03-09
Tweak tolerances for detdata equality tests.

tskisner:


tskisner:
Approved verbally on slack

PR #650, opened 2023-03-23
Code modification introduced for benchmarking

nestordemeure:
This branch introduces a number of modifications that were useful for benchmarking.

In particular:
* having the `TOAST_FUNCTIME` read its value instead of considering any value true (this is useful when switching from true/1 to false/0 inside a script),
* updating the `jax_accel_assign_device` function to deal properly with gpu oversubscription (several process using the same gpu)


tskisner (approved):
Thanks for testing this!

PR #651, opened 2023-03-31
Fix JAX multi process/nodes

nestordemeure:
Fixed device picking function to deal with both multi processes and multinode cases properly.


tskisner (approved):
Looks fine.

PR #652, opened 2023-04-03
Refactor of OpenMP target operations

tskisner:
This is still a work in progress.  Opening a draft PR in order to more easily track progress.


src/toast/__init__.py
nestordemeure, line 1: Can we set `TOAST_GPU_MEM_GB` to something 90% of the memory available (GPU memory divided by number of process sharing the GPU) as a more flexible default?

src/toast/accelerator/accel.py
```diff
@@ -157,6 +163,31 @@ def accel_data_create(data):
...
+    Args:
+        data (array):  The host array.
+    Returns:
+        None
+
+    """
+    if use_accel_omp:
+        omp_accel_reset(data)
+    elif use_accel_jax:
+        pass
```
nestordemeure, line 185: accel_data_reset needs a JAX implementation, I can implement that in a later PR.
otherwise, something like this should be enough: 

```python
data.data = jnp.zeroes_like(data.data)
```

src/toast/accelerator/kernel_registry.py
```diff
@@ -80,7 +80,8 @@ def dispatch(
         if (
             (impl == ImplementationType.JAX) or (impl == ImplementationType.COMPILED)
         ) and accel_enabled():
-            registry[name][impl] = function_timer(function_datamovementtracker(f))
+            # registry[name][impl] = function_timer(function_datamovementtracker(f))
```
nestordemeure, line 83: why was `datamovementtracker` removed?

src/toast/observation_data.py
```diff
@@ -535,6 +529,18 @@ def _accel_delete(self):
         elif use_accel_jax:
             self._data = accel_data_delete(self._data)
 
+    def _accel_reset(self):
+        log = Logger.get()
+        log.verbose(f"DetectorData _accel_reset")
+        if use_accel_omp:
+            accel_data_reset(self._raw)
+        elif use_accel_jax:
+            # NOTE: we create an array of zeroes directly on GPU to avoid
```
nestordemeure, line 538: this could become a call to accel_reset once it is updated, i will do it in a later PR

src/toast/ops/noise_weight/kernels.py
```diff
@@ -4,14 +4,19 @@
...
     from .kernels_jax import noise_weight_jax
 
 
+@kernel(impl=ImplementationType.COMPILED, name="noise_weight")
+def noise_weight_compiled(*args, use_accel=False):
+    return libtoast_noise_weight(*args, use_accel)
+
+
 @kernel(impl=ImplementationType.DEFAULT)
 def noise_weight(
```
nestordemeure, line 21: this can now call libtoast_noise_weight

nestordemeure (approved):
I added a few comments but i can do the JAX specific fixes ,yself in a later PR.

src/toast/__init__.py
tskisner: This variable was to control the size of the OpenMP target mempool, but that feature is currently disabled, since at least the nvidia implementation seems to be very fast with target alloc.

src/toast/accelerator/accel.py
tskisner: Done

src/toast/observation_data.py
tskisner: Ok, I left this for the next PR :-)

src/toast/ops/noise_weight/kernels.py
tskisner: Fixed.

src/toast/accelerator/kernel_registry.py
tskisner: I updated the comment.  This was disabled for now since it breaks for kernel parameters that are not buffer objects (scalars, bool, etc).  As soon as the movement record can handle or ignore those, we can re-enable.

PR #654, opened 2023-04-04
Fixed offset kernel bug

nestordemeure:
* Fixes the bug in the add and project offset kernels (C++ and JAX) which fixes a segfault in the ground benchmark.
* Fixes a bug in a JAX kernel causing its result to deviate from the C++ CPU reference.
* Most classes now takes `use_accel=True|None|False`.
* Fixed performance bug due to initializing templates on GPU.
* Removed template reinitialization after solve.
* Small fixes to get some of the unit tests running.


src/toast/_libtoast/ops_mapmaker_utils.cpp
```diff
@@ -314,12 +314,9 @@ void init_ops_mapmaker_utils(py::module & m) {
                                 use_det_flags
                             );
 
-                            // #pragma omp critical
-                            {
-                                for (int64_t iw = 0; iw < nnz; iw++) {
-                                    #pragma omp atomic
-                                    raw_zmap[zoff + iw] += zmap_val[iw];
-                                }
+                            for (int64_t iw = 0; iw < nnz; iw++) {
```
tskisner, line 317: This is also changed in #652, so I'll deal with that conflict after this is merged.

src/toast/_libtoast/template_offset.cpp
```diff
@@ -172,34 +173,34 @@ void init_template_offset(py::module & m) {
...
+                        int64_t first_samp = dev_intervals[iview].first;
                         # pragma omp parallel for default(shared)
                         for (
                             int64_t isamp = dev_intervals[iview].first;
                             isamp <= dev_intervals[iview].last;
                             isamp++
                         ) {
                             int64_t d = data_index * n_samp + isamp;
-                            int64_t amp = offset + (int64_t)(isamp / step_length);
+                            int64_t amp = offset +
```
tskisner, line 202: Also changed in #652...

src/toast/ops/mapmaker_templates.py
```diff
@@ -191,6 +191,9 @@ def _exec(self, data, detectors=None, use_accel=False, **kwargs):
         # On the first call, we initialize all templates using the Data instance and
         # the fixed options for view, flagging, etc.
         if not self._initialized:
+            if use_accel:
```
tskisner, line 194: Good to have this check, thanks!

src/toast/ops/mapmaker_utils/kernels_numpy.py
```diff
@@ -93,8 +93,8 @@ def build_noise_weighted_numpy(
             )
 
 
-@kernel(impl=ImplementationType.NUMPY, name="cov_acccum_diag_hits")
-def cov_accum_diag_hits_numpy(nsub, nsubpix, nnz, submap, subpix, hits):
+@kernel(impl=ImplementationType.NUMPY, name="cov_accum_diag_hits")
```
tskisner, line 96: Good catch on that typo!

src/toast/ops/operator.py
```diff
@@ -22,7 +22,7 @@ def _exec(self, data, detectors=None, **kwargs):
         raise NotImplementedError("Fell through to Operator base class")
 
     @function_timer_stackskip
-    def exec(self, data, detectors=None, use_accel=False, **kwargs):
+    def exec(self, data, detectors=None, use_accel=None, **kwargs):
```
tskisner, line 25: I definitely think this tri-state is a good idea

src/toast/scripts/benchmarking_utilities.py
```diff
@@ -804,7 +804,8 @@ def run_mapmaker(job_ops, args, tmpls, data):
     job_ops.mapmaker.output_dir = args.out_dir
 
     # Run the map making
-    job_ops.mapmaker.apply(data)
+    # NOTE: we set `use_accel=False` as templates cannot be initialized on device
+    job_ops.mapmaker.apply(data, use_accel=False)
```
tskisner, line 808: This should be `None` instead of `False`.  Setting it to False will disable GPU use for the entire mapmaking.  The specific case of building the RHS on the CPU is already handled separately in this PR.

tskisner (changes requested):

tskisner:
Everything fine once you re-enable GPU mapmaking in the benchmark tools.

src/toast/scripts/benchmarking_utilities.py
nestordemeure: Fixed!

PR #655, opened 2023-04-07
Kernel timing

nestordemeure:
* added a `toast_merge_timings.py` script
* default kernel now call the compiled kernel directly with no intermediate dispatch (makes merge between default and compiled/jax/numpy runs possible)
* added function timer to most device data moving function
* added optional display of data movement at the end of the sat/ground benches


tskisner (approved):
Looks good, there are several things in here that conflict with work in #652, so all the more reason to get that merged before working further on these files.

PR #656, opened 2023-04-12
Multi flavor mapmaking

keskitalo:
Modify `SimScan` to support multiple flavors at once and modify 'toast_sim_ground.py' to use them.


keskitalo:
The timer unit tests that seem to fail on MacOS also fail on the Ubuntu Linux subsystem for Windows and are not related to this PR.

tskisner (approved):
This all seems fine.

tskisner:
For the timer test failures, we could either disable them or figure out how to tell we are running inside Linux on windows...

PR #657, opened 2023-04-17
JAX related fixes

nestordemeure:
- small performance improvement to some JAX kernels
- reintroduced datamovement tracker
- improved mergin of timing csv
- updated accel jax reset
- minor cleanups


src/toast/ops/scan_map/kernels.py
```diff
@@ -84,10 +84,10 @@ def scan_map(
 
 @kernel(impl=ImplementationType.COMPILED, name="scan_map")
 def scan_map_compiled(*args, use_accel=False):
-    return scan_map_compiled(*args, use_accel)
+    return libtoast_scan_map(*args, use_accel)
 
 
-def scan_map_compiled(
+def libtoast_scan_map(
```
tskisner, line 90: I would like to keep the name of this function as `scan_map_compiled`, because it is a function defined here in python, not an imported name from libtoast.

src/toast/scripts/toast_merge_timings.py
```diff
@@ -37,41 +38,40 @@ def process_timer_path(s):
     :return: A tuple containing the simplified path, kernel type, and operation name.
     """
     # Remove occurrences of '(function) ' and '._exec'
-    simplified_path = s.replace("(function) ", "").replace("._exec", "")
+    simplified_path = s.replace('(function) ', '').replace('._exec', '')
```
tskisner, line 41: Something is changing the quotes here.  The black formatter should keep everything as double quotes, which is the desired behavior unless there are nested quotes.

tskisner (changes requested):
Thanks for the fixes, just one small change plus formatting needed.

src/toast/ops/scan_map/kernels.py
nestordemeure: I avoided `scan_map_compiled` to avoid having two function having the sane name, do you have a suggested other name?

src/toast/scripts/toast_merge_timings.py
nestordemeure: Okay, I will run the formatter!

src/toast/ops/scan_map/kernels.py
tskisner: Ah, did not notice from the highlighted line that we had duplicate naming...  Ok, then leave it as it is.

Issue #661, opened 2023-05-01
NameError when applying common mode noise


antolonappan:
A NameError occurs when trying to generate common mode noise  using the following code:

```
noise_model = ops.DefaultNoiseModel()
noise_model.apply(data)

common_mode_model = ops.CommonModeNoise(
            fmin=1e-5 * u.Hz,
            fknee=1.0 * u.Hz,
            alpha=2.0,
            NET=1.0 * u.K / u.Hz**0.5,
            det_data="corr_noise"
        )
common_mode_model.apply(data)
```

The error message shows:

```
NameError: name 'detset' is not defined
```

antolonappan:
Yes, this issue is with toast3 branch.
Thanks.

keskitalo:
Closing the issue with the fix merged.

PR #665, opened 2023-05-17
Common mode fix

giuspugl:
with this PR we aim at fixing the issue on-going with the common mode operator, https://github.com/hpc4cmb/toast/issues/661 
to also make sure that the error will be caught  up in the future i've also included an extra-test.

### DIFF
--- a/src/toast/ops/common_mode_noise.py
+++ b/src/toast/ops/common_mode_noise.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2015-2022 by the parties listed in the AUTHORS file.
+# Copyright (c) 2015-2023 by the parties listed in the AUTHORS file.
 # All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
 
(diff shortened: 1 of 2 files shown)


tskisner (approved):
Thanks for the fix and the unit test.  Based on the code where it is checking if a detector is in `detset`, it seemed like @keskitalo was maybe intending to specify a smaller set of detectors as a trait or somewhere else?  Again, this is fine and happy to merge, but the original check for a detector being in a detset containing all detectors seems redundant.

giuspugl:
>it seemed like @keskitalo was maybe intending to specify a smaller set of detectors as a trait or somewhere else? 
then i am a bit confused, isn't enough to specify this with `self.focalplane_key` https://github.com/hpc4cmb/toast/blob/0ca8204cb7a5087bcb881f876ebe11c6caa6d270/src/toast/ops/common_mode_noise.py#L46 ?

keskitalo:
Thank you for the PR. The `detset` trait is indeed missing from the operator. The idea for the operator was this:

1. If `focalplane_key` is set, then create common modes between detectors that share that key
2. If there is no `focalplane_key`, then the default is to create common modes between all detectors
3. If there is no `focalplane_key` but the user provided `detset`, then add a common mode only between the specified detectors.

This is not captured in the doc strings because I forgot to add the `detset` trait. Let me edit the branch directly to add the trait.

Thanks for the unit test!

keskitalo:
@giuspugl Please have a look at the changes I just pushed.

keskitalo (approved):

PR #668, opened 2023-05-20
fix np.bool to np.bool_

ickc:
command used: `find -iname '*.py' -exec sed -i -E 's/np\.bool([^_])/bool\1/g' {} +`

This is fixing a failure in TOAST test "test_binned (toast.tests.binned.BinnedTest)" with the following error with newer versions of Numpy:

AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

### DIFF
--- a/src/toast/fod/noise_estimation.py
+++ b/src/toast/fod/noise_estimation.py
@@ -438,7 +438,7 @@ class OpNoiseEstim:
 
                 psdvalues = np.array([x[col] for x in all_psds])
                 smooth_values = scipy.signal.medfilt(psdvalues, 11)
-                good = np.ones(psdvalues.size, dtype=np.bool)
+                good = np.ones(psdvalues.size, dtype=bool)
                 good[psdvalues == 0] = False
 
                 for i in range(10):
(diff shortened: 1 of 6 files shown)


keskitalo:
I would prefer replacing `np.bool->bool` but this branch will soon be obsolete so I don't mind the proposed solution.

tskisner:
Since this is just running a regex on the source, can you replace `np.bool_` with just `bool`?  The main development branch (toast3) already has this change.  The current master branch is in a legacy maintenance mode.  It will become the `2.x` branch and the toast3 branch will become the `main` branch and the new default.  This is just waiting on the documentation PR (and the current flurry of other PRs).

ickc:
Thanks! I just changed it to `bool`.

PR #671, opened 2023-05-24
JAX: recycling memory

nestordemeure:
Introduced code to recycle memory when doing a reshaping operation with `change_detectors` with JAX gpu data.

Previously one had to allocate a new JAX array to deal with the shape change.


tskisner (approved):
This only touches the jax codepath, so it seems fine to merge.

PR #672, opened 2023-06-06
fix more numpy dtypes

ickc:
c.f. #668


tskisner:
The failing unit test is due to an unrelated issue that has been fixed in the `toast3` branch.  Merging this now because it helps this branch continue to function a little while longer.

PR #673, opened 2023-06-21
JAX bugfix

nestordemeure:
* Fixes a bug with the number of arguments in the JAx implementation of `stokes_weight`
* Updated the corresponding test to catch the error when running in GPU mode


src/toast/ops/stokes_weights/kernels_jax.py
```diff
@@ -191,18 +133,18 @@ def stokes_weights_IQU_interval(
...
-    IAU_sign = 1.0 - 2.0 * IAU
-
+    # Are we using a half wave plate?
     if hwp.size == 0:
-        # Not using a half wave plate
-        new_weights_interval = stokes_weights_IQU_inner(
-            epsilon, cal, gamma, quats_interval, IAU_sign
-        )
+        # No half wave plate
+        hwp_interval = jnp.zeros_like(intervals.indices)
```
tskisner, line 139: Are we sure that this allocation will be optimized away inside the `inner` function?

src/toast/tests/ops_stokes_weights.py
```diff
@@ -48,15 +48,15 @@ def create_data(self, I_sky, Q_sky, U_sky, hwp):
...
-        pixels.apply(data)
         weights = ops.StokesWeights(
             mode="IQU",
             hwp_angle=hwp_name,
             detector_pointing=detpointing,
             fp_gamma="gamma",
             IAU=False,
         )
-        weights.apply(data)
+        pipe = ops.Pipeline(operators=[pixels, weights])
```
tskisner, line 58: Thanks for this change, which will now test the stokes weights kernel if GPU is enabled.

tskisner (approved):
This looks fine- maybe run format_source.sh to clean up the missing newline at the end of a file.

src/toast/ops/stokes_weights/kernels_jax.py
nestordemeure: Yes, it is optimized out by the compiler.

PR #674, opened 2023-06-27
Fix the simulated focalplane layout spacing.

tskisner:
* Clarify that the angular "width" corresponds to the spacing of extreme pixel location along a specific axis.

* Fix off-by-one to match this specification.

* Add pixel centers to unit test plots to verify correct layout.


tskisner:
Just some plots from the unit tests, where the "width" of the fake focalplane is given as 2 degrees.  You can see from the centers labeled on the plot that the width between extreme pixel locations is indeed 2 degrees.
fake_hex_xieta pdf
fake_rhombihex_xieta pdf

src/toast/instrument_sim.py
```diff
@@ -230,16 +234,14 @@ def hex_layout(npos, angwidth, prefix, suffix, pol, center=None, pos_offset=0):
     rtthree = np.sqrt(3.0)
     rtthreebytwo = 0.5 * rtthree
 
-    # compute the diameter (vertex to vertex width)
-    angdiameter = angwidth.to_value(u.radian) / np.cos(thirty)
+    # The diameter (vertex to vertex width)
+    angdiameter = angwidth.to_value(u.radian)
```
keskitalo, line 238: This looks like a big change. What happened here?

src/toast/instrument_sim.py
```diff
@@ -740,7 +744,8 @@ def fake_rhombihex_focalplane(
     xaxis, yaxis, zaxis = np.eye(3)
 
     # Gap between rhombi as a fraction of the total width
-    gap = 0.1 * width
+    if gap is None:
+        gap = 0.5 * width / int(np.sqrt(n_pix_rhombus) - 1)
```
keskitalo, line 748: Why `int()`? Seems to me that the gap is already quantized by the number of pixels.

keskitalo (approved):

src/toast/instrument_sim.py
tskisner: The issue is that in some downstream code this layout function was being called with the vertex-vertex width, not the width of the "flats".  That is actually the quantity that is easiest to define since it is a whole number of pixel spacings.  So I am bringing all 3 codes into line with that definition.  So yes, this is a change to the definition and cannot be merged until `sotodlib` and `s4sim` have corresponding PRs with any needed changes for consistency.

src/toast/instrument_sim.py
tskisner: Old C-programmer habits die hard :-)  I can remove it, especially since the gap is a float.

PR #676, opened 2023-07-15
Redesigning OpenMP target offload loops

nestordemeure:
Precomputing the maximum interval size to be able to fully collapse the triple loop in most operators leaded to significantly improved computation density on the GPU.

### DIFF
--- a/src/toast/_libtoast/template_offset.cpp
+++ b/src/toast/_libtoast/template_offset.cpp
@@ -205,6 +205,7 @@ for (int64_t iview = 0; iview < n_view; iview++) {
                                 amp_view_off[0 : n_view], \
                                 use_flags)
 {
+                // TODO the paralelism can likely be improved on this function
 #pragma omp target teams distribute collapse(2)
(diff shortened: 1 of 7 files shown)


src/toast/_libtoast/template_offset.cpp
```diff
@@ -182,52 +190,63 @@ void init_template_offset(py::module & m) {
...
-                                    contrib = dev_det_data[d];
+                                    contrib += dev_det_data[d];
                                 }
-                                # pragma omp atomic update
-                                dev_amplitudes[amp] += contrib;
                             }
+
+                            int64_t amp = amp_offset + amp_view_off[iview] +
+                                          (int64_t)(isamp / step_length);
+                            dev_amplitudes[amp] += contrib;
```
tskisner, line 249: The threaded reduction to the team-local `contrib` variable looks fine, and for this particular template, the contribution to an amplitude is localized to just this range of samples from one detector.  So no need for atomic update.  Great!

tskisner (approved):
Looks good, I'll update #677 after this is merged.

PR #680, opened 2023-07-21
make entry points callable as python modules

ickc:
such that `python -m toast.scripts.toast_ground_schedule -h` would work or we could do
```py
from toast.scripts.toast_ground_schedule import main
main(...)
```


tskisner:
This change seems good.  When building wheels through the `setup.py` wrapper that calls cmake, [these entry points are already installed](https://github.com/hpc4cmb/toast/blob/fd03af033b8ae6ad269db2cbea4a7038cb4f4c2a/setup.py#L287-L303).  However, when building directly with cmake, they are not installed.  This small change brings the cmake install into line with the setup.py install.

PR #681, opened 2023-07-25
Enable wheel tests prior to tag

tskisner:


tskisner:
This PR has many changed files, but that is almost entirely due to bumping the versions of the bundled googletest and pybind11 source.  The remaining changes were simply build system and wheel-building updates.  Merging now before tagging.

PR #682, opened 2023-07-25
Small perf improvement for the JAX Healpix kernel

nestordemeure:
Small perf improvement for the JAX Healpix kernel (get speedup to expected value when running in non-hybrid).


tskisner (approved):
Ok, thanks for the improvements!

PR #687, opened 2023-08-07
Add compress_detdata to save_hdf5 operator

keskitalo:
Make the user-facing interface to hdf5 compression more intuitive.

### DIFF
--- a/src/toast/io/compression_flac.py
+++ b/src/toast/io/compression_flac.py
@@ -46,7 +46,11 @@ def float2int(data, quanta=None, precision=None):
     min_quanta = amp / max_flac
     if precision is not None:
         rms = np.std(data)
-        quanta = rms / precision
+        quanta = rms / 10**precision
+        if quanta < min_quanta:
+            msg = f"Set precision (={precision}) cannot be supported with 32bit FLAC: "
+            msg += f"{quanta} < {min_quanta}"
+            raise RuntimeError(msg)
     if quanta is None:
         quanta = min_quanta
     if quanta == 0:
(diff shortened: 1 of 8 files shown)


src/toast/io/compression_flac.py
```diff
@@ -25,21 +25,34 @@ def float2int(data, quanta=None):
...
+        raise RuntimeError("Cannot convert data with NaNs to integers")
     dmin = np.amin(data)
     dmax = np.amax(data)
     offset = 0.5 * (dmin + dmax)
     amp = 1.01 * max(np.abs(dmin - offset), np.abs(dmax - offset))
+    # Use the full bit range of int32 FLAC.  Actually we lose one bit
+    # Due to internal FLAC implementation.
+    max_flac = np.iinfo(np.int32).max // 2
+    min_quanta = amp / max_flac
+    if precision is not None:
```
tskisner, line 47: It is great to have this feature!  Thanks for adding it.

src/toast/ops/mapmaker_utils/kernels.py
```diff
@@ -92,7 +92,7 @@ def build_noise_weighted(
         weight_index,
         weights,
         data_index,
-        det_data,
+        det_data.astype(np.float64),
```
tskisner, line 95: obviously no harm here, though we could also enforce the up-cast during HDF5 loading.  We just have to remember to find all these places again once we have the compiled functions templated over the types.

tskisner (approved):
Just one minor comment that is not critical.  Thanks for adding support for fixed precision and the additional tests!

PR #688, opened 2023-08-08
Updating JAX

nestordemeure:
replaced DeviceArray with Jax.Array to comply with the [jax.Array migration](https://jax.readthedocs.io/en/latest/jax_array_migration.html#whats-going-on).


tskisner (approved):
Thanks!

PR #689, opened 2023-08-09
Update required language standard to C++14

tskisner:
Our bundled copy of the google test framework now requires C++14. We could relax this in the future if needed once we reorganize the compiled extension and ensure that all features are tested from the python interface.  In that scenario, we could then completely remove the compiled test runner and gtest.


tskisner:
Actually, this was not a problem with the language standard.  The bundled gtest is the last release that still supports C++11.  The issue was that an external (newer) gtest was installed in my testing conda environment, and the headers from that were being picked up instead of the bundled headers.  So if anything, it is a problem with our header search order.  Closing this now and will either revisit in the future or remove gtest altogether.

PR #694, opened 2023-08-17
fix broken scheduling mode

keskitalo:
Forgot to add the `verbose` option to Pole schedules. Clearing this out to file another change in the scheduler format.


tskisner (approved):
Looks good, thanks!

PR #695, opened 2023-08-21
Improve flagging and noise estimation

tskisner:
We have two primary flag bits for shared and detector data:  
"invalid" and "processing".  The invalid bit means that the 
data does not exist or is wrong / corrupted in some way that
makes it useless for any application.  The processing bit 
means that the data should not "typically" be used for 
science results.

This work attempts to change all built-in simulation and 
reduction operators to follow these basic guidelines:

1. Operators that load data:
  - Boresight pointing marked as invalid if no pointing 
    solution exists
  - Detector data marked as invalid if it is missing (e.g. 
    dropped packets, etc)
  - In some cases, there may be data that should not 
    routinely be used for science, and these samples should 
    be flagged with the processing bit.

2. Operators that simulate telescope observing:
  - Unless the operator is also simulating failed pointing
    solutions or readout problems, no samples should be flagged
    as invalid.
  - Some samples might be flagged with the processing bit if
    those features are typically known by low-level data
    acquisition systems.  Examples might be cooler cycles or
    scan turnarounds, although those samples might also be
    flagged by downstream operators.

3. Operators that simulate detector timestreams:
  - Existing invalid detector samples should have their flags 
    passed through and remain invalid
  - Detector samples flagged with the processing bit should have
    this bit passed through.
  - If the simulation requires data inputs (e.g. pointing), and
    that data is flagged as invalid, then the output samples of
    the simulation operator should be flagged as invalid.
  - If the simulation uses inputs flagged with the processing
    bit, it is up to the operator whether to use those samples.
  - If the simulation is unable to generate some samples, then
    those samples should be marked as invalid since they do not
    contain the full expected detector response
  - If the simulation generates poor-quality data for some reason,
    those samples might be marked with the processing flag

4. Operators that process detector timestreams:
  - Existing invalid detector samples should have their flags 
    passed through and remain invalid
  - Detector samples flagged with the processing bit should have
    this bit passed through as well.
  - It is up to the operator whether to use samples flagged with
    the processing bit.
  - If the operator is modifying the detector data, and fills
    some samples with garbage, zeros, etc, then those samples
    should be marked as invalid.
  - If the operator is modifying the detector data, and is unable
    to use / process some data, then those samples should have
    the processing bit set.

This PR also include improvements to flagging in the Offset template and a typo fix in fitting noise models.


src/toast/ops/mapmaker_utils/mapmaker_utils.py
```diff
@@ -60,7 +60,8 @@ class BuildHitMap(Operator):
     )
 
     det_flag_mask = Int(
-        defaults.det_mask_invalid, help="Bit mask value for optional detector flagging"
+        defaults.det_mask_proc_or_invalid,
```
keskitalo, line 63: A hit map should not be affected by a processing mask.

src/toast/ops/noise_estimation.py
```diff
@@ -84,9 +85,10 @@ class NoiseEstim(Operator):
         help="Observation detdata key for processing mask flags",
     )
 
-    mask_flag_mask = Int(
-        defaults.det_mask_invalid, help="Bit mask for raising processing mask flags"
-    )
+    # FIXME:  This trait is unused, ok to remove?
```
keskitalo, line 88: This could actually provide the solution to the processing mask vs. processing flags issue. Perhaps we need a separate flag bit for the processing mask rather than lump them with other processing flags?

keskitalo (commented):
We need to work on this PR a little more. There needs to be a distinction between samples that should not be used in template fitting and filtering and samples that are excluded from all processing. Point source flags are a good example of usable but excluded samples. We still want the point sources in the final map. I have flagged completely excluded samples such as turnarounds as "invalid" but now they seem to drift into the processing mask.

tskisner:
Summarizing a discussion from other channels:

### Principles

- Invalid data is a "one way street".  Any operator which modifies data and fails to modify some samples should mark those samples as invalid.
- Default flagging masks across all operators should be self-consistent and "just work" for unit tests and typical usage.
- Operators that raise flag bits should support taking the mask to raise as a trait, to support custom use cases.
- Other default flag bits will be defined to support common data selection criteria.  

### Implementation Plan

- In the default flag mask definitions, make these changes:
  - det_mask_sso set to “4”
  - turnaround set to “2” and changed to “shared_mask_turnaround”
  - elnod set to “4” and changed to “shared_mask_elnod”
  - Similar rename of other ground shared masks
  - Build composite mask aliases:
    - shared_mask_nonscience = invalid | unstable | irregular
    - det_mask_nonscience = invalid | processing | sso
- Go through all operators and update defaults to reflect correct mask (usually invalid or nonscience)
- Go through all operators and ensure that failed data modifications are flagged with invalid.

tskisner:
Ok @keskitalo , this PR is ready for re-review.  There is a small issue when running on GPU with hybrid pipeline data movement, but that will likely just be a typo fix somewhere.  Will not merge until those tests pass (which are not tested by CI anyway).

src/toast/ops/noise_model.py
```diff
@@ -425,6 +431,18 @@ def _fit_log_jac(self, x, *args, **kwargs):
         ) / fmalpha
         return J
 
+    def _get_err_ret(self, psd_unit):
+        eret = dict()
+        eret["fit_result"] = types.SimpleNamespace()
+        eret["fit_result"].success = False
+        # Use a large value, to avoid divide by zero elsewhere
```
keskitalo, line 438: This comment seems to be in conflict with line 440

src/toast/ops/polyfilter/polyfilter.py
```diff
@@ -65,11 +65,14 @@ class PolyFilter2D(Operator):
     )
 
     det_flag_mask = Int(
-        defaults.det_mask_invalid | defaults.det_mask_processing,
+        defaults.det_mask_invalid,
```
keskitalo, line 68: Why were the processing flags dropped here? The polyfilter seems like an ideal application for them.

src/toast/ops/polyfilter/polyfilter.py
```diff
@@ -447,13 +458,13 @@ class PolyFilter(Operator):
...
```
keskitalo, line 461: Processing flags skipped here as well.

src/toast/ops/sss.py
```diff
@@ -97,7 +97,9 @@ def _exec(self, data, detectors=None, **kwargs):
         comm = data.comm.comm_group
 
         for obs in data.obs:
-            dets = obs.select_local_detectors(detectors)
+            dets = obs.select_local_detectors(
+                detectors, flagmask=defaults.det_mask_invalid
```
keskitalo, line 101: Should we add the detector mask as a separate trait here as well?

src/toast/ops/statistics.py
```diff
@@ -88,6 +90,13 @@ def _exec(self, data, detectors=None, **kwargs):
         log = Logger.get()
         nstat = 3  # Variance, Skewness, Kurtosis
 
+        if self.output_dir is not None:
```
keskitalo, line 93: Why use `comm_world` here? Is it guaranteed that all process groups execute the operator at the same time?

src/toast/templates/fourier2d.py
```diff
@@ -124,6 +124,8 @@ def _initialize(self, new_data):
 
             # Build up detector list
             for d in ob.local_detectors:
+                if ob.local_detector_flags[d] & self.det_flag_mask:
```
keskitalo, line 127: Could we not use `ob.select_local_detectors()` here?

src/toast/templates/fourier2d.py
```diff
@@ -265,6 +267,8 @@ def evaluate_template(theta, phi, radius):
                 norms_view = self._norms[norm_slice].reshape((-1, self._nmode))
 
                 for det in ob.local_detectors:
+                    if det not in self._all_dets:
```
keskitalo, line 270: Another place to use select_local_detectors()

src/toast/templates/gaintemplate.py
```diff
@@ -65,9 +65,11 @@ def _initialize(self, new_data):
         # but sorted in order of occurrence.
         all_dets = OrderedDict()
 
+        # Build up detector list
         for iob, ob in enumerate(new_data.obs):
-            # Build up detector list
             for d in ob.local_detectors:
+                if ob.local_detector_flags[d] & self.det_flag_mask:
```
keskitalo, line 71: Another place to use select_local_detectors()

src/toast/templates/subharmonic.py
```diff
@@ -58,9 +58,11 @@ def _initialize(self, new_data):
         # but sorted in order of occurrence.
         all_dets = OrderedDict()
 
+        # Build up detector list
         for iob, ob in enumerate(new_data.obs):
-            # Build up detector list
             for d in ob.local_detectors:
+                if ob.local_detector_flags[d] & self.det_flag_mask:
```
keskitalo, line 64: Another place to use select_local_detectors()

keskitalo (approved):
Mostly small observations.

The one big question is about `det_flag_mask` being repurposed and whether that is general enough.

src/toast/_libtoast/ops_filterbin.cpp
```diff
@@ -311,6 +311,87 @@ void build_template_covariance(std::vector <int64_t> & starts,
     }
 }
 
+// void build_template_covariance(
```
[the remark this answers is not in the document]
tskisner: When investigating a test failure of the FilterBin operator, I thought perhaps it was due to the way that data was being cast through the pybind11 interface.  So I wrote an implementation using our buffer extraction / dimension checking helpers.  This turned out not to be the problem, but I left that new code in the file (commented out) for when we port these functions into GPU kernels.  I can leave a comment to that effect- sorry for the confusion.

src/toast/ops/cadence_map.py
```diff
@@ -154,7 +155,9 @@ def _exec(self, data, detectors=None, **kwargs):
             buf[:, :] = False
             for obs in data.obs:
                 obs_data = data.select(obs_uid=obs.uid)
-                dets = obs.select_local_detectors(detectors)
+                dets = obs.select_local_detectors(
+                    detectors, flagmask=self.det_flag_mask
```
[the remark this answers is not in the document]
tskisner: That is an interesting point- for our default simple cases it is fine, since we typically have "invalid" or "nonscience" and nonscience is a superset that includes the invalid bit.  However I guess we should support different masks for per-detector and per-sample flags.  Maybe `det_mask` and `det_flag_mask`?

src/toast/ops/demodulation.py
```diff
@@ -192,21 +192,44 @@ def _exec(self, data, detectors=None, **kwargs):
...
         for obs in demodulate_obs:
-            dets = obs.select_local_detectors(detectors)
+            # Get the detectors which are not cut with per-detector flags
+            local_dets = obs.select_local_detectors(
+                detectors, flagmask=self.det_flag_mask
+            )
+            # print(f"DEMOD Proc {obs.comm.group_rank}: local {local_dets}")
+            if obs.comm.comm_group is None:
+                all_dets = local_dets
+                # print(f"DEMOD Proc {obs.comm.group_rank}: comm_group is None")
```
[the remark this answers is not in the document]
tskisner: Good catch, will fix that.

src/toast/ops/demodulation.py
```diff
@@ -634,7 +687,7 @@ class StokesWeightsDemod(Operator):
     )
 
     weights = Unicode(
-        defaults.weights, help="Observation detdata key for output weights"
+        f"{defaults.weights}_demod", help="Observation detdata key for output weights"
```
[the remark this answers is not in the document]
tskisner: While testing the demodulation I discovered a case where having the same name as the pre-demodulation weights caused a problem.  Unfortunately I don't recall the details, so I can try switching it back and see what breaks.

src/toast/ops/polyfilter/polyfilter.py
tskisner: Since a filter modifies the data, any samples that are **not** modified become invalid (assuming the notes earlier in this PR discussion).  So if a filter uses the processing flag, then on output those unfiltered samples will be flagged invalid.  (Thinking out loud) I guess this makes sense.  If you are applying a filter that does not make sense to apply across a turnaround (for example), then you really are never going to use the turnarounds later and it makes sense that they are invalid.  If we have an operator applying some other filter to an entire observation (for example timeconstant deconvolution), then the default would be to consider all valid data.  So this choice of defaults is operator-dependent, and I agree for the polynomial filters it makes sense to include the processing mask.

src/toast/ops/cadence_map.py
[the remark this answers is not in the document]
tskisner: This is now implemented throughout the code base.  The `det_mask` trait describes the bit mask applied to per-detector flags and the `det_flag_mask` applies to per-sample flags.

src/toast/ops/polyfilter/polyfilter.py
tskisner: I have gone through every operator and tried to set the defaults to sensible ones:

- if an operator is simulating data, it only uses the `invalid` mask and tries to fill as many samples as possible
- if an operator is doing low-level filtering of data that is independent of any sky / processing mask, then it uses the `invalid` bitmask.
- if an operator is doing higher-level filtering, then it also uses the `processing` bit.
- if an operator is doing high-level mapmaking operations, it uses the `nonscience` compound mask.

Obviously these are just the defaults, and can be overridden on a case by case basis.  For example when making planet maps you would not set the SSO bit.

src/toast/ops/polyfilter/polyfilter.py
tskisner: Fixed

src/toast/ops/sss.py
tskisner: Done for this and all operators

src/toast/ops/statistics.py
tskisner: Fixed

src/toast/templates/fourier2d.py
tskisner: I think I have now caught all of these and fixed them.

src/toast/ops/demodulation.py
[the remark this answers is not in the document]
tskisner: I tried to reproduce my earlier problem, but could not.  In principle, the previously created stokes weights will get deleted and recreated if the demodulated weights have the same name.  This would only be a problem if you wanted to use the pre-demodulated weights for something later.  I have reverted this change.  However, note that all of our workflow tools currently get the name of the weight directly from the operator, so you could change the name to anything and no downstream operators would be affected.

keskitalo (approved):
Looks great!

src/toast/templates/offset/offset.py
```diff
@@ -181,24 +185,26 @@ def _initialize(self, new_data):
...
                     continue
+                self._obs_dets[iob].add(d)
                 if d not in all_dets:
                     all_dets[d] = None
 
         self._all_dets = list(all_dets.keys())
 
         # Go through the data one local detector at a time and compute the offsets into
-        # the amplitudes.
+        # the amplitudes.  Also save the detectors in use for
```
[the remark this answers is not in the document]
tskisner: Fixed

PR #696, opened 2023-08-30
Work on refactoring config files and build scripts

tskisner:


src/toast/_libtoast/ops_pixels_healpix.cpp
```diff
@@ -11,14 +11,9 @@
...
 
-// 2 * Pi
-const double TWOPI = 2.0 * M_PI;
 
 // 2/3
-const double TWOTHIRDS = 2.0 / 3.0;
-
-const int64_t hpix_jr[] = {2, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4};
-const int64_t hpix_jp[] = {1, 3, 5, 7, 0, 2, 4, 6, 1, 3, 5, 7};
+#define TWOTHIRDS 0.66666666666666666667
```
nestordemeure, line 16: why not a const double?

nestordemeure (approved):
Looks good.

Long term we might want to investigate existing package managers tailored for Python in order to reduce the amount of install scripts we need to write.

Do the changes to pixel healpix and tod filter need to be ported to the JAX and Numpy versions of the kernels?

src/toast/_libtoast/ops_pixels_healpix.cpp
tskisner: Defining static const variables here gave compile errors from nvc++ when used inside device functions.  LLVM was ok with them.  I reverted to a preprocessor macro for now, although perhaps it is silly to avoid a single floating point division in the bigger picture of the device kernels.

PR #698, opened 2023-09-12
Fix flagging of noise model fits

tskisner:
This extracts just the bug fixes from #695, and can hopefully be merged more quickly.

### DIFF
--- a/src/toast/ops/noise_model.py
+++ b/src/toast/ops/noise_model.py
@@ -130,6 +130,15 @@ class FitNoiseModel(Operator):
         None, allow_none=True, help="Create a new noise model with this name"
     )
 
+    det_flag_mask = Int(
+        defaults.det_mask_invalid,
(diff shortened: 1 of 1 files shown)


src/toast/ops/noise_model.py
```diff
@@ -407,6 +424,16 @@ def _fit_log_jac(self, x, *args, **kwargs):
             f_alpha * log_freqs + fmin_alpha * np.log(fmin)
         ) / fmalpha
         return J
+    
+    def _get_err_ret(self, psd_unit):
```
keskitalo, line 428: Could use a comment to explain what it does :)

keskitalo (approved):
Looks fine.

src/toast/ops/noise_model.py
tskisner: Added...

PR #700, opened 2023-09-18
Unwrap the HWP angle before perturbing and guard against stepped HWP

keskitalo:
Handle situations where the HWP angle is not continuous.


sbiquard:
I would only say that we may want to take the principal value of the angle after perturbing it, ie. take new_angle % 2 pi

tskisner (approved):
Looks good, thanks for the test stub!

PR #701, opened 2023-09-19
Add a unit test for a corner case met in DC0

keskitalo:


keskitalo:
@tskisner Can you add a fix for this failing unit test?

tskisner:
This pull request has greatly expanded from its original scope, and it now also includes fixes for the pointing matrix weights.  The unit tests have been expanded to include plots for the detector response across multiple HWP and fixed sky Stokes components.  I'll rebase to clean up the commit log before merging, but I believe everything is ready @keskitalo .  The documentation of the pointing weights is now added to the `toast3_docs` branch where the documentation overhaul is ongoing.  Attached is PDF of that notebook, which includes panel plots from the new unit tests.

[pointing_weights.pdf](https://github.com/hpc4cmb/toast/files/12841819/pointing_weights.pdf)

src/toast/_libtoast/ops_pixels_healpix.cpp
```diff
@@ -1,39 +1,62 @@
...
 
+#include <module.hpp>
 #include <qarray.hpp>
-
 #include <intervals.hpp>
-
 #include <accelerator.hpp>
 
+// PI
+#define PI 3.14159265358979323846
```
keskitalo, line 14: This seems a little dated, given that mathematical constants are available through standard libraries but since it works I would not worry about it.

src/toast/_libtoast/ops_pixels_healpix.cpp
```diff
@@ -1,39 +1,62 @@
...
         utab[m] = (m & 0x1) | ((m & 0x2) << 1) | ((m & 0x4) << 2) |
                   ((m & 0x8) << 3) | ((m & 0x10) << 4) | ((m & 0x20) << 5) |
                   ((m & 0x40) << 6) | ((m & 0x80) << 7);
     }
     return;
 }
 
+void hpix_init_ctab(int64_t * ctab) {
+    for (int64_t m = 0; m < 0x100; ++m) {
+        ctab[m] = (m & 0x1) | ((m & 0x2) << 7) | ((m & 0x4) >> 1) |
```
keskitalo, line 41: "Hacker's delight" is strong in this one :)

src/toast/_libtoast/ops_stokes_weights.cpp
```diff
@@ -56,83 +59,83 @@ void stokes_weights_alpha(
...
     int32_t w_indx = weight_index[idet];
-    int64_t off = (q_indx * 4 * n_samp) + 4 * isamp;
+    int64_t qoff = (q_indx * 4 * n_samp) + 4 * isamp;
 
     double alpha;
-    stokes_weights_alpha(&(quats[off]), &alpha);
+    stokes_weights_alpha(&(quats[qoff]), alpha);
 
-    double ang = 2.0 * (alpha + 2.0 * (hwp[isamp] - gamma[idet]));
+    double ang = 2.0 * (2.0 * (gamma[idet] - hwp[isamp]) - alpha);
```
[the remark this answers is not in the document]
tskisner: The presence of the HWP does introduce a change in response, even if the fast axis is aligned with the detector orientation.  I know the `pointing_weights.pdf` linked above in this thread is "dense", but maybe skip to the bottom and examine the plots of expected detector response (generated by the unit tests in this PR).  Basically the code implements the math in that document, so if something is wrong we should fix it in both.  We can chat more later today.

src/toast/_libtoast/ops_pixels_healpix.cpp
tskisner: True, I could include `cmath` for some of them- that would be a small change.  Although there may still be some differences between systems.  [Looks like C++20 finally adds them](https://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c), so someday it will be easier.

src/toast/_libtoast/ops_pixels_healpix.cpp
tskisner: That code has been with us since we extended our original healpix implementation from 32bit to 64bit.  No way we should touch it :-)

src/toast/_libtoast/ops_stokes_weights.cpp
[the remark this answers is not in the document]
tskisner: Discussed offline...

src/toast/_libtoast/ops_pixels_healpix.cpp
tskisner: I will make this change in #702 instead.

tskisner:
This work was merged in #702.  Closing this.

PR #702, opened 2023-10-09
Pointing Matrix Cleanup

tskisner:
Refactor healpix and stokes weights operations:
    
- Move all current bundled healpix ops into the same source file as the PixelsHealpix code so that all device kernels are in the same compilation unit as the python wrappers.
    
- Add full unit test coverage for all bundled operations, comparing to the equivalent functions in healpy.
    
- Remove dead versions of the code from the deprecated libtoast internal library.
    
- Remove dead versions of the stokes weights code from the libtoast library and call the modern kernels in all unit tests.

- In the bundled ang to pix operations, perform range reduction to 0-2PI and also round phi angles to zero if they are close to machine precision from zero.  This fixes the issue that was the original motivation for this PR.
    
- Update Stokes pointing weights to match latest document on polarization response.
    
- Remove stale versions of the stokes weights code and move all helper functions closer to the relevant operator kernel.
    
- Fix the use of the "cal" trait of the StokesWeights operator which was designed to be the name of the observation key to use for a detector calibration dictionary.  This change propagated into the kernel code.
    
Additionally, add a small fix to the spt3g unit tests.


tskisner:
Supercedes #701

tskisner:
Since #701 will be deleted without merging, just wanted to attach the pointing weights document here for reference (the source notebook is in the new docs branch).

[pointing_weights.pdf](https://github.com/hpc4cmb/toast/files/12851110/pointing_weights.pdf)

tskisner:
This PR is the restoration of #701, which was already approved.  Merging now.

PR #704, opened 2023-10-12
Stokes_weights fixes

nestordemeure:
Fixed:
* shape and use of `cal` in the `stokes_weights` kernels since it is now an array of size `ndets`,
* formula to compute `ang` in the Numpy and JAX `stokes_weights_IQU` kernels (ported the updated C++ one).


tskisner (approved):
Thanks for fixing these remaining things that were missed in #702 !

PR #706, opened 2023-10-16
Make polynomial regression more robust

keskitalo:
Existing implementation of the polynomial filter is fragile. It was using SVD to overcome singular template covariance matrix and applied an arbitrary rcond cut to small singular values. When small singular values were encountered, the results were compromised.

The new implementation accepts singular values down to machine precision and does away with the squaring of the template matrix. The latter greatly improved the condition number of the system.

### DIFF
--- a/src/libtoast/src/toast_tod_filter.cpp
+++ b/src/libtoast/src/toast_tod_filter.cpp
@@ -5,6 +5,7 @@
 
 #include <string.h>
 #include <algorithm>
+#include <sstream>
 
 #ifdef _OPENMP
 # include <omp.h>
@@ -34,7 +35,7 @@ void toast::filter_polynomial(int64_t order, size_t n, uint8_t * flags,
     double fone = 1.0;
 
     #pragma \
-    omp parallel for schedule(static) shared(order, signals, flags, n, nsignal, starts, stops, nscan, norder, upper, lower, notrans, trans, fzero, fone)
+    omp parallel for schedule(static) default(none) shared(order, signals, flags, n, nsignal, starts, stops, nscan, norder, upper, lower, notrans, trans, fzero, fone)
     for (size_t iscan = 0; iscan < nscan; ++iscan) {
         int64_t start = starts[iscan];
         int64_t stop = stops[iscan];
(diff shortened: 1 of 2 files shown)


src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -161,6 +134,11 @@ void toast::filter_polynomial(int64_t order, size_t n, uint8_t * flags,
                 }
             }
         }
+
+        if (info != 0) {
+            std::cerr << "ERROR: DGELLS info = " << info << std::endl;
+        }
```
tskisner, line 140: Do we want to remove this debugging statement?  Or use the toast logging to make an error message and raise an exception pointing to the line number?  Like this:
```
if (info != 0) {
    auto log = toast::Logger::get();
    std::ostringstream o;
    o << "DGELLS:  " << ngood << "/" << scanlen << " good samples, order " << norder;
    o << " failed with info " << info;
    log.error(o.str().c_str(), TOAST_HERE());
    throw std::runtime_error(o.str().c_str());
}
```

src/libtoast/src/toast_tod_filter.cpp
```diff
@@ -34,7 +34,7 @@ void toast::filter_polynomial(int64_t order, size_t n, uint8_t * flags,
     double fone = 1.0;
 
     #pragma \
-    omp parallel for schedule(static) default(none) shared(order, signals, flags, n, nsignal, starts, stops, nscan, norder, upper, lower, notrans, trans, fzero, fone)
```
tskisner, line 37: Maybe use default(shared)?

tskisner (approved):
Thanks for this cleanup / testing.

src/libtoast/src/toast_tod_filter.cpp
keskitalo: logging and throwing is much better. I adopted that.

src/libtoast/src/toast_tod_filter.cpp
keskitalo: I prefer not to use defaults. This way every shared variable is an explicit choice. Fewer surprises that way.

tskisner: But that is exactly what `default(none)` (now removed) does.  If no `default` is specified, then all unspecified variables are shared.  [One reference](https://rookiehpc.org/openmp/docs/default/index.html) and [another](https://www.openmp.org/spec-html/5.2/openmpsu35.html).

keskitalo: Correct. I removed default(none) temporarily during development and forgot to add it back. It is now restored.

PR #712, opened 2023-10-26
Az-tracking raster scans

keskitalo:


keskitalo:
Left, DC0 SPLAT hit map without azimuth tracking. Right, with the tracking enabled:
image

tskisner (approved):
These changes look fine, and it is a nice feature- assuming that a telescope and its control software were capable of scheduling the scans in this way.

keskitalo:
It is the mode SPT presently scans so there is precedent.

I had to tweak the CES simulation code to accommodate this new mode. Comparing to an earlier raster scan hit map (both without az tracking) I saw cases where one detector sample had shifted between neighboring sky pixels.  Seems like a small effect but it means that the simulated boresight motion is ever so slightly changed by this PR.

PR #713, opened 2023-10-31
Support passing 'None' string in argument parsing

tskisner:
Add a custom argparse action to support setting traits to None.

### DIFF
--- a/src/toast/widgets.py
+++ b/src/toast/widgets.py
@@ -272,7 +272,7 @@ class ObservationWidget(object):
             )
             with w_plot:
                 clear_output(wait=True)
-                fig.show_dash(mode="inline", host="localhost")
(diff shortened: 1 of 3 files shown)


src/toast/widgets.py
```diff
@@ -272,7 +272,7 @@ def response(change):
             )
             with w_plot:
                 clear_output(wait=True)
-                fig.show_dash(mode="inline", host="localhost")
+                fig.show_dash(mode="inline_persistent", host="localhost")
```
keskitalo, line 275: This does not seem related but that is fine.

keskitalo (approved):
Awesome!

src/toast/widgets.py
tskisner: yes, a small tweak to the interactive display widget, and unrelated.

PR #714, opened 2023-11-06
Update perlmutter configs for omp target offload

tskisner:
Add content to the platforms/README file documenting install on perlmutter with LLVM and NVHPC.


nestordemeure (approved):
Looks great!

PR #717, opened 2023-12-05
Jax interval map

nestordemeure:
Introduced `imap`, an extension of `xmap` able to deal with JAX (variable size) intervals. It simplifies the code used to write the intervals.

TODO:
- rerun all tests with JAX on,
- remove previous interval type from the codebase,
- clean-up implementation details.


tskisner (approved):
Thanks for working to simplify the boilerplate.  I did not look in great detail at the imap internals...

PR #718, opened 2023-12-06
Correct bug in the creation of the WCS

gabrielecoppi:
I solved a couple of bugs in the creation of the WCS, in particular on the `bounds` creation. I run the tests and everything seems OK. 

image


tskisner:
Thanks for the help on this!

PR #722, opened 2023-12-14
better timing

keskitalo:
Made a few helpful edits in the process of tracking down a slow part of the code.


tskisner (approved):
Thanks Reijo, looks good.

PR #723, opened 2023-12-21
Tweaks 12 2023

keskitalo:
Small tweaks that didn't deserve a separate PR:
- add a short circuit for empty detector list in demodulation
- fix atmospheric snapshot output for debugging
- noise estimator will automatically create the output directory if it does not exist
and some bigger items:
- the three concentric atmospheric cones are now scaled to fit the specified `zmax` so we don't simulate useless volume elements. This change means that old and new atmospheric simulations will be incompatible.
- Volume element size (`xstep`, `ystep`, `zstep`) is now defined in terms of the largest, not the smallest concentric cone. This is more intuitive since it is the same length scale as the total simulated volume.
- fixed broken snapshot plot generation in atmospheric simulation

### DIFF
--- a/src/toast/ops/demodulation.py
+++ b/src/toast/ops/demodulation.py
@@ -731,6 +731,8 @@ class StokesWeightsDemod(Operator):
 
         for obs in data.obs:
             dets = obs.select_local_detectors(detectors)
+            if len(dets) == 0:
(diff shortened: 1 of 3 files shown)


tskisner (approved):
Looks good!

PR #728, opened 2024-01-24
Fix parallelization problems in noise estimation

keskitalo:
Noise estimation code struggled with observations composed of a number of short intervals, especially when `lagmax` was much longer than individual window.

Furthermore, the data redistribution that is used to allow arbitrary crosscorrelators to be evaluated was affecting the results.

This PR addresses both of those concerns. Furthermore, it removes the unnecessary detrending step what was redundant with the highpass filter already being applied.

Finally, noise estimation is now performed independently in each interval. Earlier, the entire observation was analyzed together with the gaps between intervals simply flagged.  This way there is no risk of sample pairs spanning two intervals.


tskisner (approved):
Thanks for all this work.  I tested the latest commit on this branch on 16 processes with realistic flagging and confirmed everything works as expected.

PR #730, opened 2024-01-25
Minor fixes to noise estimation work

tskisner:
Some small fixes, discovered with unit tests and batch jobs with 8 or 16 processes per group:

* When duplicating an observation, also duplicate per-detector flags.

* When redistributing an observation, reset per-detector flags before setting.

* Make the common mode removal an option (default True) prior to noise estimation.  This is useful if the data has already had the common mode removed prior to this operator.

* When high-pass filtering data that is distributed by time slices there may be some chunks that are fully flagged.  This should not be an error.  Instead, the flagged timestream is set to zero.

* When communicating overlaps, ensure that the span used (lagmax
  + half_average) is less than the number of samples.

### DIFF
--- a/src/toast/observation.py
+++ b/src/toast/observation.py
@@ -656,6 +656,7 @@ class Observation(MutableMapping):
             sample_sets=self.all_sample_sets,
             process_rows=self.dist.process_rows,
         )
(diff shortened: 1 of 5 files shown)


keskitalo (approved):
Looks good!

PR #732, opened 2024-01-30
Remove unnecessary flag bits

keskitalo:
Resolve a conflict in default shared flag bits by removing unnecessary flag masks


tskisner (approved):
Looks good, thanks for solving the mask bit conflict!

PR #734, opened 2024-02-01
When parsing config files, handle compatible types

tskisner:
Due to the way we are writing TOML and YAML files for user-friendly editing, the true type of traits loaded from those formats can be ambiguous.  This work adds checks to allow assignment from compatible types when merging the values loaded from config files with the current state of trait properties.


keskitalo (approved):
Awesome!

PR #735, opened 2024-02-05
Observation matrix class

keskitalo:
Implement an observation matrix class to simplify use of observation matrices.


tskisner (approved):
Sorry this slipped off my radar, looks fine.

PR #736, opened 2024-02-06
Implement a mapmaking template for HWP synchronous signal

tskisner:
- Implement truncated Fourier series technique based on C++ code developed for real-time filtering in kst in 2009.

- Add unit tests for low-level math functions and a synthetic scan strategy without atmosphere.

- Format source with latest version of black (24.1.1)

### DIFF
--- a/src/toast/templates/hwpss.py
+++ b/src/toast/templates/hwpss.py
@@ -1,0 +1,534 @@
+# Copyright (c) 2023-2024 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+import ast
+import json
+from collections import OrderedDict
+
+import h5py
+import numpy as np
+from scipy.linalg import lu_factor, lu_solve
+
+from ..data import Data
+from ..mpi import MPI
+from ..observation import default_values as defaults
+from ..timing import function_timer
+from ..traits import Bool, Float, Instance, Int, Unicode, trait_docs
+from ..utils import Logger
+from ..vis import set_matplotlib_backend
+from .amplitudes import Amplitudes
+from .template import Template
+
+
+@trait_docs
+class Hwpss(Template):
+    """This template represents the HWP synchronous signal."""
+
+    # Notes:  The TraitConfig base class defines a "name" attribute.  The Template
+    # class (derived from TraitConfig) defines the following traits already:
+    #    data             : The Data instance we are working with
+    #    view             : The timestream view we are using
+    #    det_data         : The detector data key with the timestreams
+    #    det_data_units   : The units of the detector data
+    #    det_mask         : Bitmask for per-detector flagging
+    #    det_flags        : Optional detector solver flags
+    #    det_flag_mask    : Bit mask for detector solver flags
+    #
+
+    hwp_angle = Unicode(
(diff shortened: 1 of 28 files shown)


tskisner:
An example of one of the plots generated by unit tests, showing the removal of HWPSS as part of the mapmaking:
hwpss_test

src/toast/templates/hwpss.py
```diff
@@ -0,0 +1,534 @@
...
+                cov[4 * hr + 3, 4 * hc + 3] = np.dot(
+                    np.multiply(times[good], sincos[good, 2 * hr + 1]),
+                    np.multiply(times[good], sincos[good, 2 * hc + 1]),
+                )
+
+        # Fill in lower triangle
+        for hr in range(0, 4 * n_harmonics):
+            for hc in range(0, hr):
+                cov[hr, hc] = cov[hc, hr]
+        # LU factorization for later solve
```
keskitalo, line 243: Should we check the condition number for pathological systems?

src/toast/tests/template_hwpss.py
```diff
@@ -0,0 +1,956 @@
+# Copyright (c) 2015-2023 by the parties listed in the AUTHORS file.
```
keskitalo, line 1: Copyright year is out of date

src/toast/tests/template_hwpss.py
```diff
@@ -0,0 +1,956 @@
...
+        ops.Combine(
+            op="add",
+            first="noise",
+            second="input",
+            result="input",
+        ).apply(data)
+
+        return data
+
+    def create_ground_sim(self, outdir, width, sky_proj, sky_res):
```
keskitalo, line 411: Don't we already have helper functions for this?

keskitalo (approved):
Looks great! We need a better system for running format_source.sh between PRs. I almost choked on my coffee seeing that 23 files had been changed :)

src/toast/tests/template_hwpss.py
tskisner: We have helper functions to simulate a small observing schedule (which is called by this function), but we do not have a higher level function that simulates all the components (since that might be different between tests).  We could extract this in the future though if we want to use it in multiple tests.

src/toast/tests/template_hwpss.py
tskisner: Fixed.

src/toast/templates/hwpss.py
tskisner: Done- for observations where this covariance is poorly conditioned, the template amplitude are flagged.

src/toast/templates/hwpss.py
```diff
@@ -0,0 +1,534 @@
...
+        help="If not None, make debugging plots in this directory",
+    )
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def _initialize(self, new_data):
+        log = Logger.get()
+
+        # For each harmonic with the sin, cos, and time drift terms for each.
```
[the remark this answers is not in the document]
tskisner: In the case where the time drift in the data is zero, I don't think the extra parameters will cause a problem since there is so much data being used to constrain a small number of amplitudes.  It would require a bit of work to change the code to support this, so would prefer to punt on that unless it is needed.

PR #737, opened 2024-02-12
Fix timer imports, run isort and black on scripts

keskitalo:
Many thanks to @chervias for the bug report!


tskisner (approved):
Thanks for catching those imports we missed on the previous PR!

PR #740, opened 2024-03-06
Improve YAML config loading

tskisner:
Small changes to the config loader to better distinguish between format types.


keskitalo (approved):
I like it!

PR #741, opened 2024-03-07
Make AzimuthIntervals more robust

tskisner:
Cut initial and trailing unstable pointing.


tskisner:
> Looks a little ad hoc but I expect it to get the job done. Can we make trimming the short subscans optional?

Done, this is now controlled by a trait, defaulting to True.

PR #742, opened 2024-03-08
Check for zenith in the scan range

keskitalo:
The method that checks for scan range and accounts for field of view gives nonsensical results if the top of the focalplane is path the zenith. This PR raises RuntimeError when that happens.


tskisner (approved):
Thanks for adding this check!

PR #745, opened 2024-03-12
Fix flagging unfiltered data

keskitalo:


tskisner (approved):
Thanks for catching this!

PR #747, opened 2024-03-13
Use the start and stop times to make the session ID more unique

keskitalo:
This will make existing cached atmosphere obsolete but will also reduce clashes in the future.


tskisner (approved):
Looks good, thanks!

PR #748, opened 2024-03-13
Restore HWP Stokes model changes

tskisner:
This restores the HWP Stokes response to the model before #736. The changes in that PR were based on observations of a source with known polarization, but resulted in breakage of common-sense tests with discrete detector / HWP orientations.  Those polarized source results should be revisited.


keskitalo (approved):
Can't wait to have this matter fully settled.

PR #750, opened 2024-03-18
Add gpu staging to coordinate rotated boresight

tskisner:
* Eventually, this could be moved to a dedicated operator which creates rotated boresight pointing.

* For now, this is handled by the pointing_detector operator, and the rotated boresight is created if needed, and moved to the appropriate device.


keskitalo (approved):
Looks good.

PR #751, opened 2024-03-21
Implement an operator for quick-and-dirty white noise modeling

keskitalo:


tskisner (approved):
This is great, and will be useful for bootstrapping a noise model when we know the timestream is dominated by other components like atmosphere loading.  Thanks!

PR #752, opened 2024-03-25
Small update to plotting

tskisner:
Better support maps with only intensity.  Add override to max hits to plot.  Several unrelated changes from running format_source.sh

### DIFF
--- a/src/toast/vis.py
+++ b/src/toast/vis.py
@@ -119,6 +119,7 @@ def plot_wcs_maps(
     range_I=None,
     range_Q=None,
     range_U=None,
+    max_hits=None,
     truth=None,
     xmin=None,
     xmax=None,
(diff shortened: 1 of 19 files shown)


src/toast/vis.py
```diff
@@ -400,20 +406,20 @@ def plot_healpix_maps(
     figsize = (12, 6)
     figdpi = 100
 
-    def plot_single(data, vmin, vmax, out, gnomrot=None):
+    def plot_single(data, vmin, vmax, out, gnomrot=None, reso=4.0, xsize=1000):
```
keskitalo, line 409: You could also consider adding `**kwargs` to the signature here and pass them to the plotting command. Obviously not needed now.

src/toast/vis.py
```diff
@@ -449,9 +459,21 @@ def sym_range(data):
         )
         mlon = np.mean(lon)
         mlat = np.mean(lat)
+        mlon_deg = 180 * mlon / np.pi
```
keskitalo, line 462: I prefer `np.degrees()` over explicit conversion but that is a matter of taste.

keskitalo (approved):
Looks good. It was about time to run `format_source.sh`!

PR #753, opened 2024-04-24
Do not redistribute data unless computing cross spectra.

tskisner:
When estimating the noise PSDs, computing autospectra does not require redistribution.  Also apply a small fix to one of the plotting routines.


keskitalo (approved):
Looks good.

PR #756, opened 2024-05-10
Improve AzimuthIntervals and other small fixes.

tskisner:
- Modify the AzimuthIntervals operator to more robustly exclude false "throw" intervals.  Instead we first detect the stable pointing periods and then find the exact turnaround sample between those (and raise an exception if there is more than one).  This does mean that beginning of the first throw and the end of the last throw are truncated to the stable scan boundary.

- In the mapmaker, keep the noise weighted map if `keep_final_products` is True.

- In the mapmaker, if the binner is configured to compute and save full detector pointing, use that option when initially computing the pixel distribution. This avoids computing the pointing twice.

- When scanning from a map in the template solver, if we already have full detector pointing, then run that over all detectors.

- Fix typo in mapmaker from recent refactor.

### DIFF
--- a/src/toast/ops/azimuth_intervals.py
+++ b/src/toast/ops/azimuth_intervals.py
@@ -149,13 +149,6 @@ class AzimuthIntervals(Operator):
                 # Smooth with moving window
                 wscan_vel = uniform_filter1d(scan_vel, size=window, mode="nearest")
 
-                # When the velocity changes sign, we have a turnaround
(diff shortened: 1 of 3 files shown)


tskisner:
Plot from some test data passing through `AzimuthIntervals`  Note that the initial non-scan timespan is no longer picked up incorrectly as a throw.
az_0

keskitalo (approved):
Looks good.

PR #757, opened 2024-05-11
Refactor of WCS pixelization operator

tskisner:
This work is a superset of #727, and includes a larger reorganization of WCS projection support.  One note:  the WCS standard does not support Alt-Az / Az-El horizon coordinates.  Currently the code uses the RA/DEC CTYPE for that case, but I am open to suggestions if there is some common convention for CTYPE in that case.

Changes in this PR include:

- Move the application of source centering in the projection to a separate helper function in `pointing_utils.py`.  Thanks to @gabrielecoppi for identifying this fix.  Optionally use this new function when computing the scan range for autoscaling.

- In `PixelsWCS`:

  - Add a new general class method that computes the WCS parameters.
  - Add support for SFL projection.
  - Allow projection traits to be changed in any order and only recompute the WCS if needed when exec() is called.
  - Default to a single submap, which is the most efficient choice for the common case of data distributed by detector and many observations co-incident on the sky.

- In the PixelsWCS unit tests:

  - Ensure projection and plotting works for every supported projection type with both fixed parameters and autoscaling.
  - Test mapmaking in both normal mode and with source-centered projections in RA/DEC and Az/El.

- In `plot_wcs_maps`:

  - Set the figure size based on the DPI and the actual size of the image in pixels.
  - Set the unhit pixels to gray.
  - Allow specifying the color map, and default to one of the perceptially uniform ones.

Some examples from unit tests showing projection and plotting defaults (input data was noise):
mapmaking_CAR_map fits_I
source_TAN_map fits_I
drone_SFL_map fits_I

### DIFF
--- a/src/toast/ops/pixels_wcs.py
+++ b/src/toast/ops/pixels_wcs.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2015-2020 by the parties listed in the AUTHORS file.
+# Copyright (c) 2015-2024 by the parties listed in the AUTHORS file.
 # All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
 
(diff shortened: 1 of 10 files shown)


gabrielecoppi (approved):

keskitalo (approved):
Looks good. Very happy to see these fixes.

PR #761, opened 2024-05-28
Add a sidereal targeting mode

keskitalo:
The sidereal target is an extension of a horizontal target. It observes at a fixed elevation and azimuth but only in a specified sidereal time window.


tskisner (approved):
Looks like a useful addition, thanks!

PR #762, opened 2024-06-04
Do not try to invert 0-dimensioned covariance matrices

keskitalo:
Small tweaks to avoid feeding nonsense to compiled kernels.


tskisner (approved):
Looks good, thanks for these sanity checks.

PR #763, opened 2024-06-07
Improve checks of azimuth range for atmosphere simulation

sbiquard:
'Standardize' the az/el range checks against the simulated range in order to not raise unjustified errors.

Also, update the toast_sim_ground_simple workflow.


keskitalo (approved):
Elegant solution!

sbiquard:
Yes, at least I tried the same script that gave me the error before and the atmosphere simulation finished without errors.

keskitalo:
Excellent. I've started the CI tests and will merge once they complete. Thanks a lot for the rapid turnaround!

PR #764, opened 2024-06-10
Small fixes to binned map accumulation and HDF5 dumping

tskisner:
When dumping HDF5, catch an additional exception when dumping python objects.  Switch the binned map accumulation code back to an older version that avoids omp atomic at the cost of duplicate work.


keskitalo (approved):
Looks good. I like the simpler loop structure and the absence of `atomic`.

PR #765, opened 2024-06-11
Add processing bit to non-science shared mask

tskisner:
Many data reduction operators default to using the combined "non-science" masks for shared and detector flags.  The shared mask non-science mask currently does not include the processing bit.  This PR tests the inclusion of that.


keskitalo (approved):
Looks good.

PR #767, opened 2024-06-17
Extend syntax to support multiple flavors

keskitalo:


tskisner (approved):
Looks useful, thanks!

PR #768, opened 2024-06-17
Expand the list of supported units

keskitalo:


tskisner:
The single unit test failure here is unrelated, and we should explore that separately.

tskisner (approved):
Looks good, thanks!

PR #769, opened 2024-06-18
Fix parsing rectangular patches that cross the zero meridian

keskitalo:


tskisner (approved):
This all looks fine, and the unit test failures are unrelated.  As soon as this is merged I will open a new PR for debugging those test failures, which are new and somehow involve astropy table HDF5 metadata.

PR #770, opened 2024-06-18
Ensure that hdf5 attributes are stored as base python types

tskisner:
This attempts to fix recent github test run failures.


tskisner:
This trivial change fixes the recent CI unit test failures.

PR #772, opened 2024-06-25
Extend scan-synchronous signal to have polarization

keskitalo:


tskisner:
Note that the unit test failures here are due to deprecation of `scipy.integrate.trapz` function.  The healpy team already has an [open issue](https://github.com/healpy/healpy/issues/952) on that.  I'll double check that we are not using that directly anywhere in our codebase.  Note that this PR is the first time that numpy-2.0 has been pulled into the conda environment for test running, so there might be other things needed.  We could temporarily pin numpy and scipy to older versions if it takes too long to sort out.

tskisner:
Let's merge this for now, but keep in mind to pin numpy to 1.x to prevent the scipy API change (and hence the healpy breakage).  I'll separately implement that pin in our test environment until healpy is fixed.

tskisner (approved):
Looks good, thanks!

PR #773, opened 2024-06-25
Various small fixes

tskisner:
- Fix deadlock in 1D polyfilter if a process has no good detectors

- When flagging noise model outliers, iterate until no further detectors are cut.

- In the template classes and Amplitude class, handle the case where some processes have no good detectors.


keskitalo (approved):
Looks good. Happy to see the code become more robust against extreme flagging.

PR #776, opened 2024-08-06
Add option in the detector pointing operator to deflect the pointing …

keskitalo:
…based on the HWP angle

### DIFF
--- a/src/toast/ops/pointing_detector/pointing_detector.py
+++ b/src/toast/ops/pointing_detector/pointing_detector.py
@@ -51,6 +51,10 @@ class PointingDetectorSimple(Operator):
         defaults.hwp_angle, allow_none=True, help="Observation shared key for HWP angle"
     )
 
+    hwp_angle_offset = Quantity(
+        0 * u.deg, help="HWP angle offset to apply when constructing deflection"
+    )
+
     hwp_deflection_radius = Quantity(
         None,
         allow_none=True,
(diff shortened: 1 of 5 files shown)


src/toast/ops/pointing_detector/pointing_detector.py
```diff
@@ -46,6 +47,17 @@ class PointingDetectorSimple(Operator):
...
+    hwp_angle = Unicode(
+        defaults.hwp_angle, allow_none=True, help="Observation shared key for HWP angle"
+    )
+
+    hwp_deflection_radius = Quantity(
+        None,
+        allow_none=True,
+        help="If non-zero, nominal detector pointing will be deflected in a circular "
+        "pattern according to HWP phase.",
+    )
```
tskisner, line 59: I think we also need a `hwp_deflection_phase` here (default zero), since the deflection is not necessarily aligned with the fast axis of the HWP.

src/toast/ops/pointing_detector/pointing_detector.py
```diff
@@ -226,6 +238,18 @@ def _exec(self, data, detectors=None, use_accel=None, **kwargs):
                 use_accel=use_accel,
             )
 
+        # Optionally apply HWP deflection
+        if self.hwp_deflection_radius is not None and \
+           self.hwp_deflection_radius.value != 0:
+            xaxis, yaxis, zaxis = np.eye(3)
+            deflection = qa.rotation(
+                xaxis, self.hwp_deflection_radius.to_value(u.radian)
+            )
```
tskisner, line 247: I think the total deflection should be the quaternion product of this rotation and the phase rotation about the `zaxis`.

tskisner (changes requested):
Great to have this feature!  I think if we add a phase to the deflection then it should cover all known cases so far.

src/toast/ops/pointing_detector/pointing_detector.py
keskitalo: I believe this is the case

src/toast/ops/pointing_detector/pointing_detector.py
keskitalo: Please see the revised code. I also added a unit test

tskisner (approved):

tskisner:
Thanks for the update, merge when ready

PR #778, opened 2024-08-15
Add option to reject scans too close to North or South

keskitalo:
A simple extension of the scheduler arguments.


tskisner (approved):
A useful update, thanks!

PR #779, opened 2024-08-17
Change HWP deflection of detector pointing to not alter polarization

tskisner:
* Modify deflection code so that the overall detector polarization direction stays fixed (i.e. the deflection is a translation)

* Add a unit test which makes plots of the detector pointing

* Small fixes to the plotting utility

* Small fix for case where quaternions are constructed from an array of axes.

Here is a plot comparing the existing code's behavior with the changes in this PR.  Note how the old code rotates the detector frame as well as deflecting it.  Old code:

pointing_deflection_old

And the code in this PR:

pointing_deflection_new


keskitalo (approved):
Awesome!

PR #781, opened 2024-08-26
dtype member was removed. Double precision is default.

keskitalo:


tskisner (approved):
Thanks for this fix to allow use of latest suitesparse.  I'll lift that version restriction when testing wheels shortly.

PR #786, opened 2024-09-06
fix trait validation in noise_estimation.py

sbiquard:
The validation for the `nbin_psd` trait was assuming an integer, but it could be `None`.


tskisner (approved):
Thanks for the fix!  Will merge after tests complete.

tskisner (approved):
Thanks for the fix!  Will merge after tests complete.

PR #787, opened 2024-09-19
Apply coadd to only non-zero pixels

keskitalo:
Only co-add nonzero pixels.


tskisner (approved):
This is a very specialized post-processing script, so whatever works in practice is great.  Since every process has a copy of the hit portion of the map, as the number of hit pixels increases, the number of processes per node will be limited by the available memory (e.g imagine a footprint with half the sky at nside=16384).  

If this ever becomes an issue, each of the current "tasks" could be converted to a process group with a `PixelDistribution` where the number of submaps is set to the number of processes in the group (where each process has one submap).  Then all of the usual tools for loading data in parallel, inverting and applying covariance, etc can just be used without ever storing the whole map on a single process.

PR #790, opened 2024-10-01
Add method to sort observing schedule by RA

keskitalo:


tskisner (approved):
Thanks, this is a nice feature for simulated schedules!

PR #791, opened 2024-10-02
New operator for estimating HWP synchronous signal

tskisner:
This work introduces a new operator that can model HWPSS in various
    ways, and optionally estimate the relative gains between detectors
    based on the 2f harmonics.   Features include:
    
- Detection and flagging of samples with a stopped HWP.
    
- Model coefficients can be estimated on the whole observation, on
      fixed-length chunks, or on pre-defined intervals.  The chunk-wise
      model coefficients are then smoothly interpolated over the
      observation.
    
- Model can optionally include a time drift separate from the
      chunking.
    
- The 2f harmonic is used to estimate the relative gain between
      detectors, either as a fixed value per observation or as a
      continuous calibration timestream.  This can be used to flag
      outlier detectors and generate calibration tables / timestreams
      for application.
    
- Includes proper treatment of flagged samples in normalization of
      model coefficients and covariance.

- Extensive optional debug plots.

Here are some comparison plots generated by the unit tests.  The "basic" case has an input HWPSS with fixed coefficients for the observation.  Note that data in the flagged regions can be ignored for the comparison.

compare_basic_small_patch-0-0_D0A-150_26150-26950

The "fixed" case has a random, constant relative gain applied to each detector, which is computed in the new operator and then applied.

compare_fixed_small_patch-0-0_D0A-150_26150-26950

The "continuous" case has a random sine-like calibration drift across the observation.  The new operator is configured to estimate the HWPSS model in overlapping chunks which are interpolated and subtracted.  The relative gain is similarly estimated on chunks and the resulting calibration timestream is applied afterwards.  As expected, the simpler HWPFilter operator (which estimates a single template for the whole observation) does not handle this case well.

compare_continuous_small_patch-0-0_D0A-150_26150-26950


src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+            ch_reltime = reltime[slc]
+
+        obs_cov = hwpss_compute_coeff_covariance(
+            sincos[slc],
+            sh_flags[slc],
+            comm=obs.comm.comm_group,
+            times=ch_reltime,
+            time_drift=self.time_drift,
+        )
+        if obs_cov is None:
```
keskitalo, line 739: If one chunk fails to fit, it could just be a section of flagged data. Is the assumption that such chunks should be screened somewhere earlier?

src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+                obs_cov[0],
+                obs_cov[1],
+                times=ch_reltime,
+                time_drift=self.time_drift,
+            )
+            if idet == 0:
+                cfstr = ""
+                for ic in cf:
+                    cfstr += f"{ic} "
+                # print(f"Det Coeff {det}[{indx}] = {cfstr}")
```
keskitalo, line 774: Maybe time to remove this and other debug statements?

src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+        for idet, det in enumerate(dets):
+            if self.det_flags is None:
+                det_flags[det] = shared_flags
+            else:
+                det_flags[det] = np.copy(obs.detdata[self.det_flags][det])
+                det_flags[det] &= self.det_flag_mask
+                det_flags[det] |= shared_flags
+        return (shared_flags, det_flags)
+
+    def _stopped_flags(self, obs):
```
keskitalo, line 872: If the observation contains a time when the HWP is stopped, we should also flag deceleration/acceleration.

keskitalo (approved):
Looks good and the diagnostic plots are convincing. The only question is if we should scrap the older HWPSS filter altogether.

src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+    chunk_view = Unicode(
+        None,
+        allow_none=True,
+        help="The intervals over which to independently compute the HWPSS template",
+    )
+
+    chunk_time = Quantity(
+        None,
+        allow_none=True,
+        help="The overlapping time chunks over which to compute the HWPSS template",
```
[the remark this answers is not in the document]
tskisner: No, and there is a check for that condition at the start of `_exec()`.

src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+    )
+
+    relcal_fixed = Unicode(
+        None,
+        allow_none=True,
+        help="Build a relative calibration dictionary in this observation key",
+    )
+
+    relcal_continuous = Unicode(
```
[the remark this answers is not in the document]
tskisner: Correct, and there is a check for that at the start of `_exec()`.

src/toast/ops/hwpss_model.py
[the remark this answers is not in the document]
keskitalo: I noticed that looking at the code too. Perhaps the help string should say so?

src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+        for ob in data.obs:
+            timer = Timer()
+            timer.start()
+
+            if ob.comm_row_size != 1:
+                msg = f"{ob.name} is not distributed by detector"
+                raise RuntimeError(msg)
+
+            if self.hwp_angle not in ob.shared:
+                # Nothing to do, but if a relative calibration
```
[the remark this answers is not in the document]
tskisner: The challenge is that sometimes data both with and without a HWP is combined in a single processing.  It seems like printing a warning-level message that the observation is being skipped would be a good compromise.

src/toast/ops/hwpss_model.py
tskisner: That is a good point- currently I exclude "tiny" chunks from the fit, but chunks without a solution could just be skipped in the interpolation.

src/toast/ops/hwpss_model.py
```diff
@@ -0,0 +1,936 @@
...
+            if data.comm.world_rank == 0:
+                os.makedirs(self.debug)
+            if data.comm.comm_world is not None:
+                data.comm.comm_world.barrier()
+
+        for ob in data.obs:
+            timer = Timer()
+            timer.start()
+
+            if ob.comm_row_size != 1:
```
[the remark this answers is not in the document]
tskisner: I have added these properties to the `Observation` class and replaced the manual checks throughout the code with these new properties.

src/toast/ops/hwpss_model.py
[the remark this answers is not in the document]
tskisner: This warning is now implemented.

src/toast/ops/hwpss_model.py
tskisner: This is now handled more gracefully.  Failure to fit individual chunks just results in the removal of that chunk from the overall solution.

src/toast/ops/hwpss_model.py
tskisner: I added code to detect and flag acceleration regions of the hwp angle.  The "basic" unit test now checks that case.

src/toast/ops/hwpss_model.py
tskisner: Done.

PR #792, opened 2024-10-03
Implement Operator loader pattern on execution

tskisner:
This is a small PR that is the beginning of implementing a design pattern that is useful in cases where the solved map-domain or template amplitude products require more detector data than will fit in memory.  Essentially the new `Operator.load_exec()` method loops over observations and looks for an Observation attribute containing a "Loader" class.  If that exists, it is used to "load" data prior to calling `exec()` on that observation and then used to "unload" data afterwards.  See the trivial unit test to see how this works.

Future PRs will use this feature to implement an operator that loads (or simulates) data per observation and accumulates the hits, inverse covariance, and noise weighted map.  This foundational operator can then be used in larger workflows making a binned map or solving for template amplitudes without persistent storage of all detector data.

### DIFF
--- a/src/toast/ops/operator.py
+++ b/src/toast/ops/operator.py
@@ -22,7 +22,7 @@ class Operator(TraitConfig):
         raise NotImplementedError("Fell through to Operator base class")
 
     @function_timer_stackskip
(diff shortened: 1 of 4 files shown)


keskitalo (approved):
An elegant solution.

PR #793, opened 2024-10-11
Jump finder fix

Bai-Chiang:
- Fix jump finder that only finds the maximum jump.
- Replace `scipy.signal.fftconvolve` with `scipy.signal.convolve`. `scipy.signal.convolve` could auto select/guess best convolve method (direct calculation or FFT).
- Simplify for loop in `_get_sigma` function.

### DIFF
--- a/src/toast/ops/simple_jumpcorrect.py
+++ b/src/toast/ops/simple_jumpcorrect.py
@@ -208,10 +208,9 @@ class SimpleJumpCorrect(Operator):
 
         sigmas = []
         nn = len(toi)
-        for start in range(tol, nn - tol, 2 * tol):
+        # Ignore tol samples at the edge
+        for start in range(tol, nn - 3*tol + 1, 2*tol):
             stop = start + 2 * tol
-            if stop > nn - tol:
-                break
             ind = slice(start, stop)
             x = toi[ind][full_flag[ind] == 0]
             if len(x) != 0:
--- a/src/toast/ops/simple_jumpcorrect.py
+++ b/src/toast/ops/simple_jumpcorrect.py
@@ -234,10 +234,13 @@ class SimpleJumpCorrect(Operator):
 
         """
         corrected_signal = signal.copy()
+        nsample = len(signal)
         flag_out = flag.copy()
         for peak, _, amplitude in peaks:
             corrected_signal[peak:] -= amplitude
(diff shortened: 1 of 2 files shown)


src/toast/ops/simple_jumpcorrect.py
```diff
@@ -179,27 +165,30 @@ def _find_peaks(self, toi, flag, flag_out, lim=3.0, tol=1e4, sigma_in=None):
...
             amplitude = mytoi[imax]
             significance = np.abs(amplitude) / sigma
 
-            # Mask the peak for taking mean and finding additional peaks
-            istart = max(0, imax - tol)
-            istop = min(len(mytoi), imax + tol)
             # mask out the vicinity not to have false detections near the peak
+            istart = max(0, imax - tol)
+            istop = min(nsample, imax + tol)
             mytoi[istart:istop] = np.ma.masked
```
tskisner, line 176: Here the samples near the jump are masked for each subsequent iteration.  So this will find all jumps that are spaced at least 2 x `tol` apart.  After correcting those jumps, do we need to go back and look for other jumps that were in these masked regions?  In other words, a loop over `_find_peaks()` and `_remove_jumps()`?

tskisner (approved):
This looks good, and definitely is an improvement!  I just wonder if we need to do an additional loop to catch jumps that are closer together than the tolerance.

src/toast/ops/simple_jumpcorrect.py
Bai-Chiang: `tol` is half of `filterlen` [1] [2]. If there are two jumps within one filter length, it probably won't even find correct jump location. The peak location may not be the jump location. I'd say this is the limitation of this method.

[1] https://github.com/Bai-Chiang/toast/blob/ceb8d47e3c56a3ea4032d358242e29893b941340/src/toast/ops/simple_jumpcorrect.py#L268
[2] https://github.com/Bai-Chiang/toast/blob/ceb8d47e3c56a3ea4032d358242e29893b941340/src/toast/ops/simple_jumpcorrect.py#L126

tskisner: Good point.  Sounds like this is ready to merge then.  Thanks!

keskitalo (approved):
Looks great! Thanks for catching that looping error.

PR #794, opened 2024-10-16
deglitch and jump correct: Fix ignoring given `det_mask` and change default bit mask

Bai-Chiang:
- Fix not using given `det_mask` bit mask
- Change default bit mask


keskitalo (approved):
Nice catch!

tskisner (approved):
Thank you!

PR #795, opened 2024-10-28
Fix HEALPix unit conversion at load time

keskitalo:
Fix an issue with HEALPix map unit conversion.

The units were being parsed correctly from the FITS header but somehow `scale.to(pix.units).value` was always evaluating to `1.0`. Replacing it with `scale.to_value(pix_units)` did the trick.


tskisner (approved):
Thanks for fixing this!

In other places across the code, [this helper function is used](https://github.com/hpc4cmb/toast/blob/bd4313f9a97bdc4e9889268456b3e37f60dcfd71/src/toast/utils.py#L761-L777).  Perhaps we could just fix that helper function and then use the function here?

keskitalo:
Definitely! I just pushed an update and will merge once the unit tests pass.

PR #796, opened 2024-10-30
Create an all-inclusive example unit test for ground-based data.

tskisner:
This work aims to solve two problems:

- Lack of an example unit test for ground telescope data which can be used as a starting point (by removing various things) for future unit tests.

- Inconsistent techniques for "scanning from a map" across many unit tests.  Basically I have removed ad-hoc code for generating fake sky signal and scanning into timestreams and centralized that in several unit test helper functions.

### DIFF
--- a/src/toast/tests/_helpers.py
+++ b/src/toast/tests/_helpers.py
@@ -15,9 +15,11 @@ from astropy import units as u
 from astropy.table import Column, QTable, Table
 from astropy.table import vstack as table_vstack
 from astropy.wcs import WCS
+from scipy.ndimage import gaussian_filter
 
 from .. import ops as ops
 from .. import qarray as qa
+from .. import rng
 from ..data import Data
 from ..instrument import Focalplane, GroundSite, SpaceSite, Telescope
 from ..instrument_sim import fake_boresight_focalplane, fake_hexagon_focalplane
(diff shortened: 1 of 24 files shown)


src/toast/tests/ops_example_ground.py
AERAdler, line 1: I really like this file, although I suspect people will not use it as much as they should. Do you think an ops_example_satellite would be useful at some point? No atmosphere but probably some HWPSS

src/toast/tests/_helpers.py
```diff
@@ -611,7 +956,153 @@ def create_fake_sky_tod(
...
+        for buf in [
+            pixel_pointing.pixels,
+            stokes_weights.weights,
+            pixel_pointing.detector_pointing.quats,
+        ]:
+            if buf in ob:
+                del ob[buf]
+
+
+# def create_fake_sky(data, dist_key, map_key, hpix_out=None):
```
AERAdler, line 971: What is the purpose of this commented out section? It looks like a copy of the previous state of the code.

AERAdler (commented):
I made comments on ops_example_ground and the commented out section of helpers. I tried running some of the tests and it looks to me like everything is in order.

src/toast/tests/_helpers.py
tskisner: Thanks for catching, I removed this.  It was the old version of the helper functions that were replaced in this PR.

src/toast/tests/ops_example_ground.py
tskisner: I think an equivalent `ops_example_satellite.py` would definitely be useful for a future PR.  But beyond the scope of this one.

PR #797, opened 2024-10-30
Ensure that reset of pixel distribution clears stale products

tskisner:
If the pixel distribution used for mapmaking is being reset, then all map domain products are invalidated.


keskitalo (approved):
This is great!

PR #798, opened 2024-11-01
ops.SimpleJumpCorrect: Flag detector invalid if too many jumps.

Bai-Chiang:
Add input option `njump_limit`. If there are more than `njump_limit` jumps in one detector data, then flag the detector and time stream as invalid.

Do we need to flag the detector `ob._detflags[name] = self.det_mask` or just flag the entire time stream view?

### DIFF
--- a/src/toast/ops/simple_jumpcorrect.py
+++ b/src/toast/ops/simple_jumpcorrect.py
@@ -94,6 +94,12 @@ class SimpleJumpCorrect(Operator):
         help="Minimum number of good samples in an interval",
     )
 
+    njump_limit = Int(
+        10,
+        help="If the detector has more than `njump_limit` jumps the detector "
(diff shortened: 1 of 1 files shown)


keskitalo (approved):
Looks good. A definite improvement.

Bai-Chiang:
Do we need to flag the detector `ob._detflags[name] = self.det_mask` or just flag the entire time stream view? What if `view` is not `None`, and calculating slice of data, do we still flag entire detector?

keskitalo:
Actually, instead of `ob._detflags[name] = self.det_mask` it would be prudent to use `ob._detflags[name] |= self.det_mask` to avoid nulling any other detector flag bits.

As for flagging the detector even if the view is not `None` it seems like the safe approach. If you are finding a ton of jumps, something must be going on with the detector.

PR #800, opened 2024-11-14
Improvements to AzimuthIntervals

tskisner:
- Use a more robust technique for computing the scan velocity and acceleration in the presence of noise and flags
- Use a fit to the velocity for finding turnarounds
- Add a unit test to generate plots and check that the expected number of scans is found

Example plot of scan properties found in the unit tests:
az_intervals

### DIFF
--- a/src/toast/ops/simple_jumpcorrect.py
+++ b/src/toast/ops/simple_jumpcorrect.py
@@ -100,6 +100,18 @@ class SimpleJumpCorrect(Operator):
         "the detector and time stream will be flagged as invalid.",
     )
 
+    save_jumps = Unicode(
+        None,
+        allow_none=True,
+        help="Save the jump corrections to a dictionary of values per observation",
+    )
+
+    apply_jumps = Unicode(
+        None,
+        allow_none=True,
+        help="Do not compute jumps, instead apply the specified dictionary of values",
+    )
+
     @traitlets.validate("det_mask")
     def _check_det_mask(self, proposal):
         check = proposal["value"]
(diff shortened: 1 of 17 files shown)


src/toast/utils.py
```diff
@@ -831,3 +831,86 @@ def is_empty(self):
             if (len(v) > 0) and not all(x is None for x in v):
                 return False
         return True
+
+
+def flagged_noise_fill(data, flags, buffer, poly_order=1):
```
keskitalo, line 836: I think we ultimately want a unit test for a frequently used utility function like this but I would not hold up merging the PR.

keskitalo (approved):
Looks good. More robust interval finding, backed by a unit test.

src/toast/ops/simple_jumpcorrect.py
```diff
@@ -258,7 +274,11 @@ def _exec(self, data, detectors=None, **kwargs):
 
             local_dets = ob.select_local_detectors(flagmask=self.det_mask)
             shared_flags = ob.shared[self.shared_flags].data & self.shared_flag_mask
+            if self.save_jumps is not None:
```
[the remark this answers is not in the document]
tskisner: Good point, that was sloppy.  I'll fix shortly.

src/toast/utils.py
tskisner: Definitely agree.  Also the correct way to do this would be to iteratively fill all gaps simultaneously with a constrained noise realization.  That kind of technique was expensive the last time I implemented code for such things (circa 2005, written in C).  I should dig that up and see if it is fast enough now for this purpose.

tskisner: This function has been improved now, and a unit test is in place, along with a standalone operator that just runs the gap filling.

PR #801, opened 2024-11-19
bugfix error message in sim_ground_utils elevation modulation

AERAdler:


keskitalo (approved):
Awesome! Thank you for the fix.

PR #806, opened 2025-02-05
Add ability to create a PixelDistribution from a sky footprint.

tskisner:
In most workflows, a PixelDistribution is created by passing through the detector pointing on each process and determining the locally hit pixels and submaps for a given pixelization scheme.  This work adds support for creating a distribution from a pre-existing sky footprint.  Each process will have the same hit submaps in this case. This is a stepping stone towards building the RHS out of core while keeping the locally hit pixels in node shared memory.

### DIFF
--- a/src/toast/footprint.py
+++ b/src/toast/footprint.py
@@ -1,0 +1,162 @@
+# Copyright (c) 2024-2025 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+import numpy as np
+import healpy as hp
+import astropy.io.fits as af
+
+from .pixels import PixelData, PixelDistribution
+from .pixels_io_healpix import read_healpix
+from .pixels_io_wcs import read_wcs_fits
+
+
+def footprint_distribution(
+    healpix_nside=None,
+    healpix_nside_submap=None,
+    healpix_submap_file=None,
+    healpix_coverage_file=None,
+    wcs_coverage_file=None,
+    comm=None,
+):
+    """Create a PixelDistribution from a pre-defined sky footprint.
+
+    Usually a PixelDistribution object is created by passing through the detector
+    pointing and determining the locally hit submaps.  However, this can be expensive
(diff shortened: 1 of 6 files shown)


src/toast/footprint.py
```diff
@@ -0,0 +1,162 @@
...
+            or healpix_coverage_file is not None
+            or healpix_submap_file is not None
+        ):
+            msg = "If loading a wcs coverage file, all other options should be None"
+            raise RuntimeError(msg)
+        n_pix = None
+        if rank == 0:
+            hdulist = af.open(wcs_coverage_file)
+            n_pix = 1
+            for s in hdulist[0].data.shape:
```
keskitalo, line 81: `n_pix = np.prod(hdulist[0].data.shape)`?

src/toast/footprint.py
```diff
@@ -0,0 +1,162 @@
...
+    elif healpix_coverage_file is not None:
+        if healpix_nside_submap is None:
+            msg = "You must specify the submap NSIDE to use with the coverage file"
+            raise RuntimeError(msg)
+        n_pix = None
+        n_submap = None
+        local_submaps = None
+        if rank == 0:
+            hpix_data = read_healpix(healpix_coverage_file, field=(0,), nest=True)
+            nside = hp.npix2nside(len(hpix_data))
```
keskitalo, line 98: `nside = hp.get_nside(hpix_data)`?

keskitalo (approved):
Looks good. Do I understand correctly, that we have no submap support for rectangular pixelization?

tskisner:
Re: submaps:

- The WCS projections in general support submaps, but this domain decomposition is just over the pixel index (so blocks of rows / columns in the projection).  In practice this is not as useful as nested healpix submaps, since a typical scan strategy from the ground means that all processes essentially cross all submaps.
- One could consider supporting a tiled scheme like pixell, and treat each tile as a submap.  I don't see much technical benefit over healpix, but it is something we could support for compatibility reasons.
- For the purposes of this PR and the footprint distribution helper function, a WCS projection is just assigned one submap (for the reasons above).

I'll work on the other fixes shortly.

PR #808, opened 2025-02-20
Use the detector polarization type to infer psi_pol in Pxx

keskitalo:
Two quick fixes to support ground-based conviqt simulations.

### DIFF
--- a/src/toast/ops/conviqt.py
+++ b/src/toast/ops/conviqt.py
@@ -344,6 +344,8 @@ class SimConviqt(Operator):
         props = focalplane[det]
         if "pol" in props.colnames:
             pol = props["pol"]
+        else:
+            pol = None
         if pol == "A" or det.endswith(("a", "A", "t", "T")):
             return 0.0
         elif pol == "B" or det.endswith(("b", "B")):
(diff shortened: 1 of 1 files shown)


src/toast/ops/conviqt.py
```diff
@@ -342,6 +342,16 @@ def _get_psi_pol(self, focalplane, det):
         if det not in focalplane:
             raise RuntimeError(f"focalplane does not include {det}")
         props = focalplane[det]
+        if "pol" in props.colnames:
+            pol = props["pol"]
+        else:
+            pol = None
+        if pol == "A" or det.endswith(("a", "A", "t", "T")):
```
tskisner, line 349: So this means that if the detector table happens to have a column named `pol` (with any value) and the detector name ends in `A`, it will trigger this branch.  Is that fragile?

For simulated data from SO, this probably is fine since the [det names end in "A" or "B"](https://github.com/simonsobs/sotodlib/blob/72660b5514fca740f98d254d2e7c61e8b124a559/sotodlib/toast/sim_focalplane.py#L330) and there is a "pol" column with a matching value of "A" or "B".

For real SO data, the detectors have names like "Mv5_f150_Br06c04A", but we index the detectors by readout name (which doesn't end in "A", "B", "T", etc).  So this is probably safe for that case.

In the future, if we find that we are running this operator frequently, we could create a trait that specifies the detector table column to use for psi_pol.

tskisner (approved):
Fine for now, although it could be fragile...

PR #809, opened 2025-02-24
Add support for HDF5 I/O of observations without detector data

tskisner:
- The default list of detdata fields for save / load is now the standard signal and flags.

- An empty list of detdata fields is now interpreted as "no fields" rather than "all fields".

- Add a unit test to check save / load with empty detector data.


keskitalo (approved):
Looks good. Thanks for adding a unit test for this use case.

PR #810, opened 2025-03-05
New SSOPatch offset_from_sso option

AERAdler:
Like Katie asked. not sure about the xi/eta-> az/el correspondence.

### DIFF
--- a/src/toast/schedule_sim_ground.py
+++ b/src/toast/schedule_sim_ground.py
@@ -400,7 +400,8 @@ class SSOPatch(Patch):
         el_min=0,
         el_max=np.pi / 2,
         elevations=None,
-        offset_from_sso=None, #(xi, eta)
+        offset_eta=0, 
+        offset_xi=0
     ):
         self.name = name
         self.weight = weight
@@ -410,15 +411,14 @@ class SSOPatch(Patch):
         self.el_min = el_min
         self.el_max = el_max
         self.el_max0 = el_max
+        self.offset_eta = offset_eta
+        self.offset_xi = offset_xi
         self.parse_elevations(elevations)
-        self.offset_from_sso = offset_from_sso
         try:
             self.body = getattr(ephem, name)()
         except:
             raise RuntimeError("Failed to initialize {} from pyEphem".format(name))
         self.corners = None
-        if (self.offset_from_sso is not None) and len(self.offset_from_sso)!=2:
-            raise RuntimeError("Specified a sso offset but it is the wrong size")
         return
 
     def update(self, observer):
@@ -427,12 +427,10 @@ class SSOPatch(Patch):
         """
         self.body.compute(observer)
         ra, dec = self.body.ra, self.body.dec
-        if self.offset_from_sso:
-            xi_off = self.offset_from_sso[0]
-            eta_off = self.offset_from_sso[1]
-            az_newcen = xi_off + self.body.az
-            el_newcen = eta_off + self.body.alt
-            ra, dec = observer.radec_of(az_newcen, el_newcen)
+        if self.offset_eta !=0 or self.offset_xi !=0:
+            delta_az = -self.offset_xi*np.cos(self.body.alt)
+            delta_el = self.offset_eta
+            ra, dec = observer.radec_of(self.body.az+delta_az, self.body.alt+delta_el)
 
         # Synthesize 8 corners around the center
         phi = ra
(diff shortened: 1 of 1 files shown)


src/toast/schedule_sim_ground.py
```diff
@@ -426,11 +427,14 @@ def __init__(
         self.el_max = el_max
         self.el_max0 = el_max
         self.parse_elevations(elevations)
+        self.offset_from_sso = offset_from_sso
         try:
             self.body = getattr(ephem, name)()
         except:
             raise RuntimeError("Failed to initialize {} from pyEphem".format(name))
         self.corners = None
+        if (self.offset_from_sso is not None) and len(self.offset_from_sso)!=2:
```
keskitalo, line 436: Storing the offset as a tuple means that you have to parse it every time you need to apply it.

Instead, can you try adding a `try-except` statement where you parse `offset_from_sso` into `self.offset_az` and `self.offset_el` and raise an appropriate exception when that fails?

src/toast/schedule_sim_ground.py
```diff
@@ -439,6 +443,13 @@ def update(self, observer):
         """
         self.body.compute(observer)
         ra, dec = self.body.ra, self.body.dec
+        if self.offset_from_sso:
+            xi_off = self.offset_from_sso[0]
+            eta_off = self.offset_from_sso[1]
+            az_newcen = xi_off + self.body.az
```
keskitalo, line 449: Thinking about this a little deeper now.. Does this really work? On-sky, the provided offset should always translate to the same angular distance. However, if you apply the offset in Az/El, the distance can be tiny when the telescope is pointing high in elevation. You probably need to apply the offset using pointing vectors and quaternions.

src/toast/schedule_sim_ground.py
```diff
@@ -2893,6 +2904,11 @@ def parse_args(opts=None):
         required=False,
         help="Fixed observing elevations in a comma-separated list.",
     )
+    parser.add_argument(
```
keskitalo, line 2907: Adding the offset as a command line argument is not ideal. We want to add different offsets to different targets so that the scheduler can treat calibration scans of each wafer separately. The offset should be recorded as additional fields to the SSO patch.

keskitalo (changes requested):
I think this PR still incomplete. You can tag people in the PR discussion if you want input before requesting a formal review.

Still, good that you opened the draft as I had some observations that will need to be addressed before we can merge.

AERAdler:
@tskisner Tagging you for advice on coordiate transformations

src/toast/schedule_sim_ground.py
AERAdler: yes that seems better.

src/toast/schedule_sim_ground.py
AERAdler: Looking at the coordinate pdf it has the following line:
"One often encounters the approximation that amounts to ξ = −∆az/ cos(el), η = ∆el."
Maybe if I implement that it is good enough?

keskitalo: It is a little bit crude for large focalplanes like the SAT but we can certainly test it first.

src/toast/schedule_sim_ground.py
```diff
@@ -2654,7 +2652,7 @@ def parse_args(opts=None):
     --patch name,weight,lon,lat,radius (center and radius)
     --patch name,weight,lon_min,lat_max,lon_max,lat_min (rectangle)
     --patch name,weight,lon1,lat1,lon2,lat2,...,lonN,latN (polygon, N>=3)
-    --patch name,SSO,weight,radius (Solar System Object)
+    --patch name,SSO,weight,radius,eta_off,xi_off (Solar System Object, last two optional offset for pointingnat specific place on fp)
```
keskitalo, line 2655: Can you add square brackets around the optional fields: `,eta_off,xi_off` => `[,eta_off,xi_off]` so it is immediately obvious they are optional?

There is also a missing space at `pointingat`.

src/toast/schedule_sim_ground.py
```diff
@@ -443,12 +443,10 @@ def update(self, observer):
...
         self.body.compute(observer)
         ra, dec = self.body.ra, self.body.dec
-        if self.offset_from_sso:
-            xi_off = self.offset_from_sso[0]
-            eta_off = self.offset_from_sso[1]
-            az_newcen = xi_off + self.body.az
-            el_newcen = eta_off + self.body.alt
-            ra, dec = observer.radec_of(az_newcen, el_newcen)
+        if self.offset_eta !=0 or self.offset_xi !=0:
+            delta_az = -self.offset_xi*np.cos(self.body.alt)
```
keskitalo, line 447: We can use this approximation for now, but you should add a comment about it.

keskitalo (commented):
Very minor requests for the code.  I would like to see a simulation that verifies the offsets are working as intended. See https://github.com/simonsobs/pwg-scripts/blob/master/pwg-tds/sim-sso/run.lat.sh

src/toast/schedule_sim_ground.py
AERAdler: Looking at the sim-sso/run.lat.sh I am a bit confused by how the boresight_offset arguments gel with our new offset of patches

keskitalo: This the the opposite of what you are doing since your offset applies to the target.  If, for example, the scheduler was told to only schedule el=50 scans, applying boresight offset in the old implementation would lead to the boresight actually being at a different elevation.

The way you implemented the offset, elevation constaints will be observed and we can specify separate offsets for different targets.

keskitalo (approved):
Elegant solution! Please run `black` on the file to polish some white space issues.

PR #812, opened 2025-04-07
Make unit test helpers file into a public submodule

tskisner:
- Delete old tutorial directory    
- Move mkdocs to final docs location    
- Add github workflow for deploying docs    
- Convert _helpers.py into a submodule.  Small tweaks to create_outdir helper.
- Change imports across all unit tests
- Run format_source.sh


tskisner:
This PR has many trivial changes, mostly moving and deleting files.

src/toast/mpi.py
```diff
@@ -114,7 +114,7 @@ class Comm(object):
     """Class which represents a two-level hierarchy of MPI communicators.
 
     A Comm object splits the full set of processes into groups of size
-    "group".  If group_size does not divide evenly into the size of the given
+    `groupsize`.  If groupsize does not divide evenly into the size of the given
     communicator, then those processes remain idle.
```
keskitalo: I know you didn't change this line but it should be `some processes`, not `those processes`.

src/toast/tests/timing.py
```diff
@@ -68,19 +68,13 @@ def test_global(self):
...
-            except:
-                print("Successful exception:  stop an already stopped timer")
         for nm in tnames:
             gt.start(nm)
         for nm in tnames:
             self.assertTrue(gt.is_running(nm))
             try:
                 s = gt.seconds(nm)
-            except:
+            except Exception:
```
keskitalo: This would look more professional if we caught the subclass of Exception that we anticipate

keskitalo (approved):
We have a lot of unit tests!

The fact that we are importing methods from the `interactive` module in every unit test has me wondering if the module name is descriptive or not. Obviously not an issue for this PR.

src/toast/mpi.py
tskisner: Fixed.

src/toast/tests/timing.py
tskisner: Changed to RuntimeError, which is what is thrown from the C++ code.

tskisner:
Summarizing offline discussion:

- The `toast.interactive` module should contain just things intended to run from a notebook (i.e. that require ipython)
- The rest of the "helper" functions that setup synthetic test cases are primarily useful for the unit tests, even if we also want to use them for documentation tutorials.  For this code, it is fine to just make a public `toast.tests.helpers` module that replaces the old, large `_helpers.py`.  In principle this could also grow to include more "cook book" style involved examples- since we want to test that code as well.
- I will do this rearranging and rebase / squash to reduce the noise in this PR.

PR #813, opened 2025-04-10
Add a TOAST operator that can scan different maps for different detec…

keskitalo:
…tors, wafers or any other focalplane key

### DIFF
--- a/src/toast/ops/scan_healpix_detector.py
+++ b/src/toast/ops/scan_healpix_detector.py
@@ -1,0 +1,245 @@
+# Copyright (c) 2025 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+import os
+
+import healpy as hp
+import numpy as np
+import toast.qarray as qa
+import traitlets
+from astropy import units as u
+from pshmem import MPIShared
+
+from ..observation import default_values as defaults
+from ..pixels_io_healpix import read_healpix
+from ..timing import function_timer
+from ..traits import Bool, Instance, Int, Unicode, Unit, trait_docs
+from ..utils import Logger
+from .operator import Operator
+
+
+@trait_docs
+class ScanHealpixDetectorMap(Operator):
+    """Operator which reads a HEALPix format map from disk and scans it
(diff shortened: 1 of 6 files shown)


tskisner:
I just merged PR #812  with the switch to the `tests.helpers` module.  It will impact this branch by needing a tweak to the unit test and adding the `subdir` option like this:

https://github.com/hpc4cmb/toast/blob/dedffb2a0cd39f92dee17aff71ed9535ee14d3a9/src/toast/tests/ops_scan_healpix.py#L28

src/toast/ops/scan_healpix_detector.py
```diff
@@ -0,0 +1,250 @@
...
+                self.pixel_pointing.apply(ob_data, detectors=[det])
+                self.stokes_weights.apply(ob_data, detectors=[det])
+                pix = current_ob.detdata[self.pixel_pointing.pixels][det]
+                nest = self.pixel_pointing.nest
+                nside = self.pixel_pointing.nside
+                # Get pointing weights
+                weights = current_ob.detdata[self.stokes_weights.weights][det].T.copy()
+
+                # Load and sample the provided maps
+                for imap, file_name in enumerate(self.file_names):
```
tskisner, line 196: You are looping over detectors and then looping over maps.  So (for example) if you had 100 local detectors that were randomly ordered on the focalplane key you were using, you might be loading the same maps dozens of times on each process.  And this would be repeated for every observation.  That would be a lot of data reads.

In the existing `ScanHealpixMap` operator, we loop over maps, load them in a distributed way, and then scan all detectors over them.  However, it would be an easy change to [scan over a subset of the detectors for each map](https://github.com/hpc4cmb/toast/blob/dedffb2a0cd39f92dee17aff71ed9535ee14d3a9/src/toast/ops/scan_healpix.py#L243).  That existing operator also pipelines the steps to automatically handle pointing expansion, etc.

Would it be simpler and more performant to just add the detector / sky map dictionary to the existing operator and then do the scanning with the desired subset of detectors?

tskisner (approved):
The code is fine, but it feels like it might duplicate functionality from the existing operator.  I guess we could revisit this in the future if the map loading cost becomes a dominant cost.

PR #817, opened 2025-04-16
Mss3

keskitalo:
A bucket of tweaks needed to facilitate the MSS3 simulations:
- improve compatibility between the `FilterBin` operator and `sotodlib` `Splits` operator
- fix a bug in producing the string representation of a `Scan` object when the Celestial coordinates are not yet calculated
- Add support for flagging the inverse of a given view
- Add a unit test for `FlagIntervals`
- Fix more interval time stamp comparison issues. The default relative tolerance is completely inadequate.

### DIFF
--- a/src/toast/observation_view.py
+++ b/src/toast/observation_view.py
@@ -91,13 +91,19 @@ class SharedView(MutableMapping):
 class View(Sequence):
     """Class representing a list of views into any of the local observation data."""
 
-    def __init__(self, obj, key):
+    def __init__(self, obj, key, inverse=False):
         self.obj = obj
         self.key = key
         # Compute a list of slices for these intervals
-        self.slices = [slice(x.first, x.last, 1) for x in self.obj.intervals[key]]
+        if inverse:
+            self.slices = [slice(x.first, x.last, 1) for x in ~self.obj.intervals[key]]
+        else:
+            self.slices = [slice(x.first, x.last, 1) for x in self.obj.intervals[key]]
         self.detdata = DetDataView(obj, self.slices)
         self.shared = SharedView(obj, self.slices)
+
+    def __invert__(self):
+        return View(self.obj, self.key, inverse=True)
 
     def __getitem__(self, key):
         return self.slices[key]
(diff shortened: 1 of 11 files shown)


src/toast/observation_view.py
```diff
@@ -137,22 +143,32 @@ def __init__(self, obj):
...
                 )
             self.obj._views[view_name] = View(self.obj, key)
             # Register deleter callback
             if key is not None:
                 self.obj.intervals.register_delete_callback(key, self.__delitem__)
-        return self.obj._views[view_name]
+        if inverse:
+            return ~self.obj._views[view_name]
+        else:
+            return self.obj._views[view_name]
```
tskisner, line 186: Inverting intervals on the fly is a neat feature.  However, this is not free.  My concern is that if calling code repeatedly loops over negated intervals, then that will create a temporary object on every call.  It would be much better to just create such an interval list once.  For the use case in this PR (flagging negated intervals) it is fine, but we should avoid ever using the `~` syntax when passing views to pointing operators and mapmaking templates for example.  Another option would be to only create these negated intervals temporarily in the FlagIntervals operator and then delete them.  I am fine leaving this as-is, but we should perhaps add a warning to the documentation.

src/toast/ops/flag_intervals.py
```diff
@@ -20,7 +20,8 @@ class FlagIntervals(Operator):
     with shared flags.  The view_mask trait is a list of tuples.  Each tuple contains
     the name of the view (i.e. interval) to apply and the bitmask to use for that
     view.  For each interval view, flag values in the shared_flags object are bitwise-
-    OR'd with the specified mask for samples in the view.
+    OR'd with the specified mask for samples in the view.  If the name of the view is
+    prefixed with '~' or '-', the bitmask is applied to all samples outside the view.
```
tskisner: You mention using `-` here in addition to `~`, but the observation view manager only considers `~`.  Did I miss something?

tskisner (approved):
Some minor comments.  Thanks for the additional unit tests!

src/toast/observation_view.py
keskitalo: Good point. Added a warning that is triggered when the same view is inverted more than 100 times.

src/toast/ops/flag_intervals.py
keskitalo: That was the original idea but really only `~` makes sense.

PR #818, opened 2025-04-18
Fix intervals

keskitalo:
Current TOAST `IntervalList` class has an issue when two disjoint sets of intervals like left-going and right-going scans are combined:  if the union of the lists is continuous, the intervals get merged into one long interval.  This happens even without calling the `simplify()` method.

The problem comes from choices made on how to handle intervals that end exactly at a timestamp. Currently, that sample gets excluded, if it is also claimed by the next interval. This caused the exact span of one interval to depend on the definition of the subsequent interval.

In this PR, intervals are redefined to be open-ended.  Samples are included up-to but not including the end time stamp. `np.isclose(rtol=1e-12)` is used to compare floating point time stamps. To make the interval definition more intuitive and compatible with Numpy indexing, the sample ranges defining the interval are now also open ended. This redefinition required small changes to many source files.

Current initialization of the `IntervalList` causes the provided time spans to be rewritten in terms of the actual time stamps. This is also problematic since using the resulting `IntervalList` to initialize a new one with potentially different sampling would cause the interval time spans to repeatedly shift from their original positions.  The new code faithfully maintains the provided time spans.

### DIFF
--- a/src/toast/intervals.py
+++ b/src/toast/intervals.py
@@ -141,7 +141,7 @@ class IntervalList(Sequence, AcceleratorObject):
 
     def _find_indices(self, timespans):
         # Each interval covers all samples where the sample time meets:
-        # interval.start <= self.timestamps AND self.timestamps < self.timestamps
+        # interval.start <= self.timestamps AND self.timestamps < interval.stop
         # (open-ended interval)
         # with one exception: if the interval ends at the last timestamp, the
         # corresponding sample is included (closed interval)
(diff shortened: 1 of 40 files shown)


src/toast/intervals.py
```diff
@@ -73,73 +74,97 @@ def __init__(self, timestamps, intervals=None, timespans=None, samplespans=None)
...
+            # Handle the end of the timestamps differently
+            return self.timestamps[sample - 1]
+        else:
+            return self.timestamps[sample]
+
     def _find_indices(self, timespans):
-        start_indx = np.searchsorted(
-            self.timestamps, [x[0] for x in timespans], side="left"
+        # Each interval covers all samples where the sample time meets:
+        # interval.start <= self.timestamps AND self.timestamps < self.timestamps
```
tskisner: I think this comment has a typo...

tskisner (approved):
Thanks for this design change, and for dealing with the tedious consequences across the code base.  Changing the indexing of intervals to use standard numpy form should help avoid future errors.  This change will break downstream code in multiple other repos, but that should not be a problem as long as we fix those other codebases before making a tag with this change.

PR #821, opened 2025-05-01
Fix unit test

keskitalo:
Fix an error in a unit test which resulted from the recent change in interval indexing.

Also add a tiny fix to facilitate gcc15 checking `extern C` more rigorously.


tskisner (approved):
Thanks for this fix!

PR #823, opened 2025-05-06
Add unit test for IntervalList creation from close ended intervals

tskisner:
This unit test fails, but I think it should work, [based on the comments in the recently updated code](https://github.com/hpc4cmb/toast/blob/46230c8949e5be3e554da819ad3ee450bc77f6e1/src/toast/intervals.py#L143-L147).  @keskitalo can you take a look and see what you think?

### DIFF
--- a/src/toast/tests/intervals.py
+++ b/src/toast/tests/intervals.py
@@ -183,7 +183,9 @@ class IntervalTest(MPITestCase):
         # Time ranges are open ended *unless* they fall exactly on a sample.
         # Here we set the ending time of each span to exactly the timestamp
         # of the final sample.
-        timespans = [(stamps[x[0]], stamps[x[1] - 1]) for x in samplespans]
+        timespans = [
+            (stamps[x[0]], stamps[min(x[1], stamps.size - 1)]) for x in samplespans
+        ]
         intr_times = IntervalList(stamps, timespans=timespans)
 
         self.assertTrue(intr_samples == intr_times)
--- a/src/toast/tests/intervals.py
+++ b/src/toast/tests/intervals.py
@@ -180,9 +180,8 @@ class IntervalTest(MPITestCase):
         samplespans = [(x * n_intr, x * n_intr + n_intr) for x in range(n_intr)]
         intr_samples = IntervalList(stamps, samplespans=samplespans)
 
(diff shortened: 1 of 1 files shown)


keskitalo:
I think there is a misunderstanding here. Both time and sample definitions are open-ended, except for the very last sample time.

Say that there is an interval boundary at time `T` and sample `S`. Regardless of which definition you use, sample `S` will belong to the interval that starts at `T`, not to the one that ends at `T`.

In your unit test, the time-based intervals should be constructed from
```
timespans = [(stamps[x[0]], stamps[min(x[1], stamps.size - 1)]) for x in samplespans]
```
not as
```
timespans = [(stamps[x[0]], stamps[x[1] - 1]) for x in samplespans]
```

I realize that taking the minimum looks clunky, but if the user already has the sample indices, then there hardly is a reason to construct time spans by sampling the `stamps` vector.

tskisner:
Ok, thanks for clarifying the intended design.  Creating [observation intervals from timestamps](https://github.com/hpc4cmb/toast/blob/46230c8949e5be3e554da819ad3ee450bc77f6e1/src/toast/observation_data.py#L2077) is a common pattern.  It is used in the `spt3g` interfaces, hdf5 data loading, building azimuth intervals, etc.  The spt3g code is currently broken.  I just fixed this unit test to capture the intended behavior and will fix the spt3g code and any other problems in a separate PR.

src/toast/tests/intervals.py
```diff
@@ -171,6 +171,25 @@ def test_inverse(self):
...
+    def test_closed_time_spans(self):
+        n_samp = 100
+        n_intr = 10
+        stamps = 1000.0 * np.arange(n_samp, dtype=np.float64)
+
+        # Sample ranges are open ended
+        samplespans = [(x * n_intr, x * n_intr + n_intr) for x in range(n_intr)]
+        intr_samples = IntervalList(stamps, samplespans=samplespans)
+
+        # Time ranges are open ended *unless* they fall exactly on a sample.
```
keskitalo: time ranges are open-ended, unless the end time coincides with the observation end time.

keskitalo (approved):
I would change the comment but the test is fine.

PR #824, opened 2025-05-13
Pr4 dipole

keskitalo:
Make PR4 dipole the default and add a method to plot a map of the dipole.


tskisner (approved):
Thanks for these updates and for consolidating the numbers in one place.

PR #826, opened 2025-05-20
Stare scans

keskitalo:
A schedule made a collaborator exposed some fragility in `SimGround()`. This PR fortifies it against zero length scans and scans with no azimuthal range.

### DIFF
--- a/src/toast/ops/sim_ground_utils.py
+++ b/src/toast/ops/sim_ground_utils.py
@@ -403,6 +403,29 @@ def step_el(
 
 
 @function_timer
+def simulate_stare(t_start, t_stop, rate, el, az):
+    """ Return times, azimuth, elevation and intervals for a stare """
+    samples = int((t_stop - t_start) * rate)
+    times = t_start + np.arange(samples) / rate
+    az_sample = np.zeros(samples) + az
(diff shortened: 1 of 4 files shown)


src/toast/ops/sim_ground_utils.py
```diff
@@ -402,6 +402,29 @@ def step_el(
...
 
 
+@function_timer
+def simulate_stare(t_start, t_stop, rate, el, az):
+    """ Return times, azimuth, elevation and intervals for a stare """
+    samples = int((t_stop - t_start) * rate)
+    times = t_start + np.arange(samples) / rate
+    az_sample = np.zeros(samples) + az
+    el_sample = np.zeros(samples) + el
+    # FIXME: do we need a flag bit for stationary telescope?
```
tskisner, line 412: When such data is passed to the `AzimuthIntervals` operator, it would be marked as a very long turnaround interval.

tskisner (approved):
Looks good.

src/toast/ops/sim_ground_utils.py
keskitalo: Probably fine for now, until we actually need to deal with scans like this.

PR #827, opened 2025-05-22
Add error message

Bai-Chiang:
Print all available focalplane keys in error message when the given one is invalid.


tskisner (approved):
Seems like a helpful addition

PR #828, opened 2025-05-27
Add support for reading and writing WCS maps in HDF5 format

keskitalo:
This PR makes some substantial changes to the handling of WCS maps and pixelization:
- internal representation of the maps is row-major, which is the Numpy default
- the `wcs_shape` of an image is `(n_row, n_col)` which is more intuitive
- add `read_wcs` and `write_wcs` convenience methods that
    - support both FITS and HDF5 file format
    - input and output the arrays following the same conventions as `enmap`
- the WCS geometry can now be read out of a FITS header
- fix a corner case bug affecting automatic WCS bounds discovery: if the `data` object included observations across the zero meridian, the discovered longitude range would be greater than 360 degrees
- add a number of unit tests for existing and new WCS methods
- add a simulation operator that can scan different map for every optics tube / wafer / pixel / detector
- The changes fix an old issue causing the TOAST-made WCS maps being transposes of maps made with pixell-based tools
- make the tests in `tests/ops_pointing_wcs.py` run a lot quicker by reducing the resolution and sample rate
- fix an issue in `tests/pixels.py` that caused the tested distributed pixel object to be empty


tskisner:
This is a large PR and I am still working through it.  One initial thing i noticed was that you swapped the row / column major order of the pixel arrays.  I tried running the unit tests and noticed that in the `pixels_wcs` test output some files have headers where the NAXIS header data does not match the other metadata.  For example:
```
$> listhead toast_output/pixels_wcs/data_sub1_type-l.fits

Header listing for HDU #1:
SIMPLE  =                    T / conforms to FITS standard
BITPIX  =                   64 / array data type
NAXIS   =                    3 / number of array dimensions
NAXIS1  =                  100
NAXIS2  =                   40
NAXIS3  =                    6
WCSAXES =                    2 / Number of coordinate axes
CRPIX1  =                 19.5 / Pixel coordinate of reference point
CRPIX2  =                 49.5 / Pixel coordinate of reference point
CDELT1  =                 -1.0 / [deg] Coordinate increment at reference point
CDELT2  =                  1.0 / [deg] Coordinate increment at reference point
CUNIT1  = 'deg'                / Units of coordinate increment and value
CUNIT2  = 'deg'                / Units of coordinate increment and value
CTYPE1  = 'RA---CAR'           / Right ascension, plate caree projection
CTYPE2  = 'DEC--CAR'           / Declination, plate caree projection
CRVAL1  =                  0.0 / [deg] Coordinate value at reference point
CRVAL2  =                  0.0 / [deg] Coordinate value at reference point
LONPOLE =                  0.0 / [deg] Native longitude of celestial pole
LATPOLE =                 90.0 / [deg] Native latitude of celestial pole
MJDREF  =                  0.0 / [d] MJD of fiducial time
RADESYS = 'ICRS'               / Equatorial coordinate system
BUNIT   = ''
END
```
So `NAXIS1` has 100 points and seems to be intended to be the RA axis.  However `CRPIX1`  seems to be associated with the other axis.  Note that the files generated by the `ops_pixels_wcs` tests do seem to have correct order of the header keys (e.g. `toast_output/ops_pointing_wcs/test_wcs_CAR_center.fits`)

src/toast/pixels_io_wcs.py
```diff
@@ -202,16 +191,18 @@ def collect_wcs_submaps(pix, comm_bytes=10000000):
 
 
 @function_timer
-def write_wcs_fits(pix, path, comm_bytes=10000000, report_memory=False):
-    """Write pixel data to a FITS image
+def write_wcs_parallel(
```
tskisner: I dislike this name change.  Toast is a natively parallel framework and the default I/O routines work with the distributed `PixelData` object and perform I/O in parallel.  For the healpix case we have:
```
read_healpix_fits()
write_healpix_fits()
read_healpix_hdf5()
write_healpix_hdf5()
```
All of which perform "parallel" I/O.  The "serial" healpix functions (`read_healpix()` and `write_healpix()`) are the special case, and the only reason they exist is to handle the HDF5 format.  For FITS format those just call healpy functions.

Can we please keep the WCS function names consistent with the healpix ones?
```
read_wcs_fits()
write_wcs_fits()
read_wcs_hdf5()
write_wcs_hdf5()
```

tskisner (changes requested):
Thanks for this large amount of work to try and achieve consistency with `pixell` file formats.  The transpose of array ordering from FITS style is a big change, but as long as everything works and is consistent with other tools, then great.  I verified that I can load some of the unit test output FITS files with DS9.

The only change that I would like to request is to revert the naming of the WCS I/O routines so that they match the healpix ones.

src/toast/pixels_io_wcs.py
keskitalo: I agree with the logic of removing "parallel" from the I/O routine names.  However, the WCS/HEALPix situation is not entirely symmetric since we are not using a 3rd party methods like `healpy.read_map()` and `healpy.write_map()` with WCS.  I suppose we could use `pixell.enmap` if we wanted to achieve symmetry.

I'm hesitant to break up `write_wcs_parallel()` into `write_wcs_fits()` and `write_wcs_hdf5()` since it would result in code duplication. Instead, we should pursue similar code reuse in the HEALPix methods. How about:
```
write_wcs_parallel() => write_wcs()         # support FITS/HDF5 
write_wcs()          => write_wcs_serial()  # support FITS/HDF5
read_wcs_parallel()  => read_wcs()          # support FITS/HDF5
read_wcs()           => read_wcs_serial()   # support FITS/HDF5
```

The HEALPix serial helper methods, `read_healpix()` and `write_healpix()` need very general names since `healpy` cannot read the TOAST `hdf5` format maps.  This is not true for WCS, where `enmap` supports FITS and HDF5.

tskisner: Ok, so just to clarify, your counter proposal is to have the MPI and serial functions in both WCS and healpix support both FITS and HDF5 (based on the filename extension):

## PixelData I/O, to either FITS or HDF5
```
write_wcs()
read_wcs()
write_healpix()
read_healpix()
```

## Serial helper functions
Using `healpy.read_map()`, `healpy.write_map()`, `pixell.read_map()`, `pixell.write_map()`.  The healpix HDF5 case would use our custom code since that is not supported upstream in healpy.
```
write_wcs_serial()
read_wcs_serial()
write_healpix_serial()
read_healpix_serial()
```

keskitalo: That would restore the symmetry but `read_healpix_serial()` sounds clunky for external users who only want to load TOAST HDF5 maps.

I don't know what is the cleanest solution to this. How important do you think it is to have perfect symmetry in the method names? Could we append `_distributed` to the collective operations and leave the serial methods without suffixes?

tskisner:
I made a proposed change in #836 , which defines `PixelData.read()` and `PixelData.write()` methods for parallel I/O.  Then the serial functions can have the simpler naming conventions.

src/toast/pixels_io_wcs.py
tskisner: This is resolved with the merge of #836

PR #831, opened 2025-05-30
Added check on max acceleration of sine elnod

AERAdler:


keskitalo (approved):
Many thanks for adding this important check!

PR #834, opened 2025-06-16
Fully functional implementation

keskitalo:
Simple script to compare two TOAST configs and print out the differences.  Only operators or templates that are enabled in both configs are compared.

Here is an example output:
```
$ toast_config_compare config_log_carlos.toml config_log_baichiang.toml
Parsing config_log_carlos.toml
Parsing config_log_baichiang.toml
Comparing
                                                config_log_carlos.toml                                   config_log_baichiang.toml
operators
  demod_noise_estim_fit                                       DISABLED                                                     ENABLED
  diff_noise_cut_flag                                         DISABLED                                                     ENABLED
  gainscrambler                                                ENABLED                                                    DISABLED
  mlmapmaker
    area   /global/cfs/projectdirs/sobs...etry_v20250306_lat_f090.fits   /home/bai-chiang/devel/pwg-...geometry_v20241015_2p0.fits
    checkpoint_interval                                              0                                                          20
    downsample                                                  [2, 1]                                                         [1]
    maxerr                                                       1e-06                                                       1e-08
    maxiter                                                 [200, 200]                                                      [1000]
    nmat_dir                                                      None   /home/bai-chiang/devel/pwg-...out/noise/LAT_f030/o6/nmats
    nmat_mode                                                    build                                                        load
    skip_existing                                                False                                                        True
    srcsamp                                                       None                               masks/psmask_LAT_f030.car.hdf
  noise_cut_fit                                               DISABLED                                                     ENABLED
  noise_cut_flag                                              DISABLED                                                     ENABLED
  noise_estim_fit                                             DISABLED                                                     ENABLED
  pixels_healpix_radec                                         ENABLED                                                    DISABLED
  pixels_wcs_radec                                            DISABLED                                                     ENABLED
  scan_detector_map                                            NOT SET                                                    DISABLED
  scan_map                                                     ENABLED                                                    DISABLED
  scan_map_pixels                                              ENABLED                                                    DISABLED
  scan_wcs_detector_map                                        NOT SET                                                    DISABLED
  sim_atmosphere                                               ENABLED                                                    DISABLED
  sim_ground
    median_weather                                                True                                                       False
    randomize_phase                                              False                                                        True
    scan_accel_az                                         3.0 deg / s2                                                1.0 deg / s2
    scan_rate_az                                           1.5 deg / s                                                 1.0 deg / s
    use_ephem                                                     True                                                       False
    use_qpoint                                                   False                                                        True
  sim_hwpss
    fname_hwpss    /global/u2/c/chervias/.l...s/data/hwpss_per_chi.pck   /home/bai-chiang/micromamba.../ops/data/hwpss_per_chi.pck
  sim_noise                                                    ENABLED                                                    DISABLED
  sim_sss                                                      ENABLED                                                    DISABLED
templates
  fourier2d                                                   DISABLED                                                     ENABLED
```

### DIFF
--- a/src/toast/scripts/CMakeLists.txt
+++ b/src/toast/scripts/CMakeLists.txt
@@ -17,6 +17,7 @@ set(ENTRY_POINTS
     toast_obsmatrix_combine.py
     toast_obsmatrix_coadd.py
     toast_config_verify.py
+    toast_config_compare.py
     toast_merge_timings.py
     toast_plot_wcs.py
     toast_plot_healpix.py
--- a/src/toast/scripts/toast_config_compare.py
+++ b/src/toast/scripts/toast_config_compare.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2025 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+"""Load two TOAST config files and report the differences between them."""
+
+import argparse
+import re
+import sys
+import traceback
+from collections import OrderedDict
+
+from astropy.units import Quantity, Unit
+from toast.mpi import Comm, get_world
+from toast.utils import Environment, Logger, import_from_name, object_fullname
+
+import toast
+from toast.config import dump_json, dump_toml, dump_yaml, load_config
+
+
+def cropped_string(value, width):
+    """Crop long strings, Add color to select strings"""
+    try:
+        value = eval(value)
+    except (SyntaxError, NameError):
+        pass
+    # Parse Astropy quantities and units
+    # if re.match(r"^Quantity.*", value) is not None:
+    #     value = eval(value)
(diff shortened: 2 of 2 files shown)


tskisner (approved):
This is a really useful tool!  I know we talked offline about the use of `eval()`.  Have you considered using the [toast helper function ](https://github.com/hpc4cmb/toast/blob/e9c516f72cedaf2894a2b372ffe0303fcdf21f98/src/toast/trait_utils.py#L110) to convert string values to containers / scalars?  That way we would have a single place where `eval()` is used and could work to make that safer in the future.

Also, this tool will only work with already-parsed configs (such as log files).  I suppose we could eventually have a separate tool that took a set of input config files and command line overrides and just spit out the final config.

keskitalo:
Great point about using the TOAST-provided string parser.

As for the command line inputs: to run in that mode, we need the original workflow that declares the operators and their defaults to do that. That is a very different use case than comparing the config logs _after_ the workflows have run.  I suppose we still support that in the sense that users can execute a workflow solely for writing out a complete configuration.

PR #835, opened 2025-06-18
This work fixes the spt3g code in light of recent changes to intervals.

tskisner:
- Add the new PyPI spt3g package to the CI test environment.
- Fix stale docstring.
- When redistributing intervals, use extra care with timestamps that fall fall on process boundaries.
- Fix off-by-one when exporting intervals, due to new exclusive last sample.

### DIFF
--- a/src/toast/intervals.py
+++ b/src/toast/intervals.py
@@ -56,7 +56,7 @@ class IntervalList(Sequence, AcceleratorObject):
         timestamps (array):  Array of local sample times, required.
         intervals (list):  An existing IntervalsList or raw intervals array.
         timespans (list):  A list of tuples containing start and stop times.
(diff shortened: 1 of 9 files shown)


src/toast/intervals.py
```diff
@@ -56,7 +56,7 @@ class IntervalList(Sequence, AcceleratorObject):
         timestamps (array):  Array of local sample times, required.
         intervals (list):  An existing IntervalsList or raw intervals array.
         timespans (list):  A list of tuples containing start and stop times.
-        samplespans (list):  A list of tuples containing first and last (inclusive)
+        samplespans (list):  A list of tuples containing first and last (exclusive)
```
keskitalo, line 59: Nice catch!

keskitalo (approved):
Looks good!

PR #836, opened 2025-06-18
Move I/O routines for the PixelData class.

tskisner:
This creates `read()` and `write()` methods for the PixelData class that can work with both WCS and Healpix, as well as HDF5 and FITS formats.  These use the existence of the PixelDistribution.wcs member to distinguish between WCS and Healpix data, and they use the filename extension to determine the format.

The serial I/O routines can now claim the shorter names (read_wcs, write_healpix, etc) without conflicts.  The low-level helper functions for gathering / broadcasting map chunks are left in their current place outside the PixelData class, since that class is becoming large.


tskisner:
I ran tests locally, but if we merge this into #828 we can deal with additional surprises.

src/toast/ops/crosslinking.py
```diff
@@ -258,9 +257,7 @@ def _exec(self, data, detectors=None, **kwargs):
         # Write out the results
 
         fname = os.path.join(self.output_dir, f"{self.name}.fits")
-        write_healpix_fits(
-            data[self.crosslinking_map], fname, nest=self.pixel_pointing.nest
-        )
+        data[self.crosslinking_map].write(fname, healpix_nest=self.pixel_pointing.nest)
```
keskitalo: I wonder if we even need to pass the HEALPix ordering here. Shouldn't the PixelData object know its ordering?

src/toast/ops/filterbin.py
```diff
@@ -1470,11 +1460,10 @@ def _bin_map(self, data, detectors, filtered):
...
                                     )
                                     continue
-                            write_healpix_hdf5(
-                                data[key],
+                            data[key].write(
                                 fname,
-                                nest=self.binning.pixel_pointing.nest,
-                                force_serial=self.write_hdf5_serial,
+                                healpix_nest=self.binning.pixel_pointing.nest,
+                                hdf5_force_serial=self.write_hdf5_serial,
```
keskitalo: We can just as well call it `force_serial` and apply it to HDF5 and HEALpix however it makes sense.

src/toast/pixels.py
```diff
@@ -1162,6 +1170,529 @@ def broadcast_map(self, fdata, comm_bytes=10000000):
...
+                    else:
+                        # All others wait for their turn to send
+                        if sendbuffer is not None:
+                            dist.comm.Send(sendbuffer, dest=0, tag=rank)
+            else:
+                msg = f"Could not determine file type for '{path}'"
+                raise RuntimeError(msg)
+
+    @function_timer
+    def read(
```
keskitalo: This is a very long method and would benefit from being refactored into a thin wrapper around HEALPix and WCS methods. That should not hold off on merging, just an observation.

keskitalo (approved):
This is a great PR. It makes external I/O methods into members of the `PixelData` class, exactly as object-oriented design would dictate.

src/toast/ops/crosslinking.py
tskisner: Currently the only way we know that a PixelData object is healpix or WCS is that the [PixelDistribution has the wcs appended](https://github.com/hpc4cmb/toast/blob/4cecad736ba71b2586861c602b6e8d0f923cf799/src/toast/ops/pixels_wcs.py#L604-L605) if it was built from a wcs projection.  Currently the healpix pointing operator does not add information about NEST ordering, but [it could do that here](https://github.com/hpc4cmb/toast/blob/4cecad736ba71b2586861c602b6e8d0f923cf799/src/toast/ops/pixels_healpix/pixels_healpix.py#L288-L293) if we wanted.

src/toast/ops/filterbin.py
tskisner: The only combination that supports parallel is healpix/HDF5.  The other combinations (healpix/FITS, wcs/FITS, wcs/HDF5) do not support parallel I/O.  But I agree we could rename it just "force_serial".

src/toast/pixels.py
tskisner: I agree- we could have internal methods for the different implementations.

src/toast/ops/crosslinking.py
tskisner: This is now implemented, and the nest/ring ordering is stored in the PixelDistribution similar to the WCS.  This removes the need for specifying the ordering in the read / write methods.

src/toast/ops/filterbin.py
tskisner: This is now implemented.

src/toast/pixels.py
tskisner: I have broken up the read / write methods to be thin wrappers around the specific implementations of pixelization and file formats.

PR #838, opened 2025-06-24
Add a more helpful error message for missing dependency

keskitalo:
Trivial fix to add a more helpful error message to an error condition I have met more than once.


tskisner (approved):
Thanks!

PR #839, opened 2025-06-25
Add binned ground template to filterbin

keskitalo:
Add option to use a binned ground template instead of Legendre polynomials in FilterBin. It won't be any cheaper to evaluate but makes the observation matrix more sparse.


tskisner (approved):
Looks good to me.

PR #840, opened 2025-06-26
Demod purepol

keskitalo:
Original work by @AERAdler with tweaks by @keskitalo .

With this PR it is possible to demodulate and downsample HWP data into I, Q/U or I/Q/U streams that are all consistent. One can demodulate and map I and Q/U separately and assemble an IQU map that matches combined I/Q/U demodulation and mapping.


tskisner (approved):
These changes and additions look fine to me

PR #841, opened 2025-06-29
Change interactive startup function to handle slurm

tskisner:
If running in slurm, this changes the MPI commands to use `srun`.

### DIFF
--- a/src/toast/interactive/startup.py
+++ b/src/toast/interactive/startup.py
@@ -4,6 +4,7 @@
 """Tools for starting up interactive sessions."""
 
 import os
+import shutil
 
 from IPython.core.extensions import ExtensionManager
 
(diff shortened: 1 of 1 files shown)


src/toast/interactive/startup.py
```diff
@@ -60,6 +76,8 @@ def start_parallel(procs=1, threads=1, nice=True, auto_mpi=False, shell=None):
     if procs == 1:
         rank = 0
         comm = None
+        os.environ["MPI_DISABLE"] = "1"
+        os.environ["DISABLE_MPI"] = "1"
```
keskitalo, line 80: Just curious... Why do we need both permutations?

keskitalo (approved):
Seems straightforward.

src/toast/interactive/startup.py
tskisner: `pixell` uses the other permutation :-)

PR #843, opened 2025-07-01
Store Az min/max metadata in the AzimuthIntervals operator

tskisner:
This fixes the use of the FilterBin operator on real data, which does not have the `scan_min_az` and `scan_max_az` metadata keys generated by the SimGround operator.


keskitalo (changes requested):
This solution may fail if we are scanning across az=0. For a fully general solution we would need to

1. collect a full azimuth vector on the root process
2. unwrap the vector
3. then find the min/max of the unwrapped azimuth angle

PR #844, opened 2025-07-02
Move azimuth range calculation into a separate operator

keskitalo:


tskisner (approved):
Thank you, much cleaner solution!

PR #846, opened 2025-07-03
Allow data distributions where some processes have no data.

tskisner:
This situation may occur when demodulating data with cut detectors or situations where having many processes is useful for simulation operations that are independent of the number of detectors.  Specific changes include:

- In distribute_discrete(), lift the restriction on having more groups than work units.

- In the DetectorData class, handle the case where there are no local detectors.

- Remove various warning / errors for this situation

- Add unit tests that test this "over distribution" of data with more processes than detectors.

### DIFF
--- a/src/toast/dist.py
+++ b/src/toast/dist.py
@@ -70,15 +70,14 @@ def distribute_discrete(sizes, groups, pow=1.0, breaks=None):
     weights = None
     max_per_proc = None
     target = None
-    if len(sizes) < groups:
-        msg = (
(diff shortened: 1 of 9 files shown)


tskisner:
In addition to the unit test coverage, I also ran this branch on one observation of real data where more than half the detectors were cut.  After passing through demodulation, roughly 32 / 64 process had no data.  This job completed all the way through mapmaking without issues.

keskitalo (approved):
Straightforward but tedious modification. Thanks Ted!

PR #847, opened 2025-07-03
Retain the HWP angle information in the demodulated data

keskitalo:


tskisner (approved):
This is a good catch- we definitely need the HWP angle in the demodulated data.  Thanks!

PR #848, opened 2025-07-09
More robust handling of samples that fail to filter

keskitalo:
When filter flags prevent a template from being applied, flag all affected samples so they are not projected onto a map.

This is relevant when filters are subject to a processing mask that is not applied in map projection. Certain subscans may be entirely covered by the processing mask and escape subscan filtering. Unfiltered samples are subject to large uncertainties and should not be projected.


tskisner (approved):
These changes look good and the improved flagging is great.

PR #854, opened 2025-08-18
Obsmat corner case

keskitalo:
Fix a very rare situation where two subsequent detectors have the same flagging but different number of filtering templates.


tskisner (approved):
Looks good

PR #857, opened 2025-08-29
When converting coordinates with qpoint, make use of weather info

tskisner:
The qpoint conversion from AZ / EL to RA / DEC can make use of the temperature, pressure, and humidity of the environment.  If a weather object is associated with a site, make use of this information.


keskitalo (approved):
Excellent! Consistency with sotodlib.

PR #861, opened 2025-09-09
Add equivalencies to astropy temperature conversions.

tskisner:
For conversion of normal temperatures between Kelvin and Celsius, make sure we specify the equivalencies to use.


src/toast/tests/ops_sim_ground.py
```diff
@@ -172,6 +172,69 @@ def test_exec(self):
 
         close_data(data)
 
+    def test_qpoint(self):
```
keskitalo, line 175: This incomplete test was probably committed by mistake

keskitalo (approved):
Did not know we needed these until now.

src/toast/tests/ops_sim_ground.py
tskisner: This test just applies the `SimGround` operator with simulated weather.  This weather has temperature information (in Kelvin) which is converted to Celsius in the qpoint calculation for the AzEl to RaDec conversion.  This is already enough to trigger an error before the unit equivalencies fix.

PR #862, opened 2025-09-10
Fix argument types and add optional rotation

keskitalo:


tskisner (approved):
Looks like a straightforward change.

PR #863, opened 2025-09-12
Add a simple operator to apply a "counter 1/f" filter.

tskisner:
For (at least) 25 years, the inverse time domain noise covariance has been used as a "change of basis" when solving a generalized least squares problem for the maximum likelihood map in pixel space.

For simpler "filter and bin" mapmaking techniques, application of this N_tt'^-1 filter may be useful.  This work adds an operator which uses a given noise model for detector PSDs to compute the inverse noise covariance fourier domain kernels for each detector and convolve the timestreams by those.

### DIFF
--- a/src/toast/ops/noise_model.py
+++ b/src/toast/ops/noise_model.py
@@ -104,6 +104,65 @@ class DefaultNoiseModel(Operator):
         return prov
 
 
+def estimate_net(freqs, data):
+    """Estimate the NET from the high frequency PSD.
+
+    This assumes that at high frequency the PSD has a white noise "plateau".  A simple
+    parabola is fit to the last bit of the spectrum and this is used to compute the
+    NET.
+
+    Args:
+        freqs (array):  The frequency values in Hz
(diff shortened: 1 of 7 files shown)


src/toast/ops/noise_model.py
```diff
@@ -104,6 +104,65 @@ def _provides(self):
...
+                quad_func, ffreq, fdata, p0=[1.0, ffreq[-1], fdata[-1]]
+            )
+            # It worked!
+            fdata = quad_func(ffreq, params[0], params[1], params[2])
+            fdata = np.exp(fdata)
+            return np.sqrt(fdata[-1])
+        except RuntimeError:
+            pass
+
+    params, params_cov = curve_fit(
```
keskitalo, line 157: I have a hunch that `np.polyfit()` would have worked here too, but it really doesn't matter.

keskitalo (approved):
An elegant implementation of the classic 1/f filter.

PR #864, opened 2025-09-17
Add a generic "toast_run" workflow for arbitrary nested Pipelines.

tskisner:
In practice, many workflows can be expressed as nested sequences (toast.ops.Pipeline) instances.  Internal references between objects in config files have been supported for years, but using these in complex situations required a couple of small fixes.

- Add toast_run script, which looks for a single "main" operator and runs it.

- Add HDF5 helper functions for saving / loading just the instrument models in the same format as used in HDF5 Volumes.

- In SimGround and SimSatellite, add support for loading instrument models and schedule files as traits.

- Expand unit tests for nested pipelines in config files.

### DIFF
--- a/src/toast/config/cli.py
+++ b/src/toast/config/cli.py
@@ -20,7 +20,7 @@ from tomlkit.exceptions import TOMLKitError
 from ..trait_utils import scalar_to_string as trait_scalar_to_string
 from ..trait_utils import string_to_scalar as trait_string_to_scalar
 from ..trait_utils import string_to_trait, trait_to_string
-from ..utils import Environment, Logger
+from ..utils import Environment, Logger, import_from_name
 from .json import dump_json, load_json
 from .toml import dump_toml, load_toml
 from .yaml import dump_yaml, load_yaml
@@ -81,10 +81,11 @@ def load_config(file, input=None, comm=None):
         return ret
 
 
-def dump_config(file, conf, format="toml", comm=None):
+def dump_config(file, conf, format=None, comm=None):
     """Dump a config file to a supported format.
 
-    Writes the configuration to a file in the specified format.
+    Writes the configuration to a file in the specified format.  If `format` is
+    None, the format with be determined from the file name.
 
     Args:
         file (str):  The file to write.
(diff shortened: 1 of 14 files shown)


src/toast/config/cli.py
```diff
@@ -361,53 +375,57 @@ def args_update_config(args, conf, useropts, section, prefix="", separator=r"\."
...
-
-        * Adds "--default_toml" and "--default_json" options to dump the default config
-          options to files.
-
-        * Adds a option "--job_group_size" to provide the commonly used option for
-          setting the group size of the TOAST communicator.
-
-        * Adds arguments for all object parameters in the defaults for the specified
-          classes.
+def add_job_parser_options(parser):
```
keskitalo, line 402: I really like the refactor.

src/toast/config/cli.py
```diff
@@ -361,53 +375,57 @@ def args_update_config(args, conf, useropts, section, prefix="", separator=r"\."
...
 
-    Returns:
-        (tuple):  The (config dictionary, args).  The args namespace contains all the
-            remaining parameters after extracting the operator and template options.
 
-    """
+def add_default_parser_options(parser, prefix, operators, templates):
+     # Add options to dump default values
+    parser.add_argument(
+        "--defaults_toml",
```
keskitalo: This feels a little clunky to me. If `dump_config` uses the suffix to determine the format, then we should use that approach here as well.

src/toast/config/cli.py
```diff
@@ -439,84 +457,34 @@ def parse_config(
...
-    )
-
-    # Parse commandline or list of options.
-    if opts is None:
-        opts = sys.argv[1:]
-    args = parser.parse_args(args=opts)
 
+def process_default_args(args, defaults):
     # Dump default config values.
     if args.defaults_toml is not None:
```
keskitalo: This where we would check the default file format.

src/toast/traits.py
```diff
@@ -526,26 +526,34 @@ def from_config(name, props):
...
-            # print(f"from_config: parsing {v}", flush=True)
-            if v["value"] == "None":
-                # Regardless of type, we set this to None
-                kw[k] = None
-            elif v["type"] == "unknown":
-                # This was a None value in the TOML or similar unknown object
-                pass
-            elif v["type"] in parsable:
-                kw[k] = string_to_trait(v["value"])
+            # print(f"CONF {name}: parsing {v}", flush=True)
```
keskitalo, line 537: Could probably delete the debug statements.

keskitalo (approved):
This PR adds an elegant way of defining the entire workflow from the configuration files and command line. It does complicate the use of `SimGround()` and `SimSatellite()` operators that require precomputed `Telescope` objects on disk when running in this mode. All the necessary facilities are provided and appropriate unit tests are added.

src/toast/traits.py
tskisner: I once deleted those and then had to add them all back during a subsequent refactor.  Now I just leave them in and un-comment them whenever there is a change that requires a deep dive into exploring why a particular parsing operation fails.

src/toast/config/cli.py
tskisner: I moved the file format parsing / detection to a single function and now call that from both `dump_config()` and `load_config()`.  I changed this function to just take a single argument ("--defaults") and then the format is determined by the file name.

keskitalo: Great!

PR #867, opened 2025-09-23
Add support for in-place demodulation of data

tskisner:
When running inside a larger pipeline with the `toast_run` script, dynamically changing to new data objects is challenging.  The only place where this happens is in the demodulation operator.  This work add support for replacing observations with their demodulated versions inside an existing Data object.


tskisner:
I tested this feature in a `toast_run` pipeline on one observation of real SO data.

keskitalo (approved):
This is a great feature. All too often the input data object just goes stale and is never used.

PR #869, opened 2025-09-26
Fix demodulation of shared data

tskisner:
Some small fixes found when testing branch on larger data.


keskitalo (approved):
Awesome. I just hit this issue in demodulation work as well.

PR #870, opened 2025-09-30
Variety of small fixes.

tskisner:
- Fixes to interactive startup files
- Cleanup of inplace demodulation
- Fix typo in toast_run
- Fix verbose logging in noise model cuts
- Remove HWPSS destriping template, which is not useful in the context of demodulated data


src/toast/ops/demodulation.py
```diff
@@ -129,6 +129,11 @@ class Demodulate(Operator):
         "hamming", help="Window function name recognized by scipy.signal.firwin"
     )
 
+    keep_dets_frac = Float(
```
keskitalo, line 132: I don't see the rationale of this being a part of demodulation but I do appreciate the functionality.

src/toast/ops/demodulation.py
```diff
@@ -211,20 +216,46 @@ def _exec(self, data, detectors=None, **kwargs):
             self.demod_data = Data()
 
         # Demodulation only applies to observations with HWP.  Verify
-        # that there are such observations in `data`
+        # that there are such observations in `data`.  We also cut all
+        # observations where more than 90% of optical detectors are cut.
```
keskitalo: Should reference `self.keep_dets_frac`, not 90%.

src/toast/ops/noise_model.py
```diff
@@ -834,7 +835,7 @@ def _exec(self, data, detectors=None, **kwargs):
                     }
                     msg = f"obs {obs.name}|{group}: flagged {len(group_flags)}"
                     msg += " outlier detectors"
-                    log.info(msg)
+                    log.debug(msg)
```
keskitalo: I think I would still like some sort of tally of bad detectors at the `info` level. Perhaps not per observation. That can be added at some point later.

keskitalo (approved):
Sensible fixes

src/toast/ops/demodulation.py
tskisner: That is a good point- perhaps we should have a dedicated (and more flexible) operator that cuts data at the observation level.  One could even introduce a "per-observation flag" bitmask similar to the per-detector and per-sample ones.

The only motivation for putting the cut in this operator is that we are creating whole new observations, and so that is a convenient time to remove bad ones.  I guess for now we can leave this feature and then add a new operator with more involved cutting later.

src/toast/ops/noise_model.py
tskisner: Fair enough, I was perhaps too aggressive in trying to reduce the logging :-)

PR #871, opened 2025-09-30
Add infrastructure to store additional metadata to HEALPix and WCS he…

keskitalo:
Add support for appending metadata to HEALPix and WCS headers.

The `extra_header` keyword argument defaults to `None` but expects dictionaries of
```
key : value
```
or
```
key : (value, comment)
```

Currently, only the total number of detectors and the start and stop times are recorded but it will be trivial to add
- TOAST version
- TOAST operator
- execution time
- details about the compute system and software environment


tskisner (approved):
This is a great feature addition!

PR #872, opened 2025-09-30
Detect anomalously low noise detectors

keskitalo:


tskisner (approved):
Looks good.  This should help cut those detectors which are clearly dead or have issues.

PR #873, opened 2025-10-01
Improve `toast_project_schedule`

keskitalo:
User can now define lists of SSOs and their radii to plot with the hits and background


tskisner (approved):
Looks good, thanks!

PR #874, opened 2025-10-03
Add support for reading / writing Weather base class objects.

tskisner:
This small change allows serializing the Weather base class to toast HDF5 files.


keskitalo (approved):
Thanks! This is a big help.

PR #876, opened 2025-10-03
Support past, current and future datetime modules

keskitalo:


tskisner (approved):
Looks good.  Some day the minimum supported version will be 3.11, but not yet :-)

PR #877, opened 2025-10-06
Fix weather serialization to HDF5

tskisner:
- Parse the timestamp separately
- Dump / load all parameters as Quantities
- Use the base weather class in one of the unit tests.


keskitalo (approved):
Looks good. Thanks for the fix!

PR #879, opened 2025-10-10
Add stdout / stderr redirection to toast_run

tskisner:
- Add a new context manager that uses wurlitzer to redirect both python and libc output to a file. Per-process output is written to separate files and then collated at the end.

- Use redirection in toast_run so that all output is logged to the top level out_dir.

- Several small fixes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -305,12 +305,12 @@ conf["install_requires"] = [
     "astropy",
     "healpy",
     "ephem",
+    "wurlitzer",
 ]
 conf["extras_require"] = {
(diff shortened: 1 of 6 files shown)


keskitalo (approved):
A very useful addition.

tskisner:
I will wait to merge this until the upstream [pshmem PR is merged](https://github.com/tskisner/pshmem/pull/30) and a tag has been made.  That way we can bump the pshmem version requirements to ensure that the `MPIBatch` class is available.

PR #882, opened 2025-10-15
Allow user to disable log redirection

tskisner:


tskisner:
We could also just add a boolean flag to disable redirection, rather than overload the log file name to indicate that.

keskitalo (approved):
Works for me, although I suspect an average user would not expect redirection so we might want to make that the default.

tskisner:
> Works for me, although I suspect an average user would not expect redirection so we might want to make that the default.

I guess I had primarily thought about this being something run from a batch job...  But I guess as long as we can do the redirection from higher-level applications that are calling `toast_run.main()`, I don't care what we set for the default.

Perhaps the `--out_log_name` could default to `None` (indicating no redirection) and if it is set to any string, that is name of the log file to redirect to in the output directory?

PR #883, opened 2025-10-15
Derivatives and beams

AERAdler:
Operator to create weights for higher-order derivatives of the map in a "beam projection" approach


keskitalo:
I suggest you rebase against `toast3` to resolve the conflicts.

PR #884, opened 2025-10-16
Highpass before demod

keskitalo:
This PR extends TOAST capabilities to handle SO data.

- `Demodulate` now isolates the 2f and 4f signals with a user-configured bandpass filter
- there is a new operator, `SimpleStatCut`, that measures the median-filtered skew and kurtosis and uses them to reject outlier detectors
- `CalibrateDetectors` can retrieve the gains from the focalplane database
- 'SimpleDeglitch' works on demodulated data and will flag all affected streams

### DIFF
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2024 by the parties listed in the AUTHORS file.
+# Copyright (c) 2019-2025 by the parties listed in the AUTHORS file.
 # All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
 
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -760,6 +760,10 @@ class Focalplane(object):
         return list(self._det_to_row.keys())
 
     @property
+    def properties(self):
+        return list(self.detector_data.colnames)
+
+    @property
     def n_detectors(self):
         return len(self._det_to_row.keys())
 
--- a/src/toast/ops/simple_statcut.py
+++ b/src/toast/ops/simple_statcut.py
@@ -114,11 +114,6 @@ class SimpleStatCut(Operator):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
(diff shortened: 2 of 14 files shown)


src/toast/instrument.py
```diff
@@ -759,6 +759,10 @@ def __setitem__(self, key, value):
     def detectors(self):
         return list(self._det_to_row.keys())
 
+    @property
+    def properties(self):
```
tskisner, line 763: A useful shortcut!

src/toast/ops/simple_statcut.py
```diff
@@ -0,0 +1,276 @@
...
+            raise traitlets.TraitError("medfilt_kernel_size cannot be even")
+        return check
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.net_factors = []
+        self.total_factors = []
+        self.weights_in = []
+        self.weights_out = []
+        self.rates = []
```
tskisner: It looks like these lines are left over from a copy / paste?

src/toast/tests/ops_statcut.py
```diff
@@ -0,0 +1,81 @@
...
+from astropy import units as u
+
+from .. import ops as ops
+from ..observation import default_values as defaults
+from ..vis import set_matplotlib_backend
+from .helpers import close_data, create_outdir, create_satellite_data
+from .mpi import MPITestCase
+
+
+class StatisticsTest(MPITestCase):
```
tskisner, line 19: These tests are useful for making sure that the operator executes.  In the future, we might consider another test with a specific number of expected outlier detectors and verify that exactly those detectors are flagged.

tskisner (approved):
Thanks for these great improvements!  One small comment about stale lines of code.

src/toast/ops/simple_statcut.py
keskitalo: Thanks!

src/toast/tests/ops_statcut.py
keskitalo: Definitely.

PR #886, opened 2025-10-22
Implement ops.Detrend

Bai-Chiang:
Simple detrend operator, removing mean/median/slope

Based on `sotodlib.tod_ops.detrend_tod` and `pixell.utils.deslope`. But ignoring flagged samples when calculate mean, median, slope.

### DIFF
--- a/src/toast/ops/detrend.py
+++ b/src/toast/ops/detrend.py
@@ -1,0 +1,200 @@
+# Copyright (c) 2025-2025 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+import os
+
+import numpy as np
+import traitlets
+
+from ..observation import default_values as defaults
+from ..timing import function_timer
+from ..traits import Int, Unicode, trait_docs
+from ..utils import Logger
+from .operator import Operator
+
+@trait_docs
+class Detrend(Operator):
+    """Remove mean/median/slope of detector data.
+
+    method='mean'   will subtract mean of each detector data.
+
+    method='median' will subtract median of each detector data.
+
+    method='linear' will subtract mean and slope of each detector data,
+    matching `match_nsample` samples at the beginning and end of detector data.
+    """
+
+    # Class traits
+
+    API = Int(0, help="Internal interface version for this operator")
+
+    det_data = Unicode(
+        defaults.det_data,
+        help="Observation detdata key apply detrend to",
+    )
+
+    det_mask = Int(
+        defaults.det_mask_invalid,
+        help="Bit mask value for per-detector flagging",
+    )
+
+    shared_flags = Unicode(
+        defaults.shared_flags,
+        allow_none=True,
+        help="Observation shared key for telescope flags to use",
+    )
+
+    shared_flag_mask = Int(
+        defaults.shared_mask_invalid,
+        help="Bit mask value for optional shared flagging",
+    )
+
+    det_flags = Unicode(
+        defaults.det_flags,
+        allow_none=True,
+        help="Observation detdata key for flags to use",
+    )
+
+    det_flag_mask = Int(
+        defaults.det_mask_nonscience,
+        help="Bit mask value for detector sample flagging",
+    )
+
+    detrend_flag_mask = Int(
+        defaults.det_mask_invalid,
+        help="Bit mask to use when det_data cannot be detrended.",
+    )
+
+    method = Unicode(
+        "linear",
+        help="Method to detrend. Valid methods are 'linear', 'mean', 'median'",
+    )
+
+    edge_nsample = Int(
+        100,
+        help="When using 'linear' method, number of samples to calculate mean level at the edge",
+    )
+
+    edge_nsample_method = Unicode(
+        'mean',
+        help="When using 'linear' method, method to calculate mean level at the edge. "
+        "Valid methods are 'mean' and 'median'",
+    )
+
+    @traitlets.validate("shared_flag_mask")
+    def _check_shared_flag_mask(self, proposal):
+        check = proposal["value"]
+        if check < 0:
+            raise traitlets.TraitError("Shared flag mask should be a positive integer")
+        return check
+
+    @traitlets.validate("det_flag_mask")
+    def _check_det_flag_mask(self, proposal):
+        check = proposal["value"]
+        if check < 0:
+            raise traitlets.TraitError("Det flag mask should be a positive integer")
+        return check
+
+    @traitlets.validate("method")
+    def _check_method(self, proposal):
+        check = proposal["value"]
+        allowed = ["mean", "linear", "median"]
+        if check not in allowed:
+            msg = f"method must be one of {allowed}, not {check}"
+            raise traitlets.TraitError(msg)
+        return check
+
+    @traitlets.validate("edge_nsample")
+    def _check_edge_nsample(self, proposal):
+        check = proposal["value"]
+        if check < 1:
+            raise traitlets.TraitError("edge_nsample should be an integer >= 1")
+        return check
+
+    @traitlets.validate("edge_nsample_method")
+    def _check_edge_nsample(self, proposal):
+        check = proposal["value"]
+        allowed = ["mean", "median"]
+        if check not in allowed:
+            msg = f"method must be one of {allowed}, not {check}"
+            raise traitlets.TraitError(msg)
+        return check
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    @function_timer
+    def _exec(self, data, detectors=None, **kwargs):
+        log = Logger.get()
+
+        for ob in data.obs:
+            local_dets = ob.select_local_detectors(selection=detectors, flagmask=self.det_mask)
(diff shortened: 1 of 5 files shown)


keskitalo:
I see there are some edits since you asked for the review. Is the PR now stable? If not, can you ping me when you think it is ready for review?

Bai-Chiang:
Yes

keskitalo:
:-D Is that "yes, the PR is stable and ready for review" or "yes, I will ping you when it is ready"?

Bai-Chiang:
yes, the PR is stable and ready for review (×﹏×)

src/toast/ops/detrend.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    det_flags = Unicode(
+        defaults.det_flags,
+        allow_none=True,
+        help="Observation detdata key for flags to use",
+    )
+
+    det_flag_mask = Int(
+        defaults.det_mask_invalid,
```
keskitalo: This is certainly an application where we would want to disregard planets, point sources and bright Galactic emission. The default flag mask should be `defaults.det_mask_nonscience`: https://github.com/simonsobs/LAT-scan-schedules/blob/main/iso/phase3/2025-10-22T21%3A36%3A11%2B00%3A00_phase3_type3_lat_field_schedule.txt

src/toast/ops/detrend.py
```diff
@@ -0,0 +1,172 @@
...
+        help="Bit mask value for detector sample flagging",
+    )
+
+    method = Unicode(
+        "linear",
+        help="Method to detrend. Valid methods are 'linear', 'mean', 'median'",
+    )
+
+    match_nsample = Int(
+        1,
```
keskitalo: That seems like a very small default value... If you started with pure white noise and didn't set it to anything more sensible, you would introduce a strong trend instead of removing it.

src/toast/ops/detrend.py
```diff
@@ -0,0 +1,172 @@
...
+
+        for ob in data.obs:
+            local_dets = ob.select_local_detectors(selection=detectors, flagmask=self.det_mask)
+            shared_flags = ob.shared[self.shared_flags].data & self.shared_flag_mask
+            if len(local_dets) == 0:
+                # Nothing to do for this observation
+                continue
+            for det in local_dets:
+                sig = ob.detdata[self.det_data][det]
+                assert np.ndim(sig) == 1
```
keskitalo: @tskisner how do we feel about `assert` in TOAST? It is not typical.

src/toast/ops/detrend.py
```diff
@@ -0,0 +1,172 @@
...
+        if check not in allowed:
+            msg = f"method must be one of {allowed}, not {check}"
+            raise traitlets.TraitError(msg)
+        return check
+
+    @traitlets.validate("match_nsample")
+    def _check_match_nsample(self, proposal):
+        check = proposal["value"]
+        if check < 1:
+            raise traitlets.TraitError("match_nsample should be a integer larger than 1")
```
keskitalo: The test and the error message are in conflict.

src/toast/ops/detrend.py
```diff
@@ -0,0 +1,172 @@
...
+                good_samples_flag = np.logical_and(
+                    shared_flags == 0,
+                    (det_flags & self.det_flag_mask) == 0,
+                )
+                good_samples_i = np.flatnonzero(good_samples_flag)
+                if self.method == "mean":
+                    sig -= np.mean(sig[good_samples_flag])
+                elif self.method == "median":
+                    sig -= np.median(sig[good_samples_flag])
+                elif self.method == "linear":
```
keskitalo, line 148: The linear trend removal will produce unexpected results if `self.match_nsample` is longer than `sig`.

src/toast/tests/ops_detrend.py
```diff
@@ -0,0 +1,99 @@
...
+                    detflags[-edge_nsample:] = defaults.det_mask_invalid
+                # Flag some samples at the middle
+                for j in range(
+                        edge_nsample+len(ob.local_detectors)+100,
+                        nsample-len(ob.local_detectors)-edge_nsample-100,
+                        200):
+                    detflags[j:j+10] = defaults.det_mask_invalid
+
+        if method == 'linear':
+            for match_nsample in [1,2,3,4,5]:
```
keskitalo: Maybe add one ridiculously large value for testing?

src/toast/tests/ops_detrend.py
```diff
@@ -0,0 +1,99 @@
...
+                        detdata[detflags != 0] += 1e5
+
+                linear_detrend = ops.Detrend(method='linear', match_nsample=match_nsample)
+                linear_detrend.apply(data)
+
+                for ob in data.obs:
+                    for i, det in enumerate(ob.select_local_detectors(flagmask=defaults.det_mask_invalid)):
+                        detdata = ob.detdata[defaults.det_data][det][:]
+                        detflags = ob.detdata[defaults.det_flags][det][:]
+                        detdata[detflags != 0] -= 1e5
```
keskitalo: As a matter of form, when you need the same numeric constant more than once, it should be stored in a variable. Obviously not an issue for this unit test.

keskitalo (changes requested):
Only one real request which is to fortify the operator against users setting the number of matched samples to exceed the length of the observation.

We more or less already have this functionality in the polynomial filter that can apply to the entire observation or to each subscan. However, having an operator that exactly matches the SO site pipeline will be handy.

src/toast/ops/detrend.py
tskisner: Yes, please raise an exception instead.

src/toast/ops/detrend.py
tskisner: Another option would be to default this to `None`, and then choose the number of edge samples as a fraction of the total if the user has not set the value.  Or add an additional trait to specify the fraction.

tskisner (commented):
I agree with @keskitalo 's suggestions.  Otherwise looks good!

src/toast/ops/detrend.py
Bai-Chiang: I forgot to remove this assertion line. Is `ob.detdata[self.det_data][det][:]` always 1d array? Is there cases it become multi dim array?

src/toast/ops/detrend.py
Bai-Chiang: current behavior when having too many edge samples will skip this observation. [1] I will add a log message to tell the name of observation that skipped, and a test these cases.

[1] https://github.com/Bai-Chiang/toast/blob/8ed1662e857e06a55627e970b9dccb678b46884c/src/toast/ops/detrend.py#L138-L140

src/toast/ops/detrend.py
tskisner: The detector signal always has a per-sample shape of zero (scalar).  Other detdata quantities (pointing quaternions, stokes weights, etc) have an array of values per sample.

`ob.detdata[self.det_data]["foo"]` returns a reference to the row of the internal storage for this detector.  So you don't need the `[:]` here.

If you use numpy-style "fancy indexing" like `ob.detdata[self.det_data][["foo", "bar", "blat"], 2:10]` then it will return a **list** of references to the underlying buffers for those detectors.  This is different from numpy arrays, where fancy indexing of non-contiguous rows makes a copy into a new array.  This behavior is intentional, to make sure we do not accidentally get a copy of the detector data and then modify that copy rather than the underlying buffer.

src/toast/ops/detrend.py
keskitalo: Thanks. Somehow I missed that.

src/toast/ops/detrend.py
```diff
@@ -0,0 +1,194 @@
...
+                            good_samples_i[0],
+                            good_samples_i[0]+self.edge_nsample,
+                            1)
+                    end_match_slc = slice(
+                            good_samples_i[-1]+1-self.edge_nsample,
+                            good_samples_i[-1]+1,
+                            1)
+                    if start_match_slc.stop >= end_match_slc.start:
+                        log.info_rank(f"No enough good samples, skipping observation {ob.name}")
+                        continue
```
keskitalo, line 160: Just skipping the detector can lead to surprises down the line. If the user asked us to detrend all detectors and some cannot be detrended, then they should be flagged as unusable.

keskitalo (approved):
I'm approving the PR but please consider flagging the entire detector when it cannot be detrended.

src/toast/ops/detrend.py
Bai-Chiang: What would be the flag to set? Should we add an extra argument for this flag?

tskisner: You should set the per-detector flags to `invalid` for that detector.  That is a general policy for all operators that modify the data (poly filter, hwpss, etc).  If for some reason the processing cannot be done, then the per-detector flags should be updated to mark that detector as unusable.  [Example here](https://github.com/hpc4cmb/toast/blob/345e947d26ca2b6ce42ee47374b3b0ba8a6d6397/src/toast/ops/groundfilter.py#L468-L473).

Bai-Chiang:
Any more recommendations?

keskitalo:
No more recommendations. Solid work!

Bai-Chiang:
Thanks. I will squash all commits into one.

PR #887, opened 2025-10-22
HDF5 I/O with per-detector flags

keskitalo:


tskisner:
The added equality test for local detector flags in the `Observation` class was already merged in #897.  That PR also included the fix for the problem (saving / loading the local_detector_flags to HDF5).

PR #890, opened 2025-10-28
Restore accidental trait swap

tskisner:
The recent refactor of the TimeConstant operator accidentally swapped the meaning of the deconvolve trait.  This fixes that and also adds a unit test to verify that the convolution introduces an expected phase shift.


keskitalo (approved):
Glad to have this sorted out. Thanks for the new unit test!

PR #891, opened 2025-10-30
Fix a typo that has been causing excessive memory use

tskisner:
The function kwarg to apply an operator to a subset of detectors is `detectors=`.  Unknown arguments are intentionally passed on to derived operators so that they can handle those as needed.  So this argument was simply ignored by the `apply()` method, causing the generation of the stokes weights and detector quaternions for all detectors in memory.


keskitalo (approved):
Awesome! One step closer to perfection.

PR #892, opened 2025-11-06
Clarify per-map noise weights in coaddition

tskisner:
When specifying weights in a file with the list of maps, these weights are inverse noise weights, and each one is applied directly to the corresponding noise weighted map and the inverse covariance.

### DIFF
--- a/src/toast/scripts/toast_healpix_coadd.py
+++ b/src/toast/scripts/toast_healpix_coadd.py
@@ -286,12 +286,13 @@ def main(opts=None, comm=None):
                 inmap *= args.scale
             invcov /= args.scale**2
 
-        # Apply per-map weights
+        # Apply per-map weights.  The weights loaded from the file
+        # are assumed to be inverse noise weights for each map.
         if noiseweighted:
-            inmap /= weights[infile_map]
-        else:
             inmap *= weights[infile_map]
-        invcov /= weights[infile_map] ** 2
+        else:
+            inmap /= weights[infile_map]
+        invcov *= weights[infile_map]
 
         if not noiseweighted:
             # Must reverse the multiplication with the
--- a/src/toast/scripts/toast_healpix_coadd.py
+++ b/src/toast/scripts/toast_healpix_coadd.py
@@ -286,14 +286,6 @@ def main(opts=None, comm=None):
                 inmap *= args.scale
             invcov /= args.scale**2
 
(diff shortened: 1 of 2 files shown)


tskisner:
Hmmm, the test failures here are due to spt3g wheels not available on macos x86_64.  Will investigate tomorrow.

src/toast/scripts/toast_healpix_coadd.py
```diff
@@ -286,12 +286,13 @@ def main(opts=None, comm=None):
...
 
-        # Apply per-map weights
+        # Apply per-map weights.  The weights loaded from the file
+        # are assumed to be inverse noise weights for each map.
         if noiseweighted:
-            inmap /= weights[infile_map]
-        else:
             inmap *= weights[infile_map]
-        invcov /= weights[infile_map] ** 2
+        else:
```
keskitalo: I'm still unsure about this. I would think that regardless of earlier noise-weighting, new weights should be multiplicative.

However, if the loaded map is _not_ noise-weighted to begin with, we'll also need to apply inverse noise weights according to the inverse covariance.

tskisner: Ok, backing up a second.  This coadd tool requires either a noise weighted map (`z = P^T N^-1 d`) or a data map (`m = (P^T N^-1 P)^-1 z`).  If the final map is provided, the covariance is used to reconstruct `z`.

So our starting point is always the noise weighted map (`z`) and the inverse pixel noise covariance (`N_p^-1 = P^T N^-1 P`) that used the same time-domain noise estimate (`N^-1`).

Now let's say that we discover that the noise estimate used to build `z` and `N_p^-1` was wrong and we want to correct the noise weighting in the coadd process.  When coadding, we want the effective time-domain noise weight to be `N'^-1 = N^-1 * W`, where `W` is the (constant) correction factor.  This means that the quantities we want to coadd are:
```
z' = P^T N'^-1 d = P^T (N^-1 * W) d = z * W
N'_p^-1 = P^T N'^-1 P = P^T (N^-1 * W) P = N_p^-1 * W
```
And those `W` values are what we are reading from the file, and which are multiplied by the noise weighted map and the inverse noise covariance.

So- I think we just need to move the application of these weights a few lines down in the code, after the input map is potentially converted into a noise weighted map.

keskitalo (approved):
Perfect! I think it is consistent now.

PR #894, opened 2025-11-19
Scan map fix

keskitalo:
Fix a disagreement in assigning samples to WCS pixels between TOAST and so3g.


tskisner (approved):
Thanks for checking these details and ensuring consistency!

Bai-Chiang (approved):

PR #895, opened 2025-11-19
can filter different detector patterns: polyfilter, groundfilter

AERAdler:


tskisner (approved):
Thanks for this addition, which will allow applying different ground filters to demodulated data components.

PR #897, opened 2025-12-04
Large refactor of HDF5 support

tskisner:
- Remove compile-time dependency on libFLAC and use flacarray instead

- Convert the existing version-1 HDF5 data format to use flacarray for low-level compression / decompression, while remaining backwards compatible with the existing on-disk group and dataset structure.

- Define a new version (2) of the native HDF5 data format which makes use of flacarray and enforces consistent use of per-field compression parameters.  This new version adds support for saving / loading arbitrary observation attributes, provided they define the save_hdf5() and load_hdf5() methods.

- Fix Observation duplicate() and __eq__() methods to correctly handle arbitrary attributes which are more complicated class instances.


src/toast/io/observation_hdf_save.py
```diff
@@ -337,205 +335,68 @@ def save_hdf5_detdata(obs, hgrp, fields, log_prefix, use_float32=False, in_place
...
+            if ddtype.char == "d" or ddtype.char == "f":
+                # Floating point type
+                if "quanta" in fieldcomp:
+                    quanta = float(fieldcomp["quanta"])
+                elif "precision" in fieldcomp:
+                    precision = float(fieldcomp["precision"])
+                if quanta is None and precision is None:
+                    msg = "When compressing floating point data, you"
+                    msg += " must specify the quanta or precision."
+                    raise RuntimeError("You must specify the quanta")
```
keskitalo, line 365: Probably intended to be `raise RuntimeError(msg)`

src/toast/io/observation_hdf_save_v1.py
```diff
@@ -0,0 +1,984 @@
...
+        del hdata
+
+        log.verbose_rank(
+            f"{log_prefix}  Intervals finished {field} write in",
+            comm=comm,
+            timer=timer,
+        )
+
+
+def save_instrument(parent_group, telescope, comm=None, session=None):
```
keskitalo, line 592: Is this essentially implementing the `save_hdf5()` method for the instrument class? Where do we draw the line between classes that implement `save_hdf5()` and classes that have it implemented externally like this?

keskitalo (approved):
Massive body of work and with great added value. On what timescale do you intend to remove support for the older file format?

src/toast/io/observation_hdf_save_v1.py
tskisner: That is a *really* good point.  We ended up in this situation because previously we were not saving/loading a single object here (so we had to do it manually).  But now we are writing the `telescope` and all its members to an HDF5 group.  So you are correct that we should move these functions to methods of the `Telescope` class.

src/toast/io/observation_hdf_save.py
tskisner: Good catch!

tskisner:
> Massive body of work and with great added value. On what timescale do you intend to remove support for the older file format?

I think we should support it as long as possible, and at least while v1 files exist in the wild.  The other thing to ensure is that it should always be possible to load an old format and write a newer format.  That way one could convert from an older file in stages, upgrading between versions.

Another thing I just realized is that we should support more general metadata objects in the Observation dictionary.  Currently we only handle scalars and arrays, but we should add support for serializing python containers (dict, set, list) as well.

I add that and address review comments as well.

src/toast/io/observation_hdf_save_v1.py
tskisner: This is paradigm is now implemented.  All instrument classes implement save / load methods and the observation save / load makes use of those.

src/toast/io/observation_hdf_load.py
```diff
@@ -689,47 +559,56 @@ def load_hdf5_obs_meta(
                     else:
                         msg = f"metadata object group '{obj_name}' has class "
                         msg += f"{obj.attrs['class']}, but instantiated "
-                        msg += f"object does not have a load_hdf5() method"
+                        msg += "object does not have a load_hdf5() method"
```
keskitalo, line 562: Should this be an error instead of a warning? How does a situation like this arise? Does the class only implement `save_hdf5()` and not `load_hdf5()`? or did someone write additional information into the HDF5 after is was created?

src/toast/io/observation_hdf_load.py
```diff
@@ -689,47 +559,56 @@ def load_hdf5_obs_meta(
...
+                    test_obj = objclass()
+                    if hasattr(test_obj, "load_hdf5"):
+                        # Record this in the dictionary of things to load in the
+                        # next step.
+                        attr_load[obj_name] = test_obj
+                    else:
+                        msg = f"attr object group '{obj_name}' has class "
+                        msg += f"{obj.attrs['class']}, but instantiated "
+                        msg += "object does not have a load_hdf5() method"
+                        log.error(msg)
```
keskitalo, line 594: Here we treat the missing `load_hdf5()` as an error.

keskitalo (approved):

src/toast/io/observation_hdf_load.py
tskisner: I had originally wanted to guard against files where other code has added attributes / datasets.  But this code is only executed if the `class` attribute is found, which means it should be a toast compatible object.  I can raise an exception instead of just warning the user.

PR #899, opened 2025-12-10
Add support for precomputed templates in filterbin

keskitalo:
The templates can be fit in any window and several detectors may use the same template(s).

### DIFF
--- a/src/toast/ops/filterbin.py
+++ b/src/toast/ops/filterbin.py
@@ -3,6 +3,7 @@
 # a BSD-style license that can be found in the LICENSE file.
 
 import os
+import pickle
 import re
 from glob import glob
 from time import time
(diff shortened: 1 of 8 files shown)


src/toast/ops/filterbin.py
```diff
@@ -409,6 +520,22 @@ class FilterBin(Operator):
...
+        "detector to a key in the dictionary.  That key will then return a list of "
+        "templates to fit against that detector.  See `precomputed_template_view`."
+        ,
+    )
+
+    precomputed_template_view = Unicode(
+        "throw",
+        allow_none=True,
+        help="Intervals for precomputed template filtering. See `precomputed_templates`"
+    )
```
tskisner, line 537: This view defines the intervals over which the filtering is done.  If the templates themselves are estimated on a different set of intervals (rather than the whole observation), is there any mechanism for ensuring consistency?  Or is the code that creates the templates responsible for stitching them together so that the templates in the dictionary are always full-observation?

tskisner (approved):
Thanks for the ongoing work on intensity templates!

src/toast/ops/filterbin.py
keskitalo: Each template must be a full observation-length TOD but in can have zeros in place of samples that should not be filtered. The view simply determines if the template should be fit piecewise or full length.

PR #902, opened 2025-12-21
Change the name suffixes of filterbin map and noiseweighted map

tskisner:
This makes the default names match the standard mapmaker operator.  If filtering is disabled, the 'unfiltered' suffix is kept.

### DIFF
--- a/src/toast/ops/filterbin.py
+++ b/src/toast/ops/filterbin.py
@@ -1747,8 +1747,8 @@ class FilterBin(Operator):
         cov_name = f"{self.name}_cov"
         rcond_name = f"{self.name}_rcond"
         if filtered:
-            map_name = f"{self.name}_filtered_map"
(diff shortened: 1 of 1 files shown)


keskitalo (commented):
I have a feeling that this will cause some disruption in scripts that deal with TOAST output maps. It makes sense to call FilterBin maps `filtered_map` instead of just `map` but that logic breaks down when we run standalone filters and bin the signal using `MapMaker`.

tskisner:
This change was motivated by trying to [avoid disruption in downstream scripts](https://github.com/simonsobs/sotodlib/blob/7a76e479d33a55d8ba18573a3fa43ddc685a97ee/sotodlib/toast/scripts/so_stats.py#L154-L168).  But I guess those scripts can be modified to add an additional naming suffix.

PR #905, opened 2026-01-02
Fix focalplane loading in workflows

keskitalo:
`Focalplane.load_hdf5()` recently began returning new instances instead of updating in place. This PR upgrades the three TOAST example workflows to use the new API. It also adds support for `thinfp` and `sample_rate` in the load method.

### DIFF
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -985,7 +985,13 @@ class Focalplane(object):
 
     @classmethod
     def _load_hdf5(
-        cls, handle, comm=None, detectors=None, file_det_sets=None, **kwargs
+            cls,
+            handle,
+            comm=None,
+            detectors=None,
+            file_det_sets=None,
+            sample_rate=None,
+            **kwargs,
     ):
         """Load a base class Focalplane"""
         log = Logger.get()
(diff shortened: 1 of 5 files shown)


src/toast/instrument.py
```diff
@@ -995,11 +1001,11 @@ def _load_hdf5(
         need_bcast = hdf5_use_serial(handle, comm) and comm is not None
 
         detector_data = None
-        sample_rate = None
         field_of_view = None
         if handle is not None:
             detector_data = read_table_hdf5(handle, path="focalplane")
-            sample_rate = detector_data.meta["sample_rate"]
+            if sample_rate is None:
+                sample_rate = detector_data.meta["sample_rate"]
```
tskisner, line 1008: I think this is OK, as long as we only use this override for standalone instrument HDF5 files.  If it was somehow triggered as part of loading a larger HDF5 data volume, then the sample rate in the focalplane might not match the detector data.

src/toast/ops/mapmaker.py
```diff
@@ -829,7 +829,7 @@ class Calibrate(Operator):
 
     mc_index = Int(None, allow_none=True, help="The Monte-Carlo index")
 
-    mc_root = Unicode(None, allow_node=True, help="Root name for Monte Carlo products")
+    mc_root = Unicode(None, allow_none=True, help="Root name for Monte Carlo products")
```
tskisner, line 832: Nice catch!

tskisner (approved):
Thanks for updating the example workflows!

PR #907, opened 2026-01-05
Fix issues loading version-1 HDF5 files

tskisner:
- Handle case where version format string was missing

- Add standalone script for generating / comparing HDF5 files.  This was used to generate v1 files with an older toast code version and verify loading with latest code version.

- Add an MPI barrier when writing HDF5 files to a temp location and moving them into place.

### DIFF
--- a/etc/hdf5_formats/save_load.py
+++ b/etc/hdf5_formats/save_load.py
@@ -4,7 +4,7 @@
 # All rights reserved.  Use of this source code is governed by
 # a BSD-style license that can be found in the LICENSE file.
 
-"""Test saving and loading different HDF5 formats"""
+"""Test saving and loading HDF5 formats"""
 
 import argparse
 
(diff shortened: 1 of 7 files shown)


etc/hdf5_formats/save_load.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2025-2025 by the parties listed in the AUTHORS file.
+# All rights reserved.  Use of this source code is governed by
+# a BSD-style license that can be found in the LICENSE file.
+
+"""Test saving and loading different HDF5 formats"""
```
keskitalo: This comment is a little misleading since the script always uses the current default format.

keskitalo (approved):
Glad to have better backwards compatibility!

tskisner:
The unit test failures seem triggered by the update to numpy-2.4.0.  I can reproduce them locally after updating my environment to that version of numpy.  Working on fixing it.

PR #908, opened 2026-01-12
Add support for indexing an HDF5 volume

tskisner:
- Add a new VolumeIndex class, which interfaces to an sqlite DB that can be located anywhere (e.g. a different filesystem with better sqlite performance).  Support indexing arbitrary Observation metadata and nested attributes.  Support selection of observations with standard SQL across an arbitrary MPI communicator.

- Add util function for connections to an sqlite DB with options tuned for parallel / networked filesystems.

- In SaveHDF5, support creating a hierarchy of sub-directories in the volume, arranged by the first 5 digits of UNIX time and / or session name.

- In LoadHDF5, add support for arbitrary SQL select commands when choosing observations to load.

- Expand unit tests to include various uses of the index.

- Remove un-needed imports.

### DIFF
--- a/src/toast/io/CMakeLists.txt
+++ b/src/toast/io/CMakeLists.txt
@@ -9,6 +9,7 @@ install(FILES
     observation_hdf_load_v0.py
     observation_hdf_load_v1.py
     hdf_utils.py
+    hdf_volume.py
     deprecated_compression.py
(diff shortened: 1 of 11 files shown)


keskitalo (approved):
This looks like an elegant solution to data selection. Looking forward to seeing it in action.

PR #910, opened 2026-01-15
Small changes for dealing with numpy arrays of unicode strings

tskisner:
Numpy-2 introduced arrays of variable-length unicode strings.  These are incompatible with h5py datasets.  This PR:

- Adds helper functions for replacing numpy unicode arrays with fixed width byte strings.

- Process all observation metadata through these conversion functions when saving to HDF5.

### DIFF
--- a/src/toast/io/observation_hdf_save.py
+++ b/src/toast/io/observation_hdf_save.py
@@ -538,8 +538,7 @@ def save_hdf5(
     comm = obs.comm.comm_group
     rank = obs.comm.group_rank
 
-    namestr = f"{obs.name}_{obs.uid}"
-    hfpath = os.path.join(dir, f"obs_{namestr}.h5")
+    hfpath = os.path.join(dir, f"{obs.name}.h5")
     hfpath_temp = f"{hfpath}.tmp"
 
     # Create the file and get the root group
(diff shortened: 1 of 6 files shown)


src/toast/io/observation_hdf_save.py
```diff
@@ -538,8 +538,7 @@ def save_hdf5(
     comm = obs.comm.comm_group
     rank = obs.comm.group_rank
 
-    namestr = f"{obs.name}_{obs.uid}"
-    hfpath = os.path.join(dir, f"obs_{namestr}.h5")
+    hfpath = os.path.join(dir, f"{obs.name}.h5")
```
keskitalo, line 541: `Observation._name` initializes to `None`. This is not an issue with this PR but it would be great if we could guarantee that the `name` and `uid` attributes to an observation were always meaningful.

keskitalo (approved):
Looks good.

src/toast/io/observation_hdf_save.py
tskisner: Good point.  I pushed a fix that checks if name is None, and if so, chooses a random integer for both the name and the UID.

PR #911, opened 2026-01-15
bugfix in crosslinking.py

AERAdler:
I don't fully understand why there is this code block, but based on failures of the code, surely the pi/2 value should be only for "tiny" angles?


keskitalo (approved):
Thanks for the fix!

PR #912, opened 2026-01-15
Type 3 LAT scans with starting elevation nod phase per throw

AERAdler:
The commit history isn't super clean unfortunately, I had to pull from toast3 at some point in December.


src/toast/RELEASE
```diff
@@ -1 +1 @@
-3.0.0a41
+3.0.0a41
```
tskisner, line 1: Normally I would ask to restore this, but we are about to release alpha42, so this will get wiped anyway.

tskisner (approved):
This looks fine.  The El modulation is now always done per subscan- is there ever a case where you want to have the previous behavior?  I guess not.

PR #913, opened 2026-01-16
Fix loading of HDF5 volumes

tskisner:
Older HDF5 volumes will not have an index.  If the user does not explicitly disable use of the index, simply print a warning and proceed to scan the filesystem as if the index was disabled.

Also fix a corner case where metadata objects containing the keyword "type" would break the technique we use to instantiate classes based on the contents of the HDF5 attributes before calling the class loading function.


keskitalo (approved):
Good catch!

PR #914, opened 2026-01-23
Cleanup of observation metadata handling in HDF5

tskisner:
- Remove the `obs_approx_equal()` function and simply add optional `approx=` keywords to the `__eq__()` methods of Observation and DetectorData classes.

- Add new helper functions for comparison of arrays in the case where one of the arrays might contain truncated float32 values.

- Add new helper function to compare astropy tables for equality.

- Move unicode array conversion functions from `toast.io.hdf_utils` to `toast.utils`.

- Change naming convention for observation metadata Quantities that are serialized to HDF5.  Fix regex match that was missing some metadata to load.

### DIFF
--- a/src/toast/instrument.py
+++ b/src/toast/instrument.py
@@ -21,8 +21,8 @@ import astropy.time as astime
 import h5py
 import numpy as np
 from astropy import units as u
-from astropy.io.misc.hdf5 import read_table_hdf5, write_table_hdf5
-from astropy.table import Column, QTable
+from astropy.io.misc.hdf5 import read_table_hdf5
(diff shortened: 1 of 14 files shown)


tskisner:
This is finally ready.  I tested these changes loading raw SO data as int32, along with all housekeeping metadata, and saving to HDF5 with FLAC compression.  I then loaded that data back in, used CalibrateDetectors to convert from ADC to normal phase units and compared to SO data loaded as float with all housekeeping.  Along the way I found and fixed some other small bugs.

keskitalo (approved):
Wow, that was a lot of work. Thanks!

PR #923, opened 2026-02-19
Demonstration of printing elapsed time for every exec()

tskisner:


tskisner:
With most recent commit, an example of what is printed when the timing trait is enabled:
```
polyfilter = ops.PolyFilter(
    name="custom_poly1D_name",
    timing=True,
)
polyfilter.apply(data)
```
Output:
```
TOAST INFO: Begin custom_poly1D_name (PolyFilter) exec()
TOAST INFO: End custom_poly1D_name (PolyFilter) exec() in 0.01 s
TOAST INFO: Begin custom_poly1D_name (PolyFilter) finalize()
TOAST INFO: End custom_poly1D_name (PolyFilter) finalize() in 0.00 s
```

keskitalo (approved):
I really like this. Only one question: should `finalize()` also report the time sense the first call to `exec()`?

tskisner:
Based on offline discussion, most recent commit just prints one message at the start of first call to `exec()` and then the timing at the end of `finalize()`:
```
TOAST INFO: custom_poly1D_name (PolyFilter) starting...
TOAST INFO: custom_poly1D_name (PolyFilter) applied in 0.01 s
```

PR #925, opened 2026-02-23
Testing prior to tag

tskisner:
This is a large body of features and fixes, developed in
the context of converting SO data into HDF5 files and
then loading and applying preprocessing.

HDF5:

- Require flacarray-0.4.0, which has a fix for the case
  of loading slices of detector data when using MPI.

- Cleanup detector data load to move some common code
  out of an inner loop.

- Add a new script to re-index an HDF5 volume.  This is
  useful when one wants to index new types of metadata
  that are already in the files (just not in the index).

- Allow indexing of observations where some metadata is
  missing.  A warning is given if a query on an index
  uses fields where some observations do not have that
  metadata.

Flag extensions:

- Add a new utility function that can pad flagged samples
  matching a particular bitmask.  The given bitmask value
  is extended around each flagged region.

- Add an operator that calls the utility function above
  for shared or detector flags and extends by a fixed
  time or number of samples.

- Add unit tests for flag extensions.

Convolution:

- Refactor FFT implementations into separate classes.
  In addition to numpy and internal / compiled cases,
  Add support for nonuniform FFTs and their adjoints
  (type 1 and 2) with the finufft package.
  
- Accept sample flags as input to the convolution
  function.  If these are passed, an impulse response
  is passed through the kernel convolution and the
  spread is measured.  This spread width is used to
  extend flagged regions to avoid contamination of
  good data from flagged samples.

- Update the timeconstant and 1/f noise filter operators
  to use this new flagging support.

- Expand unit tests to also include use of flags.

Pre-processing defaults:

- Do not require input noise model in demodulation
  operator.  Frequently we estimate the noise from
  scratch after demodulation.

- For jump and glitch detection operators, default
  to ignoring invalid samples rather than non-science
  ones.  This allows jump and glitch detection through
  turnarounds.  By default, these operators now fill
  flagged samples with a low-order polynomial that
  spans the gap.  No white noise is added to this.
  Estimating a white noise level from just nearby
  samples is error-prone, and the only required
  feature is that the samples in the gap are smooth
  enough to not trigger the median filter / matched
  filter.  In the future we could also add support
  for filling gaps with white noise levels from a
  noise model.


src/toast/io/hdf_volume.py
```diff
@@ -52,8 +59,21 @@ def select(self, query, comm=None):
...
+            # that their query may be over a subset of the data.
+            self._load_db_schema()
+            warn = f"Selection: '{query}' uses columns with NULL counts:"
+            do_warn = False
+            for col_name, n_null in self._db_null_count.items():
+                if re.search(col_name, query) is not None and n_null > 0:
+                    do_warn = True
+                    warn += f" {col_name} ({n_null}/{self._db_rows}),"
+            if do_warn:
+                log.warning(warn[:-1])
```
keskitalo, line 75: I spent an embarrassing amount of time trying to figure out the `[:-1]`...

keskitalo (approved):
Thanks for adding that warning.  Minimum number of surprises is a good policy.

src/toast/io/hdf_volume.py
tskisner: Sorry, just removing a trailing comma!

src/toast/io/observation_hdf_load.py
```diff
@@ -266,39 +314,25 @@ def load_hdf5_detdata(obs, file_dets, hgrp, fields, log_prefix, parallel):
...
-                    mpi_comm=obs.comm.comm_group,
-                    mpi_dist=mpi_dist,
-                    use_threads=False,
+                    .astype(orig_dtype)
+                    .reshape(obs.detdata[field].shape)
                 )
-                .astype(orig_dtype)
-                .reshape(obs.detdata[field].shape)
-            )
+            except Exception as e:
```
keskitalo, line 332: This looks like a debug statement. Did you intend to leave it in?

keskitalo: See also `dbg_rnk` defined above

src/toast/utils.py
keskitalo, line 1: The copyright is out of date

src/toast/utils.py
```diff
@@ -1037,9 +1050,71 @@ def flagged_noise_fill(data, flags, buffer, poly_order=1):
...
+            )
+
+
+def extend_flags(flags, mask, buffer):
+    """Expand flagged regions.
+
+    The `mask` is bitwise "and" applied to the flags array to determine which
+    samples are "bad".  Contiguous regions of bad samples are identified and
+    then areas of less than `buffer` samples in between two bad regions are flagged
+    with the mask value.  If `extend` is True, the same buffer size is added around
```
keskitalo: There is no `extend` argument to this method.

src/toast/utils.py
```diff
@@ -1037,9 +1050,71 @@ def flagged_noise_fill(data, flags, buffer, poly_order=1):
...
+        mask (int):  The flag mask.
+        buffer (int):  The number of samples to extend flagged regions on either
+            side.
+
+    Returns:
+        (None)
+
+    """
+    matching = np.array(flags & mask, dtype=bool)
+    start_flags = np.where(matching[1:] > matching[:-1])[0] + 1
```
keskitalo, line 1082: This feels like an overly complicated solution. If you just convolved the `matching` array with a constant kernel that is `buffer` wide, you could raise the flags for all nonzero samples. Small amount of padding is needed to avoid wrapping around.

Of course if it works, we may as well keep the index-based solution for now.

src/toast/ops/extend_flags.py
```diff
@@ -0,0 +1,191 @@
...
+        allow_none=True,
+        help="Observation detdata key for flags to use",
+    )
+
+    det_flag_mask = Int(0, help="Bit mask value for detector sample flagging")
+
+    shared_buffer_time = Quantity(
+        None,
+        allow_none=True,
+        help="Flag shared gaps smaller than this time span",
```
keskitalo: This seems misleading. Should it say "Extend flagged regions by this time span" instead?

Similar comments to `shared_buffer_samples`, `det_buffer_time` and `det_buffer_samples`.

src/toast/ops/extend_flags.py
```diff
@@ -0,0 +1,191 @@
...
+                        ob.detdata[self.det_flags][det],
+                        self.det_flag_mask,
+                        det_buf_samp,
+                    )
+                    if (
+                        np.count_nonzero(ob.detdata[self.det_flags][det])
+                        == ob.n_local_samples
+                    ):
+                        msg = f"All samples for detector {det} have been flagged"
+                        log.warning(msg)
```
keskitalo, line 173: Should the detector itself by flagged then?

src/toast/ops/extend_flags.py
keskitalo, line 1: If the data are distributed in time, then there will be edge effects that depend on the data distribution.  We could just require the data to be distributed by detector.  Redistributing data or communicating overlaps seems like a hassle.

src/toast/ops/time_constant.py
```diff
@@ -146,11 +162,28 @@ def _filter_kernel(indx, kfreqs):
...
-            # The slice of detector data we will use
+            # The slice of detector data and flags we will use
             signal = obs.detdata[self.det_data][dets, :]
+            if self.det_flags is None:
+                flags = None
+                flag_mask = None
+            else:
+                flags = obs.detdata[self.det_flags][dets, :]
+                if self.shared_flags is not None:
+                    # These shared flags will effectively be propagated to
```
keskitalo, line 176: Good point! Maybe it should be reflected in the help string to `shared_flags`?

src/toast/tests/helpers/flags.py
```diff
@@ -35,5 +51,22 @@ def fake_flags(
...
-            ob.detdata[det_name][det, :half] |= det_val
+
+        if do_half:
+            for det in ob.local_detectors:
+                ob.detdata[det_name][det, :half] |= det_val
+        if do_random:
+            for det in ob.local_detectors:
+                starts = np.unique(
+                    np.sort(
+                        np.random.randint(
```
keskitalo, line 62: Random thought: do we set the seed for `np.random` anywhere? Do we care if the unit tests run different every time they launch?

src/toast/tests/ops_extend_flags.py
```diff
@@ -0,0 +1,310 @@
...
+        for isamp in range(n_samp):
+            if input[isamp] == 0:
+                continue
+            istart = isamp - buffer
+            if istart < 0:
+                istart = 0
+            istop = isamp + buffer
+            if istop >= n_samp:
+                istop = n_samp - 1
+            for it in range(istart, istop, 1):
```
keskitalo: This loop could be replaced with `if np.any(output[istart : istop] == 0)` and that would be more readable.

src/toast/tests/ops_extend_flags.py
```diff
@@ -0,0 +1,310 @@
...
+            if output[isamp] == 0:
+                continue
+            istart = isamp - buffer
+            if istart < 0:
+                istart = 0
+            istop = isamp + buffer + 1
+            if istop >= n_samp:
+                istop = n_samp - 1
+            total = 0
+            for it in range(istart, istop, 1):
```
keskitalo: This could be `if np.all(input[istart : istop] == 0)`

src/toast/fft.py
keskitalo, line 1: Copyright notice is out of date

src/toast/tests/fft.py
```diff
@@ -524,17 +569,245 @@ def _kernel_func(indx, kfreqs):
...
                 absdiff[bad] = 0
-                if np.amax(absdiff) > 0.1:
+                if np.amax(absdiff) > 0.1 and algo != "nonuniform":
                     print(
                         f"{algo}[{itod}] fail: max absdiff = {np.amax(absdiff)}",
                         flush=True,
                     )
                     self.assertTrue(False)
+
+    def test_convolve_flags(self):
```
keskitalo, line 587: This test doesn't seem to check if the flagged regions were actually increased by the expected amount.

keskitalo (approved):
Looks great! I have made a number of comments but none of them are critical. Merge away when you want.

src/toast/io/observation_hdf_load.py
tskisner: Right, this try / except is intentional.  If calling code has a try block where it loads hdf5 data, sometimes the reason for failing to load (especially when reading flac compressed data on many processes) is really hard to track down.  So this block will log a more detailed error before re-raising.  I don't think it adds too much boilerplate.

src/toast/utils.py
tskisner: Fixed

src/toast/utils.py
tskisner: Good catch, fixed docstring which was stale from an earlier iteration of the code.

src/toast/utils.py
tskisner: I tried a quick implementation, but didn't pass tests.  So I left a comment to revisit in the future.

src/toast/ops/extend_flags.py
tskisner: Good catch, fixed.

src/toast/ops/extend_flags.py
tskisner: Yes!  not sure how I missed that, but thanks for catching.

src/toast/ops/extend_flags.py
tskisner: I added a check for data distribution in detector direction.

src/toast/tests/helpers/flags.py
tskisner: Good question (ahem, "random" thought...).  And of course different test cases also set the numpy random seed.  So if we set it here it means that every observation will always have exactly the same flagging.  I'm not sure if we want that.  Punting for a future discussion.

src/toast/fft.py
tskisner: Fixed.

src/toast/tests/fft.py
tskisner: The flagging function is tested in the ops_extend_flags.py file.  For the convolution, the calling code does not even know the flag buffer in samples that is applied- since that is computed internally from the impulse response.  So in order to test this within the convolution tests I would have to modify the convolution function to return the computed buffer size.

src/toast/tests/ops_extend_flags.py
tskisner: I know some things like this could save lines / compute time, but this is a unit test that already runs really fast.  During debugging, having the explicit sample index that failed in the test was really useful.

src/toast/ops/time_constant.py
tskisner: Good idea, implemented.

PR #926, opened 2026-03-10
Save hits

keskitalo:
Simple modification to the schedule projection tool to allow saving the simulated hit maps.


tskisner (approved):
Looks good, thanks!

PR #929, opened 2026-03-25
Cleanup use of MPIShared objects

tskisner:
MPIShared objects should be considered read-only, except when collectively updating them with `set()` or `__setitem__`.  Future versions of `pshmem` will enforce this behavior.  This PR cleans up some places where we were being sloppy with this behavior.

### DIFF
--- a/src/toast/tests/ops_pointing_wcs.py
+++ b/src/toast/tests/ops_pointing_wcs.py
@@ -201,9 +201,7 @@ class PointingWCSTest(MPITestCase):
 
                 self.assertFalse(pixels.auto_bounds)
                 self.assertTrue(pixels.center == center)
-                self.assertTrue(
-                    pixels.resolution == (0.02 * u.degree, 0.02 * u.degree)
-                )
+                self.assertTrue(pixels.resolution == (0.02 * u.degree, 0.02 * u.degree))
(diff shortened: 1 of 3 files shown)


src/toast/tests/ops_pointing_wcs.py
```diff
@@ -693,9 +691,9 @@ def test_drone_map(self):
             # be realistic, but at least should cover the source
             for obs in data.obs:
                 if obs.comm_col_rank == 0:
-                    obs.shared["boresight_azel"].data[:, :] = obs.shared[
-                        "boresight_radec"
-                    ].data[:, :]
+                    obs.shared["boresight_azel"].set(obs.shared["boresight_radec"].data)
```
keskitalo, line 694: Can you call `set()` on just one process?

keskitalo (approved):

src/toast/tests/ops_pointing_wcs.py
tskisner: No, `set()` (and assignment with `[]` notation) are collective.  That's why the non-rank-zero processes call `set(None)` in the other branch of the if/else here.  Only one process should pass a non-None argument to set().

keskitalo: Somehow I missed that. Thanks.

PR #932, opened 2026-03-26
Large build system overhaul

tskisner:
- Replace setup.py with scikit-build-core called from pyproject.toml

- Simplify cmake files.

- Remove bundled pybind11 and googletest framework. All compiled test functionality is already covered in the python tests.

- Generate version files directly from setuptools_scm controlled by pyproject.toml.

- Begin updating install docs.


tskisner:
Although there has been a recent debacle involving `mkdocs` and downstream projects, I would prefer to keep the current docs framework until after this is merged.  Then we can migrate to zensical or quarto or some other system.

src/libtoast/src/toast_sys_environment.cpp
keskitalo, line 1: Looks like this file got reformatted to a different standard than before. Was that intentional?

src/format_source.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 #
 # Apply TOAST source code formatting.  This script will run the
-# "uncrustify" and "black" formatting tools on C++ and Python code
+# "uncrustify" and "ruff" formatting tools on C++ and Python code
```
keskitalo, line 4: We are switching Python code formatter?

keskitalo (approved):
Wow, that is a lot fewer source files.

src/format_source.sh
tskisner: yes, ruff is an order of magnitude faster and seems to be the "new cool thing" :-)

src/libtoast/src/toast_sys_environment.cpp
tskisner: Not sure what happened here- will fix it.

Issue #477, opened 2021-09-19
Re-confirm mapmaker agreement with Madam in compatible situations


tskisner:
We currently have multiple unit tests that verify consistency between the toast mapmaker and madam when run in the subset of cases that are supported by madam.  @keskitalo recently did some runs with one of the benchmark scripts (as part of #461) running both mapmakers and noticed a discrepancy between their outputs.  We should:
1.  Verify in that workflow that both mapmakers are receiving identical timestream inputs, noise models, configuration inputs, etc.
2. Track down the source of the discrepancy.
3. Understand why our unit tests did not alert us to the issue and fix them.
We can wait to address this until after #461 is merged, which adds the features needed to do this test.

PR #935, opened 2026-03-30
Verify Offset template noise prior consistency with MADAM

tskisner:
- In the Offset template, change the way that the noise
  prior is computed / handled:
  
  - If noise prior is enabled, baselines are forced to be
    contiguous and fixed-length.  View is used only for 
    cutting samples.  If noise prior is not enabled, view
    is used to start / truncate offset sample ranges
    
  - Restore toast-2 spacing of high-resolution offset PSD
    frequencies for evaluation.
    
  - When building the offset PSD, better document the typo
    in the Madam paper and actually implement the fix.
    
  - Do not overwrite solver flag values- make a copy.
  
  - Dynamically changed the banded preconditioner width if
    needed.

- Fix memory bug in compiled offset template code where
  amplitudes are indexed.

- In mapmaker tests:

  - Add ground based offset test
  
  - Avoid debug files and plots when running in CI
  
  - Also compare solved baselines between madam and toast
    when running with a banded preconditioner.

- Add destriping test that works with demodulated data.

- Reset template matrix during initialization.

- Other unrelated cleanups.

### DIFF
--- a/src/toast/ops/sim_cosmic_rays.py
+++ b/src/toast/ops/sim_cosmic_rays.py
@@ -42,7 +42,7 @@ class InjectCosmicRays(Operator):
     from simulations.   For each observation and each detector, we estimate the number of hits expected
     theroretically and draw a random integer, `N`,  with a Poissonian distribution given the expected number
     of events, `Nexp`.  We then  select randomly `N` timestamps where   the hits will be injected into the tod simulated in TOAST.
-    Evaluate the function :math:\gamma at a higher sampling rate (~150 Hz), decimate it to the TOD sample rate and coadd  it.
+    Evaluate the function :math:\\gamma at a higher sampling rate (~150 Hz), decimate it to the TOD sample rate and coadd  it.
 
     Args:
         crfile (string):  A `*.npz`  file encoding 4 attributes,
@@ -142,7 +142,7 @@ class InjectCosmicRays(Operator):
         for ob in data.obs:
             # Get the detectors we are using for this observation
             dets = ob.select_local_detectors(detectors)
-            
+
             comm = ob.comm.comm_group
             rank = ob.comm.group_rank
             # Make sure detector data output exists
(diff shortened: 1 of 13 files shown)


tskisner:
With this PR we have re-established consistency with Madam, including in the case of using a noise prior and a banded preconditioner.  Below are some visual comparisons of the solved baselines and the output maps for a noise prior, 2s baseline, and preconditioner with a bandwidth of 10.  Closes #477.

bandpre_maps
bandpre_base

Small residual differences in the solved baselines are likely due to differences in the convolution used.

src/toast/ops/sim_cosmic_rays.py
```diff
@@ -42,7 +42,7 @@ class InjectCosmicRays(Operator):
     from simulations.   For each observation and each detector, we estimate the number of hits expected
     theroretically and draw a random integer, `N`,  with a Poissonian distribution given the expected number
     of events, `Nexp`.  We then  select randomly `N` timestamps where   the hits will be injected into the tod simulated in TOAST.
-    Evaluate the function :math:\gamma at a higher sampling rate (~150 Hz), decimate it to the TOD sample rate and coadd  it.
+    Evaluate the function :math:\\gamma at a higher sampling rate (~150 Hz), decimate it to the TOD sample rate and coadd  it.
```
keskitalo, line 45: Thanks! These warnings have been driving me nuts.

src/toast/templates/offset/offset.py
```diff
@@ -38,6 +38,16 @@ class Offset(Template):
     so we arrange our template amplitudes in "detector major" order and store offsets
     into this for each observation.
 
+    If a noise model is used, the baseline covariance is computed from the detector
```
keskitalo, line 41: This approach clearly works. Another alternative would have been to raise an error if mixing a noise model with intervals.

keskitalo (approved):
Quite the accomplishment after all these years!

src/toast/templates/offset/offset.py
tskisner: The view / intervals passed to the template come from the higher-level solver, and so we need to accommodate a solver view even if `use_noise_prior` is True.  I will update that docstring though to clarify in the next PR.

PR #937, opened 2026-04-06
Replace deprecated datatype alias

keskitalo:


tskisner:
Thanks for the fix.  Grep-ing through the source, [I think there is another spot here](https://github.com/hpc4cmb/toast/blob/602110b7a165516bca524c2a80fefc5e276f2fde/src/toast/noise_sim.py#L254).

tskisner (approved):

PR #938, opened 2026-04-07
Write the debug plots into the output directory instead of the workin…

keskitalo:
…g directory


tskisner (approved):
A convenient improvement.

PR #940, opened 2026-04-13
Adaptive filter fixes

keskitalo:
Tweaks emerging from working on the adaptive filter problem

### DIFF
--- a/src/toast/scripts/toast_healpix_coadd.py
+++ b/src/toast/scripts/toast_healpix_coadd.py
@@ -37,6 +37,7 @@ def load_map(fname, prefix="", cache=None, dtype=float):
             log.error(msg)
         log.info(prefix + f"Loading {fname} from cache")
         m, good, npix = cache[fname]
+        m = m.copy()  # Avoid modifying the contents of the cache later
         if len(m.shape) != 2:
             msg = f"Cached map '{fname}' does not have the right dimensions: {m.shape}"
             raise RuntimeError(msg)
--- a/src/toast/scripts/toast_healpix_coadd.py
+++ b/src/toast/scripts/toast_healpix_coadd.py
@@ -27,7 +27,7 @@ from toast.utils import Environment, Logger
 import toast
 
 
-def load_map(fname, prefix="", cache=None):
+def load_map(fname, prefix="", cache=None, dtype=float):
     log = Logger.get()
     timer = Timer()
     timer.start()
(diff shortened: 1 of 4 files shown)


src/toast/scripts/toast_healpix_coadd.py
```diff
@@ -27,7 +27,7 @@
 import toast
 
 
-def load_map(fname, prefix="", cache=None):
+def load_map(fname, prefix="", cache=None, dtype=float):
```
tskisner: Looking through the docs, I see that `dtype=float` should be equivalent to `dtype=np.float64`.  But my `C` brain cells are protesting that `float` is 32bit!  This is fine, but I plan to keep using `np.float64` elsewhere in the codebase...

src/toast/scripts/toast_healpix_coadd.py
```diff
@@ -394,12 +437,20 @@ def main(
         if have_hits:
             if infile_hits is not None:
                 hits, nnz3, npix_test, good_hits = load_map(
-                    infile_hits, prefix=prefix, cache=cache
+                    infile_hits, prefix=prefix, cache=cache, dtype=int
```
tskisner: Unlike Python `float`, which is guaranteed to map to a C `double`, `int` is system dependent.  Can we specify `np.int64` here, since I assume you always want to accumulate with int64?

src/toast/scripts/toast_healpix_coadd.py
```diff
@@ -307,14 +322,15 @@ def main(
 
     if args.double_precision:
         dtype = np.float64
+        dtype_hits = np.int64
     else:
         dtype = np.float32
+        dtype_hits = np.int32
```
tskisner: Shouldn't we always use int64 for the hits?  If we are accumulating many maps and then degrading to very low resolution, we might exceed the int32 limits.

tskisner (changes requested):
I think the main change I'm requesting is to always use 64bit integers for accumulating the hits.  In the current code the hits are loaded as float64, and so will never exceed 32bit integer limits.  Now that they are being loaded as integers, combining many maps and then degrading heavily might exceed that.

src/toast/scripts/toast_healpix_coadd.py
keskitalo: I went over the script and replaced every instance of `float` with `np.float64`

src/toast/scripts/toast_healpix_coadd.py
keskitalo: Sure. Done.

src/toast/scripts/toast_healpix_coadd.py
keskitalo: Replaced with `np.int64`

tskisner (approved):
Looks good!

PR #941, opened 2026-04-14
Sampling a_lm into TOD

keskitalo:
This PR implements a TOAST operator that samples one or more a_lm expansions of the sky into TOD. The expansions can be convolved on the fly with a symmetric Gaussian beam.

The operator requires `ducc0` to work.

### DIFF
--- a/src/toast/ops/scan_alm.py
+++ b/src/toast/ops/scan_alm.py
@@ -5,24 +5,27 @@
 import numpy as np
 import traitlets
 from astropy import units as u
+
 try:
     from ducc0 import totalconvolve
+
     ducc_available = True
 except (ModuleNotFoundError, ImportError) as e:
     ducc_available = False
 import healpy as hp
 from pshmem import MPIShared
 
+from .. import qarray as qa
 from ..observation import default_values as defaults
 from ..pixels import PixelData, PixelDistribution
-from ..timing import function_timer, Timer
-from ..traits import Bool, Instance, Int, List, Quantity, Unicode, Unit, trait_docs
+from ..timing import Timer, function_timer
+from ..traits import (Bool, Instance, Int, List, Quantity, Unicode, Unit,
+                      trait_docs)
 from ..utils import Logger
 from .operator import Operator
 from .pipeline import Pipeline
 from .pointing import BuildPixelDistribution
 from .scan_map import ScanMap, ScanMask
-from .. import qarray as qa
 
 
 @trait_docs
@@ -34,6 +37,9 @@ class ScanAlm(Operator):
     observation, the pointing model is used to expand the pointing and
     interpolate values into detector data.
 
+    Since the Stokes weights can carry information beyond the
+    polarization angle, this operator scans the sky into separate I/Q/U
+    timestreams that are co-added with the appropriate pointing weights.
     """
 
     # Class traits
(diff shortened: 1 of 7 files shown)


src/toast/ops/scan_alm.py
```diff
@@ -0,0 +1,362 @@
...
+    Since the Stokes weights can carry information beyond the
+    polarization angle, this operator scans the sky into separate I/Q/U
+    timestreams that are co-added with the appropriate pointing weights.
+    """
+
+    # Class traits
+
+    API = Int(0, help="Internal interface version for this operator")
+
+    file = Unicode(
```
tskisner, line 49: Why not use `List()`?

src/toast/ops/scan_alm.py
```diff
@@ -0,0 +1,362 @@
...
+        None,
+        allow_none=True,
+        help="Path to a_lm FITS file.  Use ';' if providing multiple files.  "
+        "If set, `alms` must be empty.",
+    )
+
+    alms = List(
+        None,
```
tskisner: Container traits cannot be `None`, since that will break config parsing.  This should default to `[]`.

src/toast/ops/scan_alm.py
```diff
@@ -0,0 +1,362 @@
...
+        help="a_lms to scan.  If set, `file` must be None.",
+    )
+
+    fwhm = Quantity(
+        0 * u.deg,
+        help="Additional smoothing to apply to loaded sky",
+    )
+
+    det_data = Unicode(
```
tskisner, line 66: Another place you could use `List()`...

src/toast/ops/scan_alm.py
```diff
@@ -0,0 +1,362 @@
...
+            if len(dets) == 0:
+                # Nothing to do for this observation
+                continue
+            for key in self.det_data_keys:
+                # If our output detector data does not yet exist, create it
+                exists_data = ob.detdata.ensure(
+                    key, detectors=dets, create_units=self.det_data_units
+                )
+                if self.zero:
+                    ob.detdata[key][:] = 0
```
tskisner: This will not modify the memory in place.  `DetectorData.__getitem__` returns a list of numpy arrays that are views into the data for each detector.  This lets us support `DetectorData` objects that are views into existing memory, where the data within each detector is contiguous but the overall memory between detectors is not required to be contiguous.  If you want to clear all detectors you could do:
```
ob.detdata[key].reset()
```
or
```
ob.detdata[key].data[:] = 0
```
or
```
for d in dets:
    ob.detdata[key][d] = 0
```

tskisner (changes requested):
Only two small fixes required:
- List trait cannot be None
- Use reset() or a similar method to clear detector data.

src/toast/ops/scan_alm.py
keskitalo: For symmetry with the other map scanning operators. We should change them all at the same time.

src/toast/ops/scan_alm.py
keskitalo: Done

src/toast/ops/scan_alm.py
keskitalo: I agree, but we should change this with `ScanHealpixDetectorMap` and `ScanHealpixMap`.

src/toast/ops/scan_alm.py
keskitalo: Thanks!

tskisner (approved):
Thanks for the fixes!

PR #944, opened 2026-04-30
LoadHDF5 should fail if the query returns an empty list

keskitalo:


tskisner (approved):
This is a good check, thanks for adding it!

PR #949, opened 2026-05-05
ops/scan_alm.py fixes

Bai-Chiang:


keskitalo (approved):
This looks fine. The source formatting is a little off but we are about to reformat the entire source tree after some changes to `format_source.sh`

PR #954, opened 2026-05-14
Apply filter config

keskitalo:
These changes are necessary to have FilterBin apply different filter configurations to different observations according to a filter configuration file.


src/toast/ops/filterbin.py
```diff
@@ -633,6 +633,29 @@ def _check_binning(self, proposal):
...
+        timer.start()
+        if self.filter_config_file is not None:
+            with open(self.filter_config_file, "r") as f:
+                self.filter_config = yaml.load(f)
+            # Find and translate all quantities
+            for obs_name, obs_config in self.filter_config.items():
+                for key, value in obs_config.items():
+                    if isinstance(value, str) and \
+                       value.endswith((" deg", " rad", " arcmin")):
+                        obs_config[key] = u.Quantity(value)
```
tskisner: Your config file is basically overriding operator traits on a per-observation basis.  In that case, could we use the [existing helper function](https://github.com/hpc4cmb/toast/blob/c6aac720a5dc56b32e4cfb193b3b3c87ce02a68f/src/toast/trait_utils.py#L110) that parses an arbitrary trait value (as a string) into the trait value?  Similarly, when saving trait parameters to a config file, can we [use the helper function](https://github.com/hpc4cmb/toast/blob/c6aac720a5dc56b32e4cfb193b3b3c87ce02a68f/src/toast/trait_utils.py#L146)?

src/toast/ops/flag_nans.py
```diff
@@ -0,0 +1,113 @@
...
+from ..traits import Bool, Float, Instance, Int, List, Quantity, Unicode, trait_docs
+from ..utils import Environment, Logger
+from .operator import Operator
+from .pipeline import Pipeline
+
+XAXIS, YAXIS, ZAXIS = np.eye(3)
+
+
+@trait_docs
+class FlagNaNs(Operator):
```
tskisner, line 28: This operator is a nice addition.  When loading real data we do check for (and flag) NaNs.  If there are NaNs in the timestreams after some processing steps, I wonder if we should `log.warning()` as well as flagging samples?  Seems like that might indicate a problem with an upstream operator.

src/toast/ops/operator.py
```diff
@@ -16,10 +16,11 @@ class Operator(TraitConfig):
 
     """
 
-    timing = Bool(False, help="If True, print timing of exec() and finalize()")
+    timing = Bool(False, help="If True, print timing of exec()")
+    timing_total = Bool(False, help="If True, print timing of finalize()")
```
tskisner, line 20: Thanks for fixing this timing code!

src/toast/data.py
```diff
@@ -102,6 +102,18 @@ def all_local_detectors(self, selection=None, flagmask=0):
...
 
+    def n_obs(self):
+        """Count the current number of observations across all groups.
+
+        Must be called collectively or will cause a hang.
+        """
+        n_obs = len(self.obs)
+        comm = self._comm.comm_group_rank
+        if comm is not None:
+            n_obs = comm.allreduce(n_obs)
```
tskisner: This will work, but if (for example) there are 10 groups of 64 processes each with one group per node, then there will be 10 x allreduce(), each involving 64 processes spread across all 10 nodes.  Alternatively one could do:
```
n_obs = len(self.obs)
if self._comm.comm_group_rank is not None:
    # Only one process from each group sums the total
    if self._comm.group_rank == 0:
        n_obs = self._comm.comm_group_rank.allreduce(n_obs)
    # Broadcast within each group
    if self._comm.comm_group is not None:
        n_obs = self._comm.comm_group.bcast(n_obs, root=0)
```
In this case there would be one allreduce() with 64 processes spread across all 10 nodes, and then 10 bcast each happening within a node.

It probably doesn't matter unless the MPI overhead on an `allreduce()` is high and the group size is large.  And of course we are only summing an integer- this is mainly just about avoiding dozens of allreduce calls.

tskisner (approved):
Several suggestions, especially about string to trait conversion.  But they don't need to hold up progress.

src/toast/ops/filterbin.py
keskitalo: Thanks! I had a hunch this was already solved.

src/toast/ops/flag_nans.py
keskitalo: Great idea!

src/toast/data.py
keskitalo: I agree, yours is better and more pedagogical implementation

PR #955, opened 2026-05-19
ops/ScanAlm: Add focalplane_keys option

Bai-Chiang:
A Comma-separated list of keys to retrieve from the focalplane. Used to expand map file names.

To avoid calculate interpolators repeatedly, we first sort all detectors according to focalplane_keys. We only recalculate interpolators when current detector focalplane_keys and previous one are different, such that they will have different input files.


keskitalo (approved):
Terrific! Thanks for this extension of the ScanAlm operator

PR #957, opened 2026-05-26
Add support for making maps from a subset of detectors

tskisner:
Add support for making maps from a subset of detectors
    
- Cleanup template re-initialization when the pixel
      distribution is changed (for example, when looping
      over subsets of detectors).
    
- Ensure that all detector data and global products
      created by mapmaking are deleted at the end.
    
- Add unit test for mapmaker pattern selection.
    
- Add test for the case where a process has no science
      detectors during template matrix application.
    
- Add support for selecting detectors by pattern for
      individual templates.


src/toast/ops/demodulation.py
```diff
@@ -297,7 +297,10 @@ def _exec(self, data, detectors=None, **kwargs):
                 continue
             n_local = len(obs.local_detectors)
             n_local_good = np.sum(
-                [1 for x, y in obs.local_detector_flags.items() if y == 0]
+                [
+                    1 for x, y in obs.local_detector_flags.items() 
+                    if y & self.det_mask == 0
```
keskitalo, line 302: Nice catch!

src/toast/ops/mapmaker.py
```diff
@@ -691,36 +734,82 @@ def _exec(self, data, detectors=None, use_accel=None, **kwargs):
             # No valid detectors, no mapmaking
             return
 
+        # Loop over focalplane splits, if needed
+        if self.focalplane_key is None:
+            splits = {None: None}
+        else:
+            # Each process gathers their local values for the focalplane keys
+            local_fp_values = dict()
```
keskitalo: Dividing all detectors by a focalplane key seems general enough that we might consider providing a separate utility function for it. We can do that once we need it in another operator.

`Focalplane.detector_groups()` might already do this.

src/toast/templates/offset/offset.py
```diff
@@ -219,10 +220,15 @@ def _initialize(self, new_data):
                         self._freq[iob] = np.logspace(powmin, powmax, 1000)
 
             # Build up detector list
+            det_pat = None
+            if self.pattern is not None:
```
keskitalo, line 226: I cannot figure out where the offset template gets the `pattern` trait

tskisner:
I can reproduce the unit test hang on 2 processes locally.  Investigating.

src/toast/ops/mapmaker.py
```diff
@@ -75,6 +76,16 @@ class MapMaker(Operator):
         defaults.det_data, help="Observation detdata key for the timestream data"
     )
 
+    pattern = Unicode(
```
keskitalo, line 79: I'm sure there will be instances where a regex pattern is an inconvenient way to communicate the detector subset. This is fine for now, but there should probably be another way like a list of detectors as well.

src/toast/ops/noise_model.py
```diff
@@ -701,191 +706,231 @@ def _exec(self, data, detectors=None, **kwargs):
...
+            nobs = data.comm.comm_group_rank.reduce(nobs)
+            nbad = data.comm.comm_group_rank.reduce(nbad)
+            ndet = data.comm.comm_group_rank.reduce(ndet)
 
         log.info_rank(
             f"Flagged {nbad} / {ndet} noise outliers over {nobs} observations",
             comm=data.comm.comm_world,
         )
 
+    def _process_group_prefix(self, obs, local_dets, group, group_dets, prefix):
```
keskitalo, line 761: This routine could use a short docstring

keskitalo (approved):
Looks great! I much appreciate the efforts to make the unit tests cheaper.

src/toast/ops/mapmaker.py
tskisner: Good suggestion to consider `Focalplane.detector_groups()`.  That takes the focalplane table (which contains ALL detectors for the observation, not just local ones) and groups detectors by the value in a column.

For the code in question, we want do essentially the same thing, but considering all observations (which might have disjoint sets of detectors) and also consider local detector flagging.  We already have `Data.all_local_detectors()` which gets the total detectors on a single process across all observations.

I will add a couple new methods to `Data` which do the following:

`Data.all_detectors()`:  get the total set of all detectors across all observations and return to every process.  Accept a det flag mask similar to `all_local_detectors()`.

`Data.all_detector_groups()`:  essentially do what is in this PR code.  Split the focalplane on every observation and build a global grouping of all detectors by focalplane key.  Accept a det flag mask as well.

src/toast/templates/offset/offset.py
tskisner: The pattern is in the `Template` base class, so that it can be used by all templates.

src/toast/ops/mapmaker.py
tskisner: I think what we should do for the general case is migrate the [Splits operator](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/toast/ops/splits.py) upstream to toast (it is completely generic and not specific to any experiment).  I'll do that in a future PR.

src/toast/ops/noise_model.py
tskisner: Done

src/toast/ops/mapmaker.py
tskisner: This is now implemented, and in fact there was another small fix to the loop over splits in the Mapmaker operator.

keskitalo (approved):
Thanks for catching the unhandled `detectors` argument!

PR #958, opened 2026-05-29
ops.ScanAlm: optimize memory

Bai-Chiang:
- blm is the same across all MPI process, store it in MPIShared
- Clear alm memory after building the interpolator.
- Only use alm file name as input. The alm list input was broken. The lmax was only set when loop through file name list. Instead if fixing it. It's better just remove this option. It doesn't make sense to put long list of complex number as arg input.
- Also fixed concurrency issue when focalplane_keys is specified.

### DIFF
--- a/src/toast/ops/scan_alm.py
+++ b/src/toast/ops/scan_alm.py
@@ -47,17 +47,9 @@ class ScanAlm(Operator):
     API = Int(0, help="Internal interface version for this operator")
 
     file = Unicode(
-        None,
-        allow_none=True,
         help="Path to a_lm FITS file.  Use ';' if providing multiple files.  "
         "Any focalplane key listed in `focalplane_keys` can be used here and even "
         "formatted. For example: {psi_pol:.0f}"
-        "If set, `alms` must be empty.",
-    )
-
-    alms = List(
-        [],
-        help="a_lms to scan.  If set, `file` must be None.",
     )
 
     fwhm = Quantity(
@@ -104,8 +96,6 @@ class ScanAlm(Operator):
         allow_none=True,
         help="This must be an instance of a Stokes weights operator",
     )
-
-    save_alm = Bool(False, help="If True, do not delete alm during finalize")
 
     @traitlets.validate("det_mask")
     def _check_det_mask(self, proposal):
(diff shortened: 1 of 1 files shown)


src/toast/ops/scan_alm.py
```diff
@@ -294,44 +284,97 @@ def _scan_alms(self, interpolators, theta, phi, weights):
...
-        # Get an mmax=0 symmetric temperature beam expansion
-        self._blm_I = np.atleast_2d(
-            hp.blm_gauss(
-                self.fwhm.to_value(u.rad),
-                self.lmax,
-                pol=False,
-            )
-        )
+        world_comm = data.comm.comm_world
+        if world_comm is None:
```
keskitalo: Use `world_rank = data.comm.world_rank`. It is always set.

src/toast/ops/scan_alm.py
```diff
@@ -382,52 +430,51 @@ def _exec(self, data, detectors=None, **kwargs):
...
-                if not self.save_alm:
-                    for alm in self.alms:
-                        alm.close()
-                    self.alms = []
+            det = fp_ob_det[-1]
+            fp = fp_ob_det[:-2]
+            if self.focalplane_keys is not None and fp != prev_fp:
+                # Clean up the interpolators
+                if prev_fp is not None:
+                    log.info_rank(f"Group {data.comm.group:4} : Done detectors with {focalplane_key_value}", timer=timer, comm=gcomm)
```
keskitalo: Are you certain that all processes in the group will call `log.info_rank()` the same number of times? If not, the operator will hang. If there is any doubt, you could use `log.debug()` instead and have all processes report.

src/toast/ops/scan_alm.py
```diff
@@ -382,52 +430,51 @@ def _exec(self, data, detectors=None, **kwargs):
...
-                    focalplane_key_value[key] = key_ob_det[ikey]
-                self._load_alm(data, focalplane_key_value)
-                self._cache_blm()
-                self._cache_interpolators()
-                prev_focalplane_keys = focalplane_keys
-                log.info_rank(f"Group {data.comm.group:4} : Start detectors with {focalplane_key_value}", timer=timer, comm=gcomm)
+                    focalplane_key_value[key] = fp_ob_det[ikey]
+                self._cache_interpolators(data, focalplane_key_value)
+                prev_fp = fp
+                log.info_rank(f"Group {data.comm.group:4} : Done building interpolator with {focalplane_key_value}", timer=timer, comm=gcomm)
```
keskitalo: Same concern over blocking communication.

src/toast/ops/scan_alm.py
```diff
@@ -382,52 +430,51 @@ def _exec(self, data, detectors=None, **kwargs):
...
-                    ref -= sig
+                    ob.detdata[det_data_key][det] -= sig
                 else:
-                    ref += sig
-            if idet % 100 == 0:
-                log.debug_rank(f"Group {data.comm.group:4} : {idet}/{ndet} detectors finished in", timer=timer, comm=gcomm)
+                    ob.detdata[det_data_key][det] += sig
+                del interpolators, sig, theta, phi, weights
+            if idet % 1000 == 0:
+                log.debug_rank(f"Group {data.comm.group:4} : {idet}/{ndet} detectors finished", comm=gcomm)
```
keskitalo: Same concern over hanging. Probably best to use `log.debug()` instead of `log.debug_rank()`.

keskitalo (changes requested):
I think there are a few places where the code could hang if the processes have different numbers of detectors.

keskitalo (approved):
Looks great!

PR #959, opened 2026-06-03
Change entrypoints to manage communicator initialization

tskisner:
Some of our scripts have a main() that is used as a function from other code and accepts commandline overrides and optional communicators.  Move the exception guards and communicator initialization to a separate cli() method and use that as the entry point.


keskitalo (approved):
This PR solves a recent regression that broke the MPI parallelism in TOAST workflows.

PR #962, opened 2026-06-22
Ensure proper functioning when some processes have no data

tskisner:
When working with real or simulated data with many detectors that are cut, a frequent scenario is that some processes will have zero good detectors for a given observation.

- Add options to the unit test helper functions for generating satellite and ground data to enable flagging of all detectors on the last process in a group.  Enabled by default.

- When working with template data in the Amplitude class and TODs in the DetectorData class, use zero-length array when a process has no local data.  This is less clunky than setting the local data array to None and then checking for None all over the place.

- In SimAtmosphere, remove shortcut where processes with no dets would return.  All processes later participate in communication, so this caused a deadlock.

- Fix ScanAlm operator, which used collective operations that only worked in specific situations.

- Linting of pixels_io_healpix.py

- In Offset template writing, handle case where some processes have no data.

- In Periodic template writing, handle case where some processes have no data.

- In HDF5 I/O tests, add a base Noise class in addition to the AnalyticNoise instance.

- Disable all-dets-flagged on a process for tests involving Madam, which does not support that case.

- Refactor the CrossLinking operator, which was basically manually running a Pipeline without using one.

- Various small cleanups

### DIFF
--- a/src/toast/observation_data.py
+++ b/src/toast/observation_data.py
@@ -107,7 +107,7 @@ class DetectorData(AcceleratorObject):
                 msg += f" than constructor shape ({self.shape[0]})"
                 log.error(msg)
                 raise RuntimeError(msg)
-            if self._shape[1:] != view_data[0].shape:
+            if self._shape[0] > 0 and self._shape[1:] != view_data[0].shape:
                 msg = f"view data sample shape ({view_data[0].shape})"
                 msg += f" does not match constructor sample shape ({self._shape[1:]})"
                 log.error(msg)
--- a/src/toast/observation_data.py
+++ b/src/toast/observation_data.py
@@ -94,9 +94,6 @@ class DetectorData(AcceleratorObject):
 
         self._fullsize = 0
         self._memsize = 0
-        self._raw = None
-        self._flatdata = None
-        self._data = None
 
         if view_data is None:
             # Allocate the data
@@ -162,29 +159,24 @@ class DetectorData(AcceleratorObject):
             self.accel_delete()
 
         # Delete existing wrapper and buffer
-        if self._data is not None:
+        if hasattr(self, "_data"):
             del self._data
-        if self._flatdata is not None:
+        if hasattr(self, "_flatdata"):
             del self._flatdata
-        if self._raw is not None:
+        if hasattr(self, "_raw"):
             self._raw.clear()
             del self._raw
 
         # Allocate new host buffer
-        if self._fullsize > 0:
-            self._raw = self._storage_class.zeros(self._fullsize)
-            # Wrap _raw
-            self._flatdata = self._raw.array()[: self._flatshape]
-            self._data = self._flatdata.reshape(self._shape)
-            # Restore device buffer if needed
-            if create_accel:
-                self._accel_create(zero_out=True)
-            if on_accel:
-                self.accel_used(True)
-        else:
-            self._raw = None
-            self._flatdata = None
-            self._data = None
+        self._raw = self._storage_class.zeros(self._fullsize)
+        # Wrap _raw
+        self._flatdata = self._raw.array()[: self._flatshape]
+        self._data = self._flatdata.reshape(self._shape)
+        # Restore device buffer if needed
+        if create_accel:
+            self._accel_create(zero_out=True)
+        if on_accel:
+            self.accel_used(True)
 
     @property
     def detectors(self):
@@ -233,7 +225,7 @@ class DetectorData(AcceleratorObject):
 
     @property
     def data(self):
-        if (not hasattr(self, "_data")) or self._data is None:
+        if not hasattr(self, "_data"):
             raise RuntimeError("Cannot use DetectorData object after clearing memory")
         if self._is_view:
             msg = "DetectorData view is not guaranteed to be contiguous.  "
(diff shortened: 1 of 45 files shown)


src/toast/ops/conviqt.py
```diff
@@ -449,7 +451,8 @@ def get_pointing(self, data, det, verbose):
         all_psi_beam = []
         all_hwp_angle = []
         for obs in data.obs:
-            if det not in obs.local_detectors:
+            good_dets = set(obs.select_local_detectors(flagmask=self.det_mask))
```
keskitalo, line 454: Nice! This transforms an O(N) operation into O(log(N)). I wonder if `select_local_detectors` should directly return a set?

src/toast/ops/crosslinking.py
```diff
@@ -28,6 +28,122 @@ def detector_weight(self, det):
         return 1.0 / (u.K**2)
 
 
+@trait_docs
+class CrossLinkingWeights(Operator):
+    """Helper operator to compute the crosslinking weights.
```
keskitalo, line 33: This refactor makes a lot of sense

src/toast/tests/helpers/space.py
```diff
@@ -202,6 +208,23 @@ def create_satellite_data(
                     det_flags[det] = defaults.det_mask_invalid
             ob.update_local_detector_flags(det_flags)
 
+    if flagged_obs:
+        for iob, ob in enumerate(data.obs):
+            if iob % 2 != 0:
+                det_flags = dict()
+                for det in ob.local_detectors:
+                    det_flags[det] = defaults.det_mask_invalid
+                ob.update_local_detector_flags(det_flags)
```
keskitalo, line 217: This is great. Should we also test what happens if all of the detector samples are flagged but the detector is not?

src/toast/tests/ops_sim_tod_conviqt.py
```diff
@@ -321,8 +321,6 @@ def test_sim_conviqt(self):
         toast_hits_path = os.path.join(self.outdir, "toast_hits.fits")
         data[cov_and_hits.hits].write(toast_hits_path)
 
-        fail = False
```
keskitalo, line 324: Nice catch. These have been left in the source for a long time.

src/toast/tests/template_fourier2d.py
```diff
@@ -46,8 +46,7 @@ def test_projection(self):
 
         # Project.
         for det in tmpl.detectors():
-            for ob in data.obs:
```
keskitalo, line 49: How did that get in there?!

src/toast/observation_data.py
```diff
@@ -349,16 +341,12 @@ def clear(self):
...
         if hasattr(self, "_is_view") and not self._is_view:
             if hasattr(self, "_flatdata"):
                 del self._flatdata
-                self._flatdata = None
             if hasattr(self, "_raw"):
-                if self._raw is not None:
-                    self._raw.clear()
-                    del self._raw
-                    self._raw = None
+                self._raw.clear()
```
keskitalo, line 348: `self._raw` will never be `None`?

src/toast/pixels_io_healpix.py
```diff
@@ -228,7 +228,7 @@ def read_healpix(filename, *args, **kwargs):
         if mapdata.shape[0] == 1:
             mapdata = mapdata[0]
 
-        if "h" in kwargs and kwargs["h"] == True:
+        if "h" in kwargs and kwargs["h"]:
```
keskitalo: I get a little uneasy with tests like this since surprising things evaluate to `True` or `False`:
```
In [1]: values=[False, True, 0, 1, [], [1], None]

In [2]: for v in values:
   ...:     print(v, end="")
   ...:     if v:
   ...:         print(" -- yes")
   ...:     else:
   ...:         print(" -- no")
   ...: 
False -- no
True -- yes
0 -- no
1 -- yes
[] -- no
[1] -- yes
None -- no
```

keskitalo (approved):
This PR makes TOAST much more robust against real world data.

src/toast/ops/conviqt.py
tskisner: There are many cases where we want to iterate over detectors in the same order they appear in memory.  So I think it is best to leave this as a list.

src/toast/tests/helpers/space.py
tskisner: If all samples were flagged but not the detector, then the detector would still be considered everywhere across the code (just like they were a good detector).  Obviously more tests would be good, but this PR was focused on the case of handling detectors that are cut.  And this PR is already pretty big...

src/toast/observation_data.py
tskisner: With the refactor in this branch, the raw buffer and its wrapper are always created (even if size is zero).  So yes, they should never be None.

src/toast/pixels_io_healpix.py
tskisner: Fair enough.  Just to compare:
```
>>> values=[False, True, 0, 1, [], [1], None]
>>> for v in values:
...     print(v, end="")
...     if v == True:
...         print(" -- yes")
...     else:
...         print(" -- no")
...
False -- no
True -- yes
0 -- no
1 -- yes
[] -- no
[1] -- no
None -- no
```
The linter was wrong!  I'll revert.

PR #963, opened 2026-06-22
Support optional jinja2 templating in yaml config files

tskisner:
In some cases we resort to writing shell scripts to generate yaml files, when the jinja2 package can already do extensive python-like control structures in the yaml.  If jinja2 is importable, pass yaml files through that when loading.


keskitalo (approved):
Looks cool! We might want to create an issue to add documentation for this.

PR #967, opened 2026-06-29
Detector grouping is global, not local

keskitalo:
The recent modifications to this code broke the detector counting. If noise cuts were done in detector groups, those groups were extracted from the Focalplane and were global. However, if the cuts were done over all detectors, the list of detectors was limited to the local detectors.

This PR handles `focalplane_key == None` through the same function as any valid key and produces consistent results.


tskisner (approved):
Thanks for catching and fixing this!

PR #968, opened 2026-07-02
Filterbin binner traits

keskitalo:
Small tweaks to support adaptive filtering.

### DIFF
--- a/src/toast/ops/filterbin.py
+++ b/src/toast/ops/filterbin.py
@@ -190,7 +190,7 @@ class SparseTemplates:
         return normalized
 
     @function_timer
-    def build_template_covariance(self, good):
+    def build_template_covariance(self, good, obs, det):
         """Calculate (F^T N^-1_F F)^-1
 
         Observe that the sample noise weights in N^-1_F need not be the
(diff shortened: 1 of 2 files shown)


src/toast/ops/filterbin.py
```diff
@@ -721,6 +723,25 @@ def _exec(self, data, detectors=None, **kwargs):
...
+            self.det_flags_save = None
+        else:
+            self.det_flags_save = self.binning.det_flags
+            self.binning.det_flags = self.det_flags
+        if self.binning.shared_flags == self.shared_flags:
+            self.shared_flags_save = None
+        else:
+            self.shared_flags_save = self.binning.shared_flags
+            self.binning.shared_flags = self.shared_flags
+
```
tskisner, line 744: This obviously works.  However you don't need the conditional checks here and when restoring.  You can just do:
```
self.det_data_save = self.binning.det_data
self.binning.det_data = self.det_data

self.det_flags_save = self.binning.det_flags
self.binning.det_flags = self.det_flags

self.shared_flags_save = self.binning.shared_flags
self.binning.shared_flags = self.shared_flags

self.det_data_units_save = self.binning.det_data_units
self.binning.det_data_units = self._det_data_units
```
And then later
```
self.binning.det_data = self.det_data_save
self.binning.det_flags = self.det_flags_save
self.binning.shared_flags = self.shared_flags_save
self.binning.det_data_units = self.det_data_units_save
```
This is what we do in the `MapMaker` operator.  Doing the unconditional copy of a string reference is probably much cheaper than doing an `if` conditional.

tskisner (approved):

PR #969, opened 2026-07-20
Clean up use of np.array constructor.

tskisner:
In some cases the np.array constructor was being used when really np.copy was the desired operation.  Also simplify the 1D poly filter code.

### DIFF
--- a/src/toast/ops/demodulation.py
+++ b/src/toast/ops/demodulation.py
@@ -864,13 +864,11 @@ class Demodulate(Operator):
                 demod_detectors.append(demod_det)
                 demod_freqs[demod_det] = freq_out
                 demod_psds[demod_det] = psd_out
-                demod_indices[demod_det] = noise.index(det) * n_mode + indexoff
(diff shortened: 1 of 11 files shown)


src/toast/ops/demodulation.py
AERAdler, line 1: Why is the detailed indexing deleted?

AERAdler (approved):
Approved, just curious about the indexing changes (see comment)

src/toast/ops/demodulation.py
tskisner: Those "indices" are actually unique random identifiers for each detector in the model.  They are limited to 32bit integers.  The way they were being chosen for the demodulated data was to take the original un-demodulated ID (`X`) and assign that to the intensity channel.  Then the 4f channels were assigned unique IDs at `3 * X + 1` and `3 * X + 2`.  If `X` is close to the 32bit max value, then these will overflow.  Instead, we pass `None` to the `Noise()` class constructor and this generates new unique IDs based on the demodulated names.

PR #971, opened 2026-08-19
Longer turnarounds

keskitalo:
This PR adds traits to the FilterBin operator to truncate the science scan views.  We are using the operator to allow using different truncation parameters to different observations. If the optimal truncation turns out to be static, then the truncation should be done in preprocessing, not in FilterBin.

### DIFF
--- a/src/toast/ops/filterbin.py
+++ b/src/toast/ops/filterbin.py
@@ -1943,6 +1943,8 @@ class FilterBin(Operator):
             "rightleft_interval",
             "poly_filter_order",
             "poly_filter_view",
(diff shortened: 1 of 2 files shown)


tskisner (approved):
This is nice to have a way to iterate over per-observation turnaround extensions.  I can work on adding a similar buffer to `AzimuthIntervals` in the case where the needed buffer is constant for many days.  My hope is that there will be a constant buffer size needed for a given time period in between hardware changes.  For example, events like service to the mounts or changes to the turnaround acceleration profiles.

PR #972, opened 2026-08-21
Add missing property

keskitalo:


tskisner (approved):
Thanks for the fix, and sorry this was missed in that earlier PR.  Obviously the general case of relying on the number of characters in a string to determine the non-zeros of the pointing matrix was fragile.  Much better to have every weights operator report exactly this number.

PR #973, opened 2026-08-26
Update the way that detector splits are implemented in mapmaking

tskisner:
- Add functions to select and de-select detectors across all
      observations using temporary per-detector flags.  This ensures
      that even for a combination of regex matching and focalplane
      splits, the correct detectors are used.
    
- Simplify some of the setup functions in the mapmaker.
    
- Add another unit test for making per-focalplane-pixel maps.

### DIFF
--- a/src/toast/ops/mapmaker.py
+++ b/src/toast/ops/mapmaker.py
@@ -193,6 +193,8 @@ class MapMaker(Operator):
 
     report_memory = Bool(False, help="Report memory throughout the execution")
 
+    _log_prefix = "MapMaker"
+
     @traitlets.validate("map_binning")
     def _check_map_binning(self, proposal):
         bin = proposal["value"]
@@ -251,7 +253,8 @@ class MapMaker(Operator):
             else:
                 fname = os.path.join(self.output_dir, f"{rootname}_{product}.fits")
             if self.mc_mode and not force and os.path.isfile(fname):
-                log.info_rank(f"Skipping existing file: {fname}", comm=self._comm)
+                msg = f"{self._log_prefix} Skipping existing file: {fname}"
+                log.info_rank(msg, comm=self._comm)
             else:
                 self._data[prod_key].write(
                     fname,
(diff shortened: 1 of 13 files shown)


src/toast/ops/mapmaker_utils/mapmaker_utils.py
```diff
@@ -1151,7 +1151,7 @@ def _exec(self, data, detectors=None, **kwargs):
                 pixel_pointing=self.pixel_pointing,
                 save_pointing=self.save_pointing,
             )
-            pix_dist.apply(data)
+            pix_dist.apply(data, detectors=detectors)
```
keskitalo, line 1157: Nice catch!

src/toast/ops/mapmaker.py
```diff
@@ -740,23 +770,40 @@ def _exec(self, data, detectors=None, use_accel=None, **kwargs):
...
-            if split_dets is None:
-                split_dets = detectors
+            n_split_dets = len(split_dets)
+            if n_split_dets == 0:
+                msg = f"{self._log_prefix} Detector split '{split_key}' "
+                msg += "has no dets, skipping"
+                log.info_rank(
+                    msg,
+                    comm=data.comm.comm_world,
+                )
```
keskitalo, line 784: Should the loop short-circuit to the next iteration here?

src/toast/ops/pipeline.py
```diff
@@ -145,7 +145,7 @@ def _exec(self, data, detectors=None, use_accel=None, **kwargs):
                 self._exec_operator(
                     op,
                     data,
-                    detectors=None,
+                    detectors=detectors,
```
keskitalo, line 148: Another nice catch :)

keskitalo (approved):
Excellent! Glad to have the new unit test added and passing.

src/toast/ops/mapmaker.py
tskisner: Thanks, fixed now.

PR #975, opened 2026-09-16
Add support for boresight offset in toast_project_schedule

tskisner:
This adds arguments to specify the Xi / Eta focalplane offset for the calculation of the hit map directly from a schedule.  The boresight sparse Az/El are first computed as usual.  Then quaternion rotations are used to convert these to the Az/El seen by the specified focalplane offset.


src/toast/scripts/toast_project_schedule.py
```diff
@@ -199,8 +215,8 @@ def get_hits(args, schedule, period_times, comm, rank):
         hits[iperiod], iscan = get_period_hits(
             args, schedule, period_times, iperiod, iscan
         )
-    if comm is not None:
-        hits = comm.allreduce(hits)
+    if comm is not None and comm.size > 1:
+        comm.Allreduce(MPI.IN_PLACE, hits, op=MPI.SUM)
```
keskitalo, line 219: I'm fine with this change but curious if the higher level call was producing issues.

keskitalo (approved):
Thanks for the added capabilities to this script!

src/toast/scripts/toast_project_schedule.py
tskisner: Yes, I made this change since the previous line seemed to double the memory use at this step.  I noticed since I was running locally and my system started to swap :-)  After this change the memory used did not spike.

